PR #3, opened 2020-08-16
Removing staging branch and merging all changes since v0 into master to use tag-based deployment approach instead.

nekokatt:


davfsa:
LGTM :+1:

PR #5, opened 2020-08-16
Added greetings workflow.

nekokatt:


.github/workflows/greetings.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  greeting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/first-interaction@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        issue-message: 'Thanks for making your first issue! If you have any questions, please hop over to our Discord at https://discord.gg/Jx4cNGG and ask!'
+        pr-message: 'Thanks for making your first PR to Hikari! We will take a look as soon as we can. Until then, feel free to ask any questions over at https://discord.gg/Jx4cNGG!'
```
FasterSpeeding, line 13: Just a random question that you probably won't have an answer to but will it by any chance parse that "!" as a part of the url at the end of line 13

davfsa: Actually, didn't even realize that, maybe add a space in between for good measure?

PR #6, opened 2020-08-17
Implement "chunk" method for lazy iterator and add __iter__ override for more helpful error message.

FasterSpeeding:
Implement "chunk" method for lazy iterator and add __iter__ override for more helpful error message.

* Fix LazyIterator.last returning the first entry of the last page instead of the first.
* Rename LazyIterator.first to "next".
* Cleanup some doc references to type hints within iterators.py

### DIFF
--- a/hikari/utilities/iterators.py
+++ b/hikari/utilities/iterators.py
@@ -33,6 +33,8 @@ __all__: typing.Final[typing.List[str]] = [
     "All",
     "AttrComparator",
     "BufferedLazyIterator",
+    "ValueT",
+    "AnotherValueT",
 ]
 
 import abc
@@ -42,7 +44,9 @@ import typing
 from hikari.utilities import spel
 
 ValueT = typing.TypeVar("ValueT")
+"""Type-hint of the type of the value returned by a lazy iterator."""
 AnotherValueT = typing.TypeVar("AnotherValueT")
+"""Type-hint of the type of a value by a mapped lazy iterator."""
 
 
 class All(typing.Generic[ValueT]):
@@ -80,7 +84,7 @@ class All(typing.Generic[ValueT]):
 
     Operators
     ---------
-    * `this(value : T) -> bool`:
+    * `this(value : ValueT) -> bool`:
         Return `builtins.True` if all conditions return `builtins.True` when
         invoked with the given value.
     * `~this`:
@@ -89,7 +93,7 @@ class All(typing.Generic[ValueT]):
 
     Parameters
     ----------
-    *conditions : typing.Callable[[T], builtins.bool]
+    *conditions : typing.Callable[[ValueT], builtins.bool]
         The predicates to wrap.
     """
 
@@ -117,9 +121,9 @@ class AttrComparator(typing.Generic[ValueT]):
         If the attribute name ends with a `()`, then the call is invoked
         rather than treated as a property (useful for methods like
         `str.isupper`, for example).
-    expected_value : T
+    expected_value : typing.Any
         The expected value.
-    cast : typing.Callable[[T], typing.Any] or builtins.None
+    cast : typing.Callable[[ValueT], typing.Any] or builtins.None
         Optional cast to perform on the input value when being called before
         comparing it to the expected value but after accessing the attribute.
     """
@@ -200,6 +204,21 @@ class LazyIterator(typing.Generic[ValueT], abc.ABC):
     """
 
     __slots__: typing.Sequence[str] = ()
+
+    def chunk(self, chunk_size: int) -> LazyIterator[typing.Sequence[ValueT]]:
+        """Return results in chunks of up to `chunk_size` amount of entries.
+
+        Parameters
+        ----------
+        chunk_size : int
+            The limit for how many results should be returned in each chunk.
+
+        Returns
+        -------
+        LazyIterator[typing.Sequence[ValueT]]
+            `LazyIterator` that emits each chunked sequence.
+        """
+        return _ChunkedLazyIterator(self, chunk_size)
 
     def map(
         self, transformation: typing.Union[typing.Callable[[ValueT], AnotherValueT], str],
@@ -452,7 +471,7 @@ class LazyIterator(typing.Generic[ValueT], abc.ABC):
 
         Returns
         -------
-        LazyIterator[T]
+        LazyIterator[ValueT]
             A paginated results view that asynchronously yields a maximum
             of the given number of items before completing.
         """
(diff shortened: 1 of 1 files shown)


hikari/utilities/iterators.py
```diff
@@ -201,6 +205,21 @@ class LazyIterator(typing.Generic[ValueT], abc.ABC):
...
+
+        Parameters
+        ----------
+        chunk_size : int
+            The limit for how many results should be returned in each chunk.
+
+        Returns
+        -------
+        LazyIterator[ValueT]
+            `LazyIterator` that emits each chunked sequence.
```
nekokatt, line 219: This isn't `ValueT` is it? It is `Sequence[ValueT]`
```suggestion
        LazyIterator[typing.Sequence[ValueT]]
            `LazyIterator` that emits zero or more sequences containing the given 
            `chunk_size` of items.
```

hikari/utilities/iterators.py
```diff
@@ -468,24 +487,24 @@ def skip(self, number: int) -> LazyIterator[ValueT]:
...
-        T
-            The first result.
+        ValueT
+            The next result.
 
         Raises
         ------
         builtins.LookupError
-            If no result exists.
+            If no more results exists.
```
nekokatt, line 507:
```suggestion
            If no more results exist.
```

hikari/utilities/iterators.py
```diff
@@ -497,19 +516,32 @@ async def last(self) -> ValueT:
 
         Returns
         -------
-        T
+        ValueT
             The last result.
 
+        !!! note
+            This method will deplete the iterator if run.
```
nekokatt, line 523: Maybe `advance` is a better word than `deplete`?

I'm also wondering if we can provide a `peek` method that does the same thing but does not displace the item from the iterator.

E.g.
```py
# Use a dummy enum as we can use `typing.Literal` with these!
class _PeekSentinel(enum.Enum):
    NO_VALUE = enum.auto()

_peek: Union[ValueT, Literal[_PeekSeninel.NO_VALUE]] 

async def peek(self) -> ValueT:
    if self._peek is self._PeekSentinel.NO_VALUE:
        item = await self.next()
        self._peek = item
    return self._peek

async def __anext__(self) -> ValueT:
    if self._peek is not self._PeekSentinel.NO_VALUE:
        item = self._peek
        self._peek = self._PeekSentinel.NO_VALUE
        return item
    else:
        # existing code.
```

This would just mean that peeking would cache the next value eagerly and then clear it next time __anext__ is called.

hikari/utilities/iterators.py
```diff
@@ -497,19 +516,32 @@ async def last(self) -> ValueT:
...
         builtins.LookupError
             If no result exists.
         """
         try:
-            return (await self)[0]
+            return (await self.reversed())[0]
         except IndexError:
             raise LookupError("No elements were found") from None
 
+    async def reversed(self) -> typing.Sequence[ValueT]:
```
nekokatt, line 535: Can we return a LazyIterator instead of a sequence here.

Reason I ask is that we can then chain stuff onto this.

hikari/utilities/iterators.py
```diff
@@ -630,6 +662,10 @@ def __aiter__(self) -> LazyIterator[ValueT]:
         # We are our own iterator.
         return self
 
+    def __iter__(self) -> LazyIterator[ValueT]:
+        # This iterator is async only.
+        raise TypeError(f"{type(self).__name__} is an async-only iterator, did you mean 'async for'?")
```
nekokatt, line 667:
```suggestion
    def __iter__(self) -> typing.NoReturn:
        # This iterator is async only.
        cls = type(self)
        raise TypeError(f"{cls.__module__}.{cls.__qualname__} is an async-only iterator, did you mean 'async for'?")
```

hikari/utilities/iterators.py
```diff
@@ -797,7 +833,30 @@ async def __anext__(self) -> ValueT:
...
+        self._iterator = iterator
+        self._chunk_size = chunk_size
+
+    async def __anext__(self) -> typing.Sequence[ValueT]:
+        chunk = []
+
+        async for item in self._iterator:
+            chunk.append(item)
+
+            if len(chunk) == self._chunk_size:
```
nekokatt, line 853: could this just be 
```py
return await self.limit(chunk_size)
```

hikari/utilities/iterators.py
FasterSpeeding: "advance" doesn't really communicate the fact that it totally uses up the lazy iterator

hikari/utilities/iterators.py
FasterSpeeding: Well it won't be lazy anymore but sure

nekokatt: It can be, if the implementation overrides this.

For example a paginated endpoint could just theoretically change the call it was making if the info were available.

hikari/utilities/iterators.py
FasterSpeeding: wouldn't that introduce extra over head?

hikari/utilities/iterators.py
FasterSpeeding: If i do that we won't be able to override it with ```python
    def __iter__(self) -> FlatLazyIterator[ValueT]:
        return self
``` in the FlatLazyIterator

hikari/utilities/iterators.py
FasterSpeeding: also if those implementations overrode it they'd have to track the point where it'd stop otherwise they'd be changing the behaviour from the reminder of the iterator's values to the iterator's values but reversed

hikari/utilities/iterators.py
nekokatt: not overly but if you prefer it like this then go for it.

hikari/utilities/iterators.py
nekokatt: the point is that the implementation detail remains consistent and doesn't impose a full implementation earlier than needed.

This way you can get the last 5 items using the same syntax: `async for item in it.reversed().limit(5):` thus remaining consistent.

If they want to just operate on a reversed list directly they should just run `list(reversed(await it))` to get that eagerly.

If you change this to a synchronous endpoint `def reverse()` that returns a lazy iterator, they can still await it as the same syntax to get the internal result, or they can continue to operate on it elsewhere.

Especially when I implement stuff like flatMap, this will be much more useful for streaming manipulations.

hikari/utilities/iterators.py
nekokatt: it is an iterator. By definition a python iterator gets "used up" by advancing.

Consume is probably what I meant, rather than advance, come to think of it.

PR #8, opened 2020-08-17
Changed event type for greetings.yml

nekokatt:


FasterSpeeding:
The greeting thing seems to be stuck, might be worth it just removing it

davfsa:
Looks like the reason this was getting stuck is because the secret was not set, but im fine with deleting it

PR #11, opened 2020-08-17
WIP: various gateway bugfixes and runtime bugfixes

nekokatt:
- Fixed regression preventing status updates.
- Fixed erraneous "update hikari" message on non-PyPI versions.
- Updated presence field types.

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -82,14 +82,19 @@ class GatewayShardImplV6(shard.GatewayShard):
         with each event that fires.
     http_settings : hikari.config.HTTPSettings
         The HTTP-related settings to use while negotiating a websocket.
-    initial_activity : hikari.utilities.undefined.UndefinedNoneOr[hikari.models.include_presences.Activity]
-        The initial activity to appear to have for this shard.
-    initial_idle_since : hikari.utilities.undefined.UndefinedNoneOr[datetime.datetime]
-        The datetime to appear to be idle since.
-    initial_is_afk : hikari.utilities.undefined.UndefinedOr[bool]
-        Whether to appear to be AFK or not on login.
-    initial_status : hikari.utilities.undefined.UndefinedOr[hikari.models.include_presences.Status]
-        The initial status to set on login for the shard.
+    initial_activity : typing.Optional[hikari.models.presences.Activity]
+        The initial activity to appear to have for this shard, or
+        `builtins.None` if no activity should be set initially. This is the
+        default.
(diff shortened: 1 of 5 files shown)


hikari/impl/shard.py
```diff
@@ -896,28 +884,54 @@ def _log_debug_payload(self, payload: str, message: str, *args: typing.Any) -> N
...
 
-        presence_pl.put("since", int(idle_since.timestamp() * 1_000) if idle_since is not None else None)
-        presence_pl.put("afk", afk)
+        return int(dt.timestamp() * 1_000)
 
-        # Turns out Discord don't document this properly. I can send "offline"
-        # to the gateway, but it will actually just result in the bot not
-        # changing the status. I have to set it to "invisible" instead to get
-        # this to work...
-        presence_pl.put("status", "invisible" if status is presences.Status.OFFLINE else status)
```
davfsa, line 918: Looks like you forgot to re-add this edge case

hikari/impl/shard.py
```diff
@@ -896,28 +884,54 @@ def _log_debug_payload(self, payload: str, message: str, *args: typing.Any) -> N
...
+    @staticmethod
+    def _serialize_activity(activity: typing.Optional[presences.Activity]) -> data_binding.JSONish:
```
davfsa, line 930: Doesn't this return `typing.Optional[data_binding.JSONish]`?

nekokatt: No, that wouldn't make sense

hikari/impl/shard.py
nekokatt: I believe this was just a red herring at the time. Otherwise I'd hope this would have been documented by now

hikari/impl/shard.py
davfsa: Code shows a bit under that if activity is none, it will return `None`:
```py
    @staticmethod
    def _serialize_activity(activity: typing.Optional[presences.Activity]) -> data_binding.JSONish:
        if activity is None:
            return None

        return {"name": activity.name, "type": int(activity.type), "url": activity.url}
```

hikari/impl/shard.py
davfsa: This is documented, you cant send `offline` to appear offline, you will have to send `invisible`

hikari/impl/shard.py
davfsa: Just realized `None` is in that Union, my bad :sweat_smile:

PR #12, opened 2020-08-17
Amended 'BotApp.run' to not reraise exceptions caused by graceful shutdown.

nekokatt:
This should reduce exception spam caused by invoking `await bot.close()` from a listener.

The application itself may still be raising other errors if you use a Windows proactor event loop, from what I have seen in other user's logs, but alas I cannot reproduce this currently.


FasterSpeeding:
lgtm

PR #13, opened 2020-08-17
Several cache optimisations and fixes

FasterSpeeding:
* Implement _chainable_get_voice_states_assets + bug fixes and remove unicode emojis from the internal unknown emoji cache
* Some refactors to avoid possible future issues
* fixes for lazy itertools.starmap calls not actually being called and a bug with lambda scoping
* Implement _chainable_get_presence_assets and _chainable_remove_presence_assets for the cache
* Expand out cache get_presences_view method to remove reliance on get_presences_view_for_guild
* Use dict.copy in the cache for mappings that are dicts instead of casting to a dict for a significant performance gain.
* add "copy" method to dm channel mru mapping

### Checklist


hikari/impl/stateful_cache.py
```diff
@@ -63,6 +62,15 @@
 _ValueT = typing.TypeVar("_ValueT")
 
 
+def _copy_mapping(mapping: typing.Mapping[_KeyT, _ValueT]) -> typing.MutableMapping[_KeyT, _ValueT]:
+    # dict.copy ranges from between roughly 2 times to 5 times more efficient than casting to a dict so we want to
+    # try to use this where possible.
+    try:
+        return mapping.copy()  # type: ignore[attr-defined, no-any-return]
+    except (AttributeError, TypeError):
```
nekokatt, line 70: When would this TypeError?

FasterSpeeding: if the "copy" method doesn't have the signature we're expecting here (assumption being made that if it has required args or kwargs then it's not the kind of copy method we were expecting)

hikari/impl/stateful_cache.py
```diff
@@ -575,20 +603,25 @@ class _PrivateTextChannelMRUMutableMapping(typing.MutableMapping[snowflake.Snowf
...
         /,
         *,
         expiry: datetime.timedelta,
     ) -> None:
         if expiry <= datetime.timedelta():
             raise ValueError("expiry time must be greater than 0 microseconds.")
 
-        self._data: typing.Dict[snowflake.Snowflake, _PrivateTextChannelData] = dict(source or ())
+        self._data: typing.Dict[
+            snowflake.Snowflake, _PrivateTextChannelData
```
nekokatt, line 615: Can we use a more meaningful name here....

maybe `channels`?

hikari/impl/stateful_cache.py
```diff
@@ -670,8 +703,7 @@ def __init__(self, app: traits.RESTAware, intents: typing.Optional[intents_.Inte
         self._invite_entries: typing.MutableMapping[str, _InviteData] = {}
         self._role_entries: typing.MutableMapping[snowflake.Snowflake, guilds.Role] = {}
         self._unknown_emoji_entries: typing.MutableMapping[
-            typing.Union[str, snowflake.Snowflake],
```
nekokatt, line 673: what makes emojis unknown? Comments for context?

hikari/impl/stateful_cache.py
```diff
@@ -792,7 +822,7 @@ def _clear_emojis(
     ) -> cache.CacheView[snowflake.Snowflake, emojis.KnownCustomEmoji]:
         emoji_ids: typing.Iterable[snowflake.Snowflake]
         if guild_id is undefined.UNDEFINED:
-            emoji_ids = tuple(self._emoji_entries.keys())
+            emoji_ids = _copy_mapping(self._emoji_entries).keys()
```
nekokatt, line 825: Do you even need a key view here? Can you just iterate over the dict directly?

hikari/impl/stateful_cache.py
```diff
@@ -863,23 +893,23 @@ def delete_emoji(self, emoji_id: snowflake.Snowflake, /) -> typing.Optional[emoj
     def get_emoji(self, emoji_id: snowflake.Snowflake, /) -> typing.Optional[emojis.KnownCustomEmoji]:
         return self._build_emoji(self._emoji_entries[emoji_id]) if emoji_id in self._emoji_entries else None
 
-    def _get_emojis_view(
+    def _get_emojis_view(  # TODO: split out?
```
nekokatt, line 896: Split out how?

If I had to take over from you, it should tell me what exactly the intent is :+1:

hikari/impl/stateful_cache.py
```diff
@@ -969,7 +999,9 @@ def get_guild(self, guild_id: snowflake.Snowflake, /) -> typing.Optional[guilds.
 
     def get_guilds_view(self) -> cache.CacheView[snowflake.Snowflake, guilds.GatewayGuild]:
         results = {
-            sf: guild_record.guild for sf, guild_record in tuple(self._guild_entries.items()) if guild_record.guild
+            sf: guild_record.guild
+            for sf, guild_record in _copy_mapping(self._guild_entries).items()
```
nekokatt, line 1003: When would it be falsy? Comments for context around stuff like this just to explain would make life 100x easier for anyone else who may need to work with this code in the future (just remember not everyone is as tuned in to how this all works as you are! Descriptions are really useful in dense classes of code!)

hikari/impl/stateful_cache.py
```diff
@@ -1022,19 +1054,24 @@ def clear_guild_channels_for_guild(
         if guild_record is None or guild_record.channels is None:
             return _EmptyCacheView()
 
-        cached_channels = {sf: self._guild_channel_entries.pop(sf) for sf in guild_record.channels}
```
nekokatt, line 1025: Can we explain in a comment what the side effect of this achieves?

hikari/impl/stateful_cache.py
```diff
@@ -1355,55 +1388,52 @@ def _can_remove_member(member: _MemberData, guild_record: _GuildRecord) -> bool:
...
-            }
-
-            if not guild_record.members:
-                guild_record.members = None
-                self._remove_guild_record_if_empty(guild_id)
-
+            #  This takes roughly half the time a two-layered for loop where we
+            #  assign to the members dict on every inner-iteration takes.
+            members_map = map(
+                lambda m: self._chainable_remove_member(m, guild_record), _copy_mapping(guild_record.members).values(),
```
nekokatt, line 1434: what does chainable do?

hikari/impl/stateful_cache.py
```diff
@@ -1415,19 +1445,13 @@ def clear_members_for_guild(
...
-            if member is not None
-        }
-
-        if not guild_record.members:
-            guild_record.members = None
-            self._remove_guild_record_if_empty(guild_id)
-
+        cached_members = _copy_mapping(guild_record.members)
+        # see https://github.com/python/mypy/issues/4297 for type ignore
+        members_map = map(
```
nekokatt, line 1450: So map is faster than generator expressions?

hikari/impl/stateful_cache.py
FasterSpeeding: technically that typing in the Generic wrapper should be an invariant typevar but it's kind of hard to explain what an unknown custom emoji with our current naming scheme

hikari/impl/stateful_cache.py
FasterSpeeding: a chainable method separates logic that's used in multiple view and clear methods during iterations and in some cases allow for chaining stuff within dict literals

hikari/impl/stateful_cache.py
FasterSpeeding: It's not quite as black and white as that and while a map along will be faster than a generator expression, introducing lambdas slows stuff down a considerable amount time (although if you're passing enough constant args the map will end up overtaking the generator expression but that's such a deep rabbit hole I don't want to think about it).
But the use of map specifically here is more an artefact of how i was doing stuff before so I'll probably switch that to a generator expression.

hikari/impl/stateful_cache.py
nekokatt: so like a linked list? 

as opposed to physically call chaing?

hikari/impl/stateful_cache.py
FasterSpeeding: yeah it'll work without, missed those when i was removing redundant .keys calls and will remove

nekokatt:
Unmark as draft and ping me when you are ready to review again based on what we discussed yesterday.

nekokatt:
are we not doing this now?

nekokatt (approved):
LGTM

PR #15, opened 2020-08-17
Added tests for Users

davfsa:
### Summary
  - Fixed broken docstrings
  - Fix black and isort fighting eachother

### Checklist

### DIFF
--- a/tests/hikari/models/test_users.py
+++ b/tests/hikari/models/test_users.py
@@ -19,8 +19,13 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import mock
+import pytest
 
 from hikari.models import users
+from hikari.utilities import constants
+from hikari.utilities import routes
+from hikari.utilities import undefined
+from tests.hikari import hikari_test_helpers
 
 
 def test_UserFlag_str_operator():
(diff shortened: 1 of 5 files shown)


tests/hikari/models/test_users.py
```diff
@@ -33,6 +38,123 @@ def test_PremiumType_str_operator():
...
+            id = 123
+            discriminator = "0001"
+            username = None
+            avatar_hash = None
+            is_bot = None
+            is_system = None
+            flags = None
+            mention = None
+
+        return hikari_test_helpers.unslot_class(StubUser)()
```
FasterSpeeding, line 55: is unslot the safest thing to use here? if we want to change class methods specifically couldn't we use the mock_class_namespace helper without discarding the slots and losing the safety those provide?

tests/hikari/models/test_users.py
```diff
@@ -33,6 +38,123 @@ def test_PremiumType_str_operator():
     assert str(type) == "NITRO_CLASSIC"
 
 
-def test_PartialUser_str_operator():
-    mock_user = mock.Mock(users.PartialUserImpl, username="thomm.o", discriminator="8637")
-    assert users.PartialUserImpl.__str__(mock_user) == "thomm.o#8637"
+class TestUser:
+    @pytest.fixture
+    def obj(self):
+        class StubUser(users.User):
```
FasterSpeeding, line 44: might want to look at the stub_class helper as well

tests/hikari/models/test_users.py
davfsa: The thing is that I don't want the functions I mock sometimes to be mocked. I just want to mock then when I need to, thus why I unslot

FasterSpeeding: Is it not possible to mock the name space for tests around properties that specifically need it, I'm not really a fan of the idea of in slotting objects everywhere for our tests because of the implications this has

tests/hikari/models/test_users.py
davfsa: I initially discarded stub_class because it instantiates the `user.User` class, which it cant because it has abstractmethods

tests/hikari/models/test_users.py
davfsa: Mocking the namespace later doesn't let me set specific attributes in each test

FasterSpeeding: Is that not more an issue of using an unrealistic class where variables are set on the class definition rather than as instance variables because slotting will make those read only, is it not possible to use a more realistic model that's possibly attrs based?

davfsa: The idea of using an instance of the actual object is that if, for example, someone opens a pr that changes where `avatar` is located to `UserImpl`, and we are using `UserImpl` to test this, then the tests won't fail, but they should.

Doing it this was just assures less cases when this could have a false positive.

FasterSpeeding: Well this is no longer an instance of the actual object, this is an instance of an object that inherits from it and overwrites all the abstractmethod with class attributes while throwing away the slotting which would normally make class attributes read only on objects for us. Not sure how that compares to making a stub attrs model where you'd at least get proper slotting.

davfsa: Solved

PR #16, opened 2020-08-18
Removed '__all__' from init

nekokatt:


hikari/__init__.py
```diff
@@ -27,12 +27,8 @@
 the REST API.
 """
 
-# We need these imported explicitly for the __all__ to be visible due to
-# Python's weird import visibility system.
-from hikari import config
-from hikari import errors
-from hikari import events
-from hikari import models
+from __future__ import annotations
```
davfsa, line 30: Do we really need these?

docs/patched_pdoc.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+def main():
+    from pdoc import cli
+
+    cli.main()
+
+
+if __name__ == "__main__":
+    main()
```
davfsa, line 9: Really curious as in what this is for. Also you are missing the license and `# cython: language_level=3`

hikari/__init__.py
nekokatt: Consistency, and reduces import time.

There is no reason not to include them.

PR #19, opened 2020-08-18
Implemented awaiting lazy iterator, fixed duplicated global ratelimit warning.

nekokatt:


hikari/impl/rate_limits.py
```diff
@@ -170,7 +170,7 @@ class ManualRateLimiter(BurstRateLimiter):
     Expect random occurrences.
     """
 
-    __slots__: typing.Sequence[str] = ()
+    __slots__: typing.Sequence[str] = ("_semaphore",)
```
nekokatt, line 173: Fix this.
```suggestion
    __slots__: typing.Sequence[str] = ()
```

hikari/utilities/iterators.py
```diff
@@ -624,6 +619,41 @@ def iter_mentioned_users(message: hikari.Message) -> typing.Iterator[Snowflake]:
...
+            input. MyPy cannot detect this nicely, so any cast is forced
+            internally.
+
+            If the item is not awaitable, you will receive a
+            `builtins.TypeError` instead.
+
+            You have been warned. You cannot escape the ways of the duck type
+            young grasshopper.
+        """
+        # Not type safe. Can I make this type safe?
```
FasterSpeeding, line 654: Shouldn't this be marked as TODO?

nekokatt: There isnt a way around this really

hikari/utilities/iterators.py
```diff
@@ -941,6 +975,9 @@ async def _generator(self) -> typing.AsyncIterator[AnotherValueT]:
         async for input_item in self._iterator:
             result_iterator = self._flattener(input_item)
 
+            if isinstance(result_iterator, typing.Iterable):
+                result_iterator = iter(result_iterator)
+
             if isinstance(result_iterator, typing.Iterator):
```
FasterSpeeding, line 981: Doesn't iterator subclass Iterable meaning this would never be true as iterators would be caught by the iterable check?

nekokatt: TIL.
```suggestion
            if isinstance(result_iterator, typing.Iterator):
```

FasterSpeeding (approved):
LGTM after that fix

PR #20, opened 2020-08-19
Bugfix/deploy

nekokatt:
Seems 2.0.0.dev69 failed to deploy for a few reasons.

- No verified email (not sure why this didn't work)
- Typos in deploy script


scripts/deploy.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/sh
+set -e
```
FasterSpeeding, line 2: just a question but what does `set -e` do here?

FasterSpeeding (approved):
lgtm

scripts/deploy.sh
nekokatt: if a command returns a non 0 exit code, it aborts the script with that exit code rather than continuing.

Shell defaults to "best effort", where it skips invalid commands.

Otherwise you'd have it where the shell would crash in interactive mode each time any command you run fails :P

PR #25, opened 2020-08-20
Added init pipeline

nekokatt:
Added a target `nox -s init` that installs everything anyone contribution may need.

Simply put, to start work on Hikari (assuming the correct Python is already installed):

- `git clone https://github.com/nekokatt/hikari`
- `pip install nox`
- `nox -s init`


davfsa (approved):
Really nice addition

PR #26, opened 2020-08-20
Flattened package as discussed for models and some utilities.

nekokatt:
Per the discussed changes from last week, I've purged the `models` package and moved some stuff around.


hikari/__init__.py
```diff
@@ -39,23 +39,33 @@
...
+from hikari.intents import *
+from hikari.invites import *
+from hikari.iterators import *
+from hikari.messages import *
+from hikari.permissions import *
+from hikari.presences import *
+from hikari.sessions import *
+from hikari.snowflakes import *
+from hikari.traits import *
+from hikari.undefined import *
```
davfsa, line 68: This imports the `count` function to the root project. Should be better have it how it was before

davfsa (changes requested):

davfsa (approved):

PR #29, opened 2020-08-20
Enable automatic dependency updates

nekokatt:
This will keep dependencies up to date by opening PRs if anything becomes outdated :-)


FasterSpeeding (approved):
Asides from some thoughts about whether we'd need to explicitly define allow dependency types here or not, lgtm

Issue #47, opened 2020-08-20
Bring coverage back up to 100%


davfsa:
Write all the missing tests

FasterSpeeding:
Make the coverage great again

CircuitSacul:
Ensure the codes integrity

PR #50, opened 2020-08-20
New tests

davfsa:
- Rename some properties for consistency
  - Rename `format_*` args for consitency

### Checklist

### Related issues
#47


hikari/guilds.py
```diff
@@ -645,12 +645,12 @@ def icon_url(self) -> typing.Optional[files.URL]:
...
         return self.format_icon()
 
-    def format_icon(self, *, format_: typing.Optional[str] = None, size: int = 4096) -> typing.Optional[files.URL]:
+    def format_icon(self, *, format: typing.Optional[str] = None, size: int = 4096) -> typing.Optional[files.URL]:
         """Generate the guild's icon, if set.
 
         Parameters
         ----------
-        format_ : typing.Optional[builtins.str]
+        format : typing.Optional[builtins.str]
```
nekokatt, line 653: Should we be shadowing builtins?

nekokatt (changes requested):
I think we should name `format` to something else to prevent builtin shadowing.

Maybe `ext` for file extension?

hikari/guilds.py
davfsa: I changed this for consistency since some where using `format` and some where using `format_`. Renaming to `ext` would be the best, will do that :+1:

FasterSpeeding: if we're going to abbreviate it, is there any reason for preferring "ext" or "fmt"?

davfsa: Don't really care which one, up to you guys, I would say maybe `ext`

PR #51, opened 2020-08-20
Adjusted error handling of events a little.

nekokatt:
Fixes an issue that @AnakiKaiver297 pointed out, where even if you have your own error handlers, the lib still spams exception tracebacks.

This changes it to only log ERROR level tracebacks if no listeners exist. If listeners exist, then it becomes DEBUG instead.


davfsa (approved):
LGTM

hikari/impl/event_manager_base.py
```diff
@@ -260,19 +260,24 @@ async def _invoke_callback(
...
-                    base_events.ExceptionEvent(
-                        app=self._app,
-                        shard=getattr(event, "shard") if isinstance(event, shard_events.ShardEvent) else None,
-                        exception=ex,
-                        failed_event=event,
-                        failed_callback=callback,
-                    )
+                exception_event = base_events.ExceptionEvent(
+                    app=self._app,
+                    shard=getattr(event, "shard") if isinstance(event, shard_events.ShardEvent) else None,
```
FasterSpeeding, line 271: do we need to use getattr here?

nekokatt: yes, because there is no guarantee the event has that attribute in the first place, so it would cause MyPy to complain.

FasterSpeeding: why would it cause mypy to complain when you've lowered the scope of the variable with an instance check?

PR #89, opened 2020-08-25
Bugfixes for race conditions in starting multisharded bots, and debug…

nekokatt:
… refactoring.

- Fixed bugs where if you sent a sigint or other signal before a shard connected but
  after starting the loop, it may get ignored leaving the bot in schrodingers box.
- Muted reaped task results that may have been raised from cancellations due to the
  event loop closing. Should fix some proactor event loop spam for Windows users too.
- Removed custom blocking time monitor and made the asyncio one able to be controlled
  in bot.run, this only is usable in debug mode now for performance.
- Moved customisation of coroutine tracking depth where supported to bot.run from
  bot.start.
- asyncio debug mode is now enabled in bot.run, not bot.start, meaning users who
  wish to have full control just can reimplement bot.run only now.
- Fixed documentation.

### Checklist

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -57,7 +57,6 @@ from hikari.impl import stateless_cache as stateless_cache_impl
 from hikari.impl import stateless_event_manager
 from hikari.impl import stateless_guild_chunker as stateless_guild_chunker_impl
 from hikari.impl import voice
-from hikari.utilities import aio
 from hikari.utilities import art
 from hikari.utilities import constants
 from hikari.utilities import date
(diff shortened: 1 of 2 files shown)


hikari/impl/bot.py
```diff
@@ -905,7 +924,7 @@ def _max_concurrency_chunker(self) -> typing.Iterator[typing.Iterator[int]]:
 
     async def _abort_shards(self) -> None:
         """Close all shards and wait for them to terminate."""
-        for shard_id in self._tasks:
+        for shard_id in self._shards:
```
nekokatt, line 927: Worst we can do is shut down a shard before it starts, which should just stop it from trying to start at all.

PR #90, opened 2020-08-27
Handling aiohttp.WSServerUpgradeError

nekokatt:
Discord appears to be sending 520 HTTP status codes from the gateway at certain times, with no info indicating what it is.

Appears to be something Cloudflare can spit out if Discord fails to communicate a response to them correctly, apparently.

Assuming this is just a regression or something on their side, so I have added a handler that can log this correctly.


davfsa (dismissed):
LGTM

davfsa (approved):

FasterSpeeding (approved):

PR #91, opened 2020-08-27
Defaulted state to being enabled by default on BotApp.

nekokatt:
State will now default to being __enabled__ on all applications unless explicitly turned off.


davfsa (approved):
This will need an announcement made along with all the changes planed for the next release. LGTM

davfsa (approved):

PR #95, opened 2020-08-27
Upgraded and baselined to black v20.8b1, which causes several major reformats.

nekokatt:


davfsa (approved):
Looks like all the parameters have to be on separate lines now, this might cause is to bump up the flake8 max line length. LGTM

PR #96, opened 2020-08-27
Bumped to pdoc 0.9.0.

### LINKED ISSUE #94: Bump pdoc3 version back to pypi releases
Pdoc3 V0.9 got released about 13h ago (at the time of writing this) which includes the lunr.js search that I madr.

This should be tested that it still builds and the search bar works correctly.

nekokatt:
Currently failing to build from pdoc issue `https://github.com/pdoc3/pdoc/issues/245`, so this
will need to be updated when that issue gets resolved or reverted.

Will close #94 once fixed.


davfsa:
The issue has been solved and this should now work.

Btw, since this is a bit in the scope of this pr, mind adding `slots` to the `BotAware` trait? Its missing them and a `weakref_slot` is being assigned to it

nekokatt:
> The issue has been solved and this should now work.
> 
> Btw, since this is a bit in the scope of this pr, mind adding `slots` to the `BotAware` trait? Its missing them and a `weakref_slot` is being assigned to it

On it.

PR #98, opened 2020-08-29
publicly exposed chunk tracking interface

FasterSpeeding:
### Summary
This overhalls the guild chunker interface to expose information publicly instead of just being a chunk event scheduler while adding extra utilitiy like "streaming" events and moved the scheduler over the shard object

### Checklist

- [X ] I have run `nox` and all the pipelines have passed.
- [ X] I have made unittests according to the code I have added/modified/deleted.


FasterSpeeding:
Opening this wip pr to see if I can get some feedback from ppl @nekokatt @davfsa

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,273 @@
...
 
 if typing.TYPE_CHECKING:
     from hikari import guilds
     from hikari import traits
+    from hikari import users as users_
+    from hikari.api import shard as gateway_shard
+    from hikari.events import base_events  # noqa F401 - Unused (False positive)
+
+
+EXPIRY_TIME = 5000
```
FasterSpeeding, line 55: : typing.Final[int]

hikari/utilities/mapping.py
```diff
@@ -0,0 +1,168 @@
...
+import typing
+
+KeyT = typing.TypeVar("KeyT", bound=typing.Hashable)
+"""Type-hint A type hint used for the type of a mapping's key."""
+ValueT = typing.TypeVar("ValueT")
+"""Type-hint A type hint used for the type of a mapping's value."""
+
+
+class MRUMutableMapping(typing.MutableMapping[KeyT, ValueT]):
+    """A most recently used mutable mapping implementation.
```
FasterSpeeding, line 44: """A most recently used limited mutable mapping implementation.

hikari/utilities/mapping.py
```diff
@@ -0,0 +1,168 @@
...
+        #  the end of the dict.
+        if key in self:
+            del self[key]
+
+        self._data[key] = (time.perf_counter(), value)
+        self._garbage_collect()
+
+
+class CMRIMutableMapping(typing.MutableMapping[KeyT, ValueT]):
+    """Implementation of a capacity most recently used mutable mapping.
```
FasterSpeeding, line 95: """Implementation of a capacity most recently used limited mutable mapping.

hikari/api/chunker.py
```diff
@@ -28,8 +28,61 @@
...
+    def chunk_count(self) -> typing.Optional[int]:
+        """Amount of chunks that are expected for this request."""
+
+    @property
+    def guild_id(self) -> snowflakes.Snowflake:
+        """Snowflake ID of the guild this chunk request is for."""
+
+    @property
+    def is_complete(self) -> bool:
+        """Whether this chunk event is finished or not.
```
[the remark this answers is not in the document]
FasterSpeeding: flake8 didn't complain about it for me

hikari/api/chunker.py
```diff
@@ -28,8 +28,61 @@
...
+
+        A chunk event may be considered finished after all chunks have been received
+        or after it's timed out.
+        """
+
+    @property
+    def last_received(self) -> typing.Optional[int]:
+        """Monotonic time of when we last received a chunk for this event in ms."""
```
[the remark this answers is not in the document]
FasterSpeeding: oh that's a typo, should be ns not ms

hikari/api/chunker.py
```diff
@@ -28,8 +28,61 @@
...
+
+    @property
+    def missing_chunks(self) -> typing.Optional[typing.Sequence[int]]:
+        """Sequence of the indexes of chunks we haven't received yet.
+
+        If this is `builtins.None` then we haven't received the initial chunk.
```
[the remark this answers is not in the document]
FasterSpeeding: seeing as we refer to them as indexs not ids, it'd be `missing_chunk_indexes`

hikari/api/chunker.py
```diff
@@ -38,7 +91,44 @@ class GuildChunker(abc.ABC):
     __slots__: typing.Sequence[str] = ()
 
     @abc.abstractmethod
-    async def request_guild_chunk(self, guild: guilds.GatewayGuild) -> None:
+    def fetch_members_for_guild(
```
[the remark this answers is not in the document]
FasterSpeeding: Chunk Information is a Protocol for a data class which is used represents a tracked chunk request

hikari/api/chunker.py
```diff
@@ -38,7 +91,44 @@ class GuildChunker(abc.ABC):
...
+    async def get_chunk_status(self, nonce: str) -> typing.Optional[ChunkInformation]:
+        ...
+
+    @abc.abstractmethod
+    async def list_chunk_statuses_for_shard(self, shard_id: int) -> typing.Sequence[ChunkInformation]:
+        ...
+
+    @abc.abstractmethod
+    async def list_chunk_statuses_for_guild(self, guild_id: snowflakes.Snowflake) -> typing.Sequence[ChunkInformation]:
+        ...
```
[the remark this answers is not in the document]
FasterSpeeding: that's on the chunker itself, not the chunk info object and while intents only allow one guild per chunk request, chunk requests can be queued and this will ultimately be tracking multiple events at a time as that's its main purpose

hikari/api/chunker.py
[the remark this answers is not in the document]
nekokatt: Put what you just said into a docstring then :P

hikari/utilities/mapping.py
nekokatt: hyphens

hikari/api/chunker.py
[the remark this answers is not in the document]
nekokatt: Also what does "event is finished" mean?

You mean request right?

hikari/api/chunker.py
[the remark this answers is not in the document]
FasterSpeeding: Tbf i was thinking about changing that to a date time and just storing the monotonic time along side a datetime just because datetimes are more user friendly

hikari/api/shard.py
```diff
@@ -374,7 +374,7 @@ async def request_guild_members(
             If not `builtins.None`, request the members which username starts with the string.
         limit: builtins.int
             Maximum number of members to send matching the query.
-        user_ids: hikari.undefined.UndefinedOr[typing.Sequence[hikari.snowflakes.SnowflakeishOr[hikari.users.User]]]
+        users: hikari.undefined.UndefinedOr[typing.Sequence[hikari.snowflakes.SnowflakeishOr[hikari.users.User]]]
```
FasterSpeeding, line 377: This should also be marked as a breaking change along with all the changes to the chunker

hikari/api/chunker.py
```diff
@@ -28,8 +28,131 @@
...
+            Whether this chunk request is considered to be finished or not yet.
+        """
+
+    @property
+    def last_received(self) -> typing.Optional[datetime.datetime]:
+        """Datetime of when we last received a chunk for this event.
+
+        Returns
+        -------
+        datetime.datetime or builtins.None
```
nekokatt, line 106: We are using typing.Optional over `or` now, as it doesn't break pdoc in some edge cases as much.

hikari/api/chunker.py
```diff
@@ -38,14 +161,168 @@ class GuildChunker(abc.ABC):
...
+        query: builtins.str
+            If not `builtins.None`, request the members which username starts with the string.
+        query_limit: builtins.int
+            Maximum number of members to send matching the query.
+        users: hikari.undefined.UndefinedOr[typing.Sequence[hikari.snowflakes.SnowflakeishOr[hikari.users.User]]]
+            If specified, the users to request for.
+
+        !!! note
+            To request the full list of members, set `query` to `builtins.None` or `""`
+            (empty string) and `limit` to 0.
```
nekokatt, line 200: Query is not optional so this is not right.

hikari/api/chunker.py
```diff
@@ -38,14 +161,168 @@ class GuildChunker(abc.ABC):
...
+            A stream of chunk events for the generated request.
+        """
+
+    @abc.abstractmethod
+    async def get_request_status(self, nonce: str) -> typing.Optional[RequestInformation]:
+        """Return the status of a request.
+
+        Parameters
+        ----------
+        nonce: str
```
nekokatt, line 223: these colons should start with spaces right

hikari/api/chunker.py
```diff
@@ -38,14 +161,168 @@ class GuildChunker(abc.ABC):
...
+        Returns
+        -------
+        typing.Sequence[RequestInformation]
+            A sequence of data objects of information about the tracked requests
+            for the given shard.
+        """
+
+    @abc.abstractmethod
+    async def list_requests_for_guild(
+        self, guild: snowflakes.SnowflakeishOr[guilds.GatewayGuild]
```
nekokatt, line 252: Is there a reason consume chunk event has a / in the signature but this does not?

hikari/api/chunker.py
```diff
@@ -38,14 +161,168 @@ class GuildChunker(abc.ABC):
...
+
+        Returns
+        -------
+        typing.Optional[RequestInformation]
+            Information about the request if found, else `builtins.None`.
+        """
+
+    @abc.abstractmethod
+    async def list_requests_for_shard(
+        self, shard: typing.Union[gateway_shard.GatewayShard, int]
```
nekokatt, line 234: Shardish type hint? :)

hikari/api/chunker.py
```diff
@@ -38,14 +161,168 @@ class GuildChunker(abc.ABC):
...
+            If specified, whether to request presences.
+        query: builtins.str
+            If not `builtins.None`, request the members which username starts with the string.
+        limit: builtins.int
+            Maximum number of members to send matching the query.
+        users: hikari.undefined.UndefinedOr[typing.Sequence[hikari.snowflakes.SnowflakeishOr[hikari.users.User]]]
+            If specified, the users to request for.
+
+        !!! note
+            To request the full list of members, set `query` to `builtins.None` or `""`
```
nekokatt, line 309: Same here, signature disallows passing None.

hikari/audit_logs.py
```diff
@@ -348,14 +348,7 @@ def __getitem__(self, slice_: slice, /) -> typing.Sequence[AuditLogEntry]:
...
     ) -> typing.Union[AuditLogEntry, typing.Sequence[AuditLogEntry]]:
-        if isinstance(index_or_slice, slice):
-            return tuple(
-                itertools.islice(self.entries.values(), index_or_slice.start, index_or_slice.stop, index_or_slice.step)
-            )
-        elif isinstance(index_or_slice, int):
-            return next(iter(itertools.islice(self.entries.values(), index_or_slice, None)))
-        else:
-            raise TypeError(f"sequence indices must be integers or slices, not {type(index_or_slice).__name__}")
+        return mapping.get_index_or_slice(self.entries, index_or_slice)
```
nekokatt, line 351: Just noticed the docstring on line 329 says the entries mapping is a sequence. Can we fix this? Ik it is outside the scope of this PR.

hikari/events/shard_events.py
```diff
@@ -236,3 +237,22 @@ class MemberChunkEvent(ShardEvent):
     typing.Optional[builtins.str]
         The request nonce if specified, or `builtins.None` otherwise.
     """
+
+    @typing.overload
+    def __getitem__(self, index_or_slice: int, /) -> guilds.Member:
```
nekokatt, line 242: Should this be called index instead?

hikari/impl/bot.py
```diff
@@ -606,6 +607,16 @@ def unsubscribe(
         # <<inherited docstring from event_dispatcher.EventDispatcher>>
         return self.dispatcher.unsubscribe(event_type, callback)
 
+    def stream(
+        self,
+        event_type: typing.Type[event_dispatcher.EventT_co],
```
nekokatt, line 612: Can you check if MyPy allows you to pass an abstract event class here for event_type?

We had issues with this elsewhere and had to defer to typing.Any as for some reason MyPy hates you passing abstract types around that get emplaced with generics.

hikari/impl/shard.py
```diff
@@ -235,6 +236,9 @@ def __init__(
     ) -> None:
         self._activity: undefined.UndefinedNoneOr[presences.Activity] = initial_activity
         self._backoff = rate_limits.ExponentialBackOff(base=1.85, maximum=600, initial_increment=2)
+        # We limit member chunk requests to a stricter ratelimit of 60 requests per minute as to ensure that chunk
+        # requests don't single-handedly exhaust the request limit when being triggered on mass.
+        self._chunk_request_ratelimiter = rate_limits.WindowedBurstRateLimiter(f"guild chunking on {shard_id}", 60, 60)
```
nekokatt, line 241: This name should probably have ratelimit in it

hikari/impl/shard.py
```diff
@@ -430,31 +435,38 @@ async def request_guild_members(
         include_presences: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
         query: str = "",
         limit: int = 0,
-        user_ids: undefined.UndefinedOr[typing.Sequence[snowflakes.SnowflakeishOr[users.User]]] = undefined.UNDEFINED,
+        users: undefined.UndefinedOr[typing.Sequence[snowflakes.SnowflakeishOr[users_.User]]] = undefined.UNDEFINED,
         nonce: undefined.UndefinedOr[str] = undefined.UNDEFINED,
     ) -> None:
+        await self._chunk_request_ratelimiter.acquire()
+
```
nekokatt, line 442: If we raise an exception on line 445, do we need to unacquire this? Should it go after the validation checks to prevent acquiring time and not using it, maybe?

nekokatt: Do we check the nonce size if present?

sending bad data will kick the gateway connection

hikari/impl/stateful_event_manager.py
```diff
@@ -152,11 +152,21 @@ async def on_guild_create(self, shard: gateway_shard.GatewayShard, payload: data
...
-        if event.guild.is_large and (self._intents is None or self._intents & intents_.Intents.GUILD_MEMBERS):
-            await self._app.chunker.request_guild_chunk(event.guild)
+        members_declared: int = True
+        presences_declared: int = True
+        if self._intents is not None:
+            members_declared = self._intents & intents_.Intents.GUILD_MEMBERS
+            presences_declared = self._intents & intents_.Intents.GUILD_PRESENCES
+
+        # When intents are enabled discord will only send other member objects on the guild create
+        # payload if presence intents are also declared, so if this isn't the case then we also want
```
nekokatt, line 166: Is this a bug or "a feature"?

Seems like a dumb feature, if it is the latter. Perhaps they need to document this clearly?

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,298 @@
...
+        query_limit: int = 0,
+        query: str = "",
+        users: undefined.UndefinedOr[typing.Sequence[snowflakes.SnowflakeishOr[users_.User]]] = undefined.UNDEFINED,
+    ) -> None:
+        super().__init__(app=app, event_type=shard_events.MemberChunkEvent, limit=limit, timeout=timeout)
+        self._guild_id = guild_id
+        self._include_presences = include_presences
+        self._limit = query_limit
+        self._missing_chunks: typing.Optional[typing.MutableSequence[int]] = None
+        self._nonce = date.uuid()
```
nekokatt, line 102: didnt you allow a custom nonce somewhere else?

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,298 @@
...
+class _TrackedRequests:
+    """`hikari.api.chunker.RequestInformation` implementation."""
+
+    __slots__: typing.Sequence[str]
+
+    average_chunk_size: typing.Optional[int] = attr.ib(default=None)
+    chunk_count: typing.Optional[int] = attr.ib(default=None)
+    guild_id: snowflakes.Snowflake = attr.ib()
+    last_received: typing.Optional[datetime.datetime] = attr.ib(default=None)
+    _mono_last_received: typing.Optional[int] = attr.ib(default=None)
```
nekokatt, line 166: Is there any point making this private when the class is anyway

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,298 @@
...
+                nonce=self._nonce,
+            )
+
 
-_LOGGER = logging.getLogger("hikari.guild_chunker")
+@attr.s(init=True, kw_only=True, hash=False, eq=True, slots=True, weakref_slot=False)
+class _TrackedRequests:
+    """`hikari.api.chunker.RequestInformation` implementation."""
+
+    __slots__: typing.Sequence[str]
```
nekokatt, line 160: hmmmmm.

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,298 @@
...
+        )
+        chunks.not_found_ids = list(self.not_found_ids)
+        return chunks
+
+    @property
+    def is_complete(self) -> bool:
+        if self.received_chunks == self.chunk_count:
+            return True
+
+        return self._mono_last_received is not None and time.monotonic_ns() - self._mono_last_received > EXPIRY_TIME
```
nekokatt, line 184: Do we need nano precision here though

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,298 @@
...
-    """Guild chunker implementation."""
+    """Guild chunker implementation.
 
-    __slots__: typing.Sequence[str] = ("_app", "_presences", "_queues", "_chunkers")
+    Parameters
+    ----------
+    app: hikari.traits.BotAware
+        The object of the bot aware app this is bound to.
+    limit: builtins.int
+        The maximum amount of requests that this chunker should store information
```
nekokatt, line 213: Space before colons please :)

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,298 @@
...
 
-    def __init__(self, app: traits.ShardAware, intents: typing.Optional[intents_.Intents]):
+    __slots__: typing.Sequence[str] = ("_app", "_tracked")
+
+    def __init__(self, app: traits.BotAware, limit: int = 200) -> None:
         self._app = app
-        self._presences: bool = intents is None or bool(intents & intents_.Intents.GUILD_PRESENCES)
-        self._queues: typing.Dict[int, typing.List[int]] = {}
-        self._chunkers: typing.Dict[int, asyncio.Task[None]] = {}
+        self._tracked: typing.MutableMapping[int, typing.Dict[str, _TrackedRequests]] = mapping.CMRIMutableMapping(
```
nekokatt, line 221: Can we just use theCRMIMapping type as the hint

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,298 @@
...
-            future.cancel()
+            return copy.copy(self._tracked[shard_id].get(nonce)) if shard_id in self._tracked else None
+
+    async def list_requests_for_shard(
+        self, shard: typing.Union[gateway_shard.GatewayShard, int]
+    ) -> typing.Sequence[chunker.RequestInformation]:
+        shard_id = shard if isinstance(shard, int) else shard.id
+
+        if shard_id in self._tracked:
+            return tuple(copy.copy(chunk) for chunk in tuple(self._tracked[shard_id].values()))
```
nekokatt, line 272: Can we use dict.copy here for performance?

hikari/impl/stateful_guild_chunker.py
```diff
@@ -23,65 +23,298 @@
...
+            include_presences=self._default_include_presences(guild_id, include_presences),
+            limit=limit,
+            nonce=nonce,
+            query=query,
+            users=users,
+        )
+        return nonce
+
+    async def close(self) -> None:
+        return None
```
nekokatt, line 320: Should this just be "pass"

hikari/impl/stateless_guild_chunker.py
```diff
@@ -27,10 +27,29 @@
...
+    from hikari import snowflakes
+    from hikari import users as users_
+    from hikari.api import shard as gateway_shard
+    from hikari.events import shard_events
+
+
+class _EmptyStream(event_stream.Streamer["shard_events.MemberChunkEvent"]):
+    """An empty stream implementation used by the stateless chunker impl."""
+
+    async def close(self) -> None:
```
nekokatt, line 45: pass?

hikari/impl/stateless_guild_chunker.py
```diff
@@ -27,10 +27,29 @@
...
 if typing.TYPE_CHECKING:
     from hikari import guilds
+    from hikari import snowflakes
+    from hikari import users as users_
+    from hikari.api import shard as gateway_shard
+    from hikari.events import shard_events
+
+
+class _EmptyStream(event_stream.Streamer["shard_events.MemberChunkEvent"]):
+    """An empty stream implementation used by the stateless chunker impl."""
```
nekokatt, line 43: Is impl a word?

hikari/snowflakes.py
```diff
@@ -136,6 +141,28 @@ def __eq__(self, other: typing.Any) -> bool:
...
 
+def calculate_shard_id(
+    app_or_count: typing.Union[traits.ShardAware, int, str], guild: SnowflakeishOr["guilds.PartialGuild"]
+) -> int:
+    """Calculate the shard ID for a guild based on it's shard aware app or shard count.
+
+    Parameters
+    ----------
+    app_or_count : hikari.traits.ShardAware or builtins.int or builtins.str
+        The shard aware app of the current application or the integer count of
```
nekokatt, line 152: why str?

nekokatt: also typing.union pls :)

hikari/utilities/event_stream.py
```diff
@@ -0,0 +1,249 @@
...
+        **attrs: typing.Any,
+    ) -> typing.Union[EventStream[EventT], iterators.LazyIterator[EventT]]:
+        if self._active:
+            return super().filter(*predicates, **attrs)
+
+        new_conditions = self._map_predicates_and_attr_getters("filter", *predicates, **attrs)
+        if self._filters:
+            self._filters |= new_conditions
+        else:
+            self._filters = new_conditions
```
nekokatt, line 235: is this reachable? Wont line 233 suffice?

hikari/utilities/mapping.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def __setitem__(self, key: KeyT, value: ValueT) -> None:
+        self._data[key] = value
+        self._garbage_collect()
+
+
+def get_index_or_slice(
+    mapping: typing.Mapping[KeyT, ValueT], index_or_slice: typing.Union[int, slice]
+) -> typing.Union[ValueT, typing.Sequence[ValueT]]:
+    """Get a dict entry at a given index as if it's a sequence of it's values.
```
nekokatt, line 137: Mapping or dict?

Can we use "it is"

nekokatt (changes requested):

hikari/snowflakes.py
FasterSpeeding: As with snowflake that's just allowing a string for an integer value

hikari/impl/stateless_guild_chunker.py
FasterSpeeding: isn't returning None more direct and explicit than "pass"ing which could also be applied to places like iteration

hikari/api/chunker.py
FasterSpeeding: Should i also fix the docs i copied that from even though it's out of the scope of this pr

hikari/api/chunker.py
FasterSpeeding: Not sure, the lib's inconsistently using both styles rn

hikari/api/chunker.py
FasterSpeeding: idk where i'd put that

hikari/utilities/event_stream.py
FasterSpeeding: it is reachable as i added a __bool__ override to All which returns False if no conditions are set else True but line 233 would also suffice so eh

hikari/api/chunker.py
FasterSpeeding: no

hikari/events/shard_events.py
FasterSpeeding: A slice isn't an index is it?

hikari/impl/bot.py
FasterSpeeding: should i also change the name for self._ratelimiter then?

hikari/impl/shard.py
FasterSpeeding: should i also change the name for self._ratelimiter then?

hikari/impl/shard.py
FasterSpeeding: if the nonce is over 32 byte characters long Discord will just ignore it, not kick the gateway connection

hikari/impl/stateful_event_manager.py
FasterSpeeding: It's a known feature

hikari/impl/stateful_guild_chunker.py
FasterSpeeding: Nonces generated by the main StatefulGuildChunker implementation will follow the format "{shard_id}.{actual_nonce}" while nonces generated by the Chunk Stream will just be a uuid, didn't allow custom nonces anywhere as the StatefulGuildChunker includes metadata that's useful in the nonce and the Chunk Stream is less open than the guild chunker

hikari/impl/stateful_guild_chunker.py
FasterSpeeding: The class may be private but the object itself is exposed in places, for reference they should be referring to the protocol.

hikari/impl/stateful_guild_chunker.py
FasterSpeeding: If that isn't included __slots__ will be treated as something like `typing.Union[typing.Iterable[str], str]]` which isn't compatible with the protocol which types __slots__ as typing.Sequence[str]

hikari/impl/stateful_guild_chunker.py
FasterSpeeding: This was just to avoid floating point math, it's either get the nano second one as an integer or get the normal one as a float which still includes nano seconds in the implementations of python i've seen

hikari/impl/stateful_guild_chunker.py
FasterSpeeding: Is there any actual reason to doing that when for all intensive purposes the CRMIMapping is still a normal MutableMapping interface wise

hikari/api/chunker.py
nekokatt: maybe in GatewayShard? Never mind

hikari/api/chunker.py
nekokatt: if you want ig

hikari/api/chunker.py
nekokatt: it shouldn't be, cause `:` without a preceding space broke pdoc3 last i checked

hikari/impl/stateless_guild_chunker.py
nekokatt: why do you need to be explicit about one line of code.

It is already obvious enough if you know basics of how python works. If you don't know that functions return None if no return is given, then you probably won't be reading this code.

If you start enforcing this, you'll have to now change every procedure in the API to also do this at the end if they don't explicitly return, just to make it consistent, which I don't think I am a fan of personally.

hikari/snowflakes.py
nekokatt: Why would this ever be a string? We store this globally as an int as soon as we receive it and never pass it back to the API explicitly as a string. It won't ever be more than 2^32 either unless you have majorly done something wrong, so there is no need to complicate this further really.

hikari/utilities/event_stream.py
```diff
@@ -0,0 +1,244 @@
...
+    ) -> None:
+        await self.close()
+
+    def __enter__(self) -> typing.NoReturn:
+        # This is async only.
+        cls = type(self)
+        raise TypeError(f"{cls.__module__}.{cls.__qualname__} is async-only, did you mean 'async with'?") from None
+
+    def __exit__(self, exc_type: typing.Type[Exception], exc_val: Exception, exc_tb: types.TracebackType) -> None:
+        return
```
nekokatt, line 125: Shouldn't this be `return None` or `pass`?

hikari/api/chunker.py
FasterSpeeding: i don't ever remember one style breaking pdoc3

hikari/utilities/event_stream.py
```diff
@@ -0,0 +1,244 @@
...
+        return self._await_all().__await__()
+
+    def __del__(self) -> None:
+        # For the sake of protecting highly intelligent people who forget to close this, we register the event listener
+        # with a weakref then try to close this on deletion. While this may lead to their consoles being spammed, this
+        # is a small price to pay as it'll be way more obvious what's wrong than if we just left them with a vague
+        # ominous memory leak.
+        if self._active:
+            _LOGGER.warning("active %r streamer fell out of scope before being closed", self._event_type.__name__)
+            asyncio.ensure_future(self.close())
```
nekokatt, line 210: We should check that the event loop is running first.

If it isn't running, this will spit out random errors, and you don't really care if the loop isn't running that you don't close this properly as the consumer will be dead anyway.

FasterSpeeding: What errors will it spit out if the loop isn't running?

hikari/utilities/mapping.py
```diff
@@ -0,0 +1,168 @@
...
+    ------
+    TypeError
+        If `index_or_slice` isn't a `builtins.slice` or `builtins.int`.
+    IndexError
+        If `index_or_slice` is an int and is outside the range of the mapping's
+        contents.
+    """
+    if isinstance(index_or_slice, slice):
+        return tuple(itertools.islice(mapping.values(), index_or_slice.start, index_or_slice.stop, index_or_slice.step))
+    elif isinstance(index_or_slice, int):
```
nekokatt, line 162:
```suggestion
    if isinstance(index_or_slice, int):
```

hikari/utilities/mapping.py
```diff
@@ -0,0 +1,168 @@
...
+        try:
+            return next(itertools.islice(mapping.values(), index_or_slice, None))
+        except StopIteration:
+            raise IndexError(index_or_slice) from None
+    else:
+        raise TypeError(f"sequence indices must be integers or slices, not {type(index_or_slice).__name__}")
```
nekokatt, line 168:
```suggestion
    raise TypeError(f"sequence indices must be integers or slices, not {type(index_or_slice).__name__}")
```

tests/hikari/hikari_test_helpers.py
```diff
@@ -149,21 +149,22 @@ def mock_class_namespace(
...
+    if implement_abstract_methods and hasattr(klass, "__abstractmethods__"):
+        for method_name in klass.__abstractmethods__:
             if method_name in namespace:
                 continue
 
             attr = getattr(klass, method_name)
 
             if inspect.isdatadescriptor(attr) or inspect.isgetsetdescriptor(attr):
-                namespace[method_name] = mock.MagicMock(__isabstractmethod__=False)
+                namespace[method_name] = mock.PropertyMock(__isabstractmethod__=False)
```
nekokatt, line 160: I used MagicMock on purpose at the time, I originally used propertymock and had issues with it. Can't remember what they were now so ig it is fine.

FasterSpeeding: This is might've been that trying to do obj.PropertyMock() = "foo" won't override the property mock which seems like sane behaviour to me

tests/hikari/impl/test_stateful_guild_chunker.py
```diff
@@ -0,0 +1,628 @@
...
+        stream._queue.put_nowait(mock_chunk)
+
+        async for event in stream:
+            assert event is mock_chunk
+            return
+
+        pytest.fail("stream should've yielded something")
+
+    @pytest.mark.asyncio
+    @hikari_test_helpers.timeout(0.5)
```
nekokatt, line 159: 0.5s is a little small; gotta remember most systems are not hard realtime, so if something happened on travis for example, you might find the systems pause for a few moments before continuing, which might fail this test.

Just something to be aware of, it is probably fine.

tests/hikari/impl/test_stateful_guild_chunker.py
```diff
@@ -0,0 +1,628 @@
...
+
+    @pytest.mark.asyncio
+    async def test___anext___waits_for_initial_chunk(self):
+        stream = hikari_test_helpers.stub_class(
+            stateful_guild_chunker.ChunkStream, _active=True, _queue=asyncio.Queue(), _missing_chunks=None, _timeout=0.3
+        )
+        mock_chunk = object()
+
+        async def add_chunk():
+            await asyncio.sleep(0.2)
```
nekokatt, line 176: is this necesarry? If it didn't wait for the chunk it would error right

tests/hikari/test_snowflake.py
```diff
@@ -120,3 +122,21 @@ class NekoUnique2(snowflakes.Unique):
...
+    assert snowflakes.calculate_shard_id(shard_count, guild_id) == expected_id
+
+
+@pytest.mark.parametrize(
+    ("guild_id", "expected_id"),
+    [(140502780547694592, 2), ("115590097100865541", 5), (snowflakes.Snowflake(105785483455418368), 7)],
+)
+def test_calculate_shard_id_with_app(guild_id, expected_id):
+    mock_app = mock.Mock(bot.BotApp, shard_count=8)
+    assert snowflakes.calculate_shard_id(mock_app, guild_id) == expected_id
```
nekokatt, line 142: what if it isn't a bot app? (i.e. called from a model that might be on a rest-only app)

FasterSpeeding: it's typed as taking ShardAware and if they gave it the wrong type it'd just error as expected

hikari/snowflakes.py
FasterSpeeding: it shouldn't need a special case to justify arbitrarily allowing a string for an integer value in some places but not others

tests/hikari/utilities/test_event_stream.py
```diff
@@ -0,0 +1,327 @@
...
+        mock_streamer = hikari_test_helpers.mock_class_namespace(event_stream.Streamer)
+        async with mock_streamer():
+            mock_streamer.open.assert_called_once()
+            mock_streamer.close.assert_not_called()
+
+        mock_streamer.open.assert_called_once()
+        mock_streamer.close.assert_called_once()
+
+    def test___enter__(self):
+        mock_streamer = hikari_test_helpers.mock_class_namespace(event_stream.Streamer)
```
nekokatt, line 48: can this be a fixture

tests/hikari/utilities/test_event_stream.py
```diff
@@ -0,0 +1,327 @@
...
+
+class TestStreamer:
+    @pytest.mark.asyncio
+    async def test___aenter___and___aexit__(self):
+        mock_streamer = hikari_test_helpers.mock_class_namespace(event_stream.Streamer)
+        async with mock_streamer():
+            mock_streamer.open.assert_called_once()
+            mock_streamer.close.assert_not_called()
+
+        mock_streamer.open.assert_called_once()
```
nekokatt, line 44: this is always true.

Maybe you wanted `assert len(mock_streamer.open.calls) == 1, "bang"` instead?

tests/hikari/utilities/test_event_stream.py
```diff
@@ -0,0 +1,327 @@
...
+from hikari.utilities import event_stream
+from tests.hikari import hikari_test_helpers
+
+
+class TestStreamer:
+    @pytest.mark.asyncio
+    async def test___aenter___and___aexit__(self):
+        mock_streamer = hikari_test_helpers.mock_class_namespace(event_stream.Streamer)
+        async with mock_streamer():
+            mock_streamer.open.assert_called_once()
```
nekokatt, line 41: don't you want this assert_awaited_once?

tests/hikari/utilities/test_event_stream.py
```diff
@@ -0,0 +1,327 @@
...
+
+    def test___enter__(self):
+        mock_streamer = hikari_test_helpers.mock_class_namespace(event_stream.Streamer)
+
+        # flake8 gets annoyed if we use "with" here so here's a hacky alternative
+        with pytest.raises(TypeError):
+            mock_streamer().__enter__()
+
+    def test___exit__(self):
+        hikari_test_helpers.mock_class_namespace(event_stream.Streamer)().__exit__(object(), object(), object())
```
nekokatt, line 55: i thought this helper returned an instance, what does calling it do then?

nekokatt, line 55: also this test doesn't actually test anything at all...

tests/hikari/utilities/test_event_stream.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+@pytest.fixture()
+def mock_app():
+    return mock.Mock(bot.BotApp)
```
nekokatt, line 60: spec set?

tests/hikari/utilities/test_event_stream.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+class TestEventStream:
+    @pytest.mark.asyncio
+    async def test__listener_when_filter_returns_false(self, mock_app):
+        stream = event_stream.EventStream(mock_app, events.Event, timeout=None)
+        stream.filter(lambda _: False)
+        mock_event = mock.Mock(events.Event)
```
nekokatt, line 68: speeeec set

tests/hikari/utilities/test_event_stream.py
FasterSpeeding: called once will error if it has multiple calls

tests/hikari/utilities/test_event_stream.py
FasterSpeeding: I thought we weren't using spec set for the sake of performance?

tests/hikari/utilities/test_mapping.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def test_get_index_or_slice_with_index_within_range():
+    result = mapping.get_index_or_slice({"i": "e", "n": "o", "b": "a", "hikari": "Rei", "p": "a", "o": "o"}, 3)
+    assert result == "Rei"
+
+
+def test_get_index_or_slice_with_index_outside_range():
+    with pytest.raises(IndexError):
+        mapping.get_index_or_slice({"i": "e", "n": "o", "b": "a", "hikari": "noa"}, 77)
```
nekokatt, line 150: require at least one abridged reference here to get approval.

tests/hikari/utilities/test_event_stream.py
FasterSpeeding: it tests that __exit__ exists and doesn't error out as doing `with` on an object that has an enter but no exit will raise an attribute error

FasterSpeeding: it returns the class, not an instance

hikari/utilities/event_stream.py
nekokatt:
```py
>>> import asyncio
>>> 
>>> async def foo():
...   pass
... 
>>> asyncio.run(foo())
>>> 
>>> asyncio.ensure_future(foo())
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/asyncio/tasks.py", line 660, in ensure_future
    loop = events.get_event_loop()
  File "/usr/lib/python3.8/asyncio/events.py", line 639, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'MainThread'.
```

FasterSpeeding: there's no good way to check in that situation, you'd get the exact same behaviour from asyncio.get_event_loop so best just to catch there

hikari/snowflakes.py
nekokatt: If "the API requires it" is classed as an unjustifiable special case, then sure.

Shard IDs are not snowflakes, they are 32 bit integers (maybe even 16 bit integers). Snowflakes are always exposed as 64 bit values which break IEEE double precision floating point literals, which is why the API allows it. 

Since Hikari allows you to directly interface with the REST or gateway API as you see fit, it makes sense to allow both formats. However, the gateway never provides these as strings, nothing uses them as strings, so all it does is allow more than one way to specify the ID here without any benefit (in fact, it just adds a tiny amount of overhead instead potentially).

Up to you though I guess.

tests/hikari/utilities/test_mapping.py
FasterSpeeding: wdym?

hikari/snowflakes.py
FasterSpeeding: > If "the API requires it" is classed as an unjustifiable special case, then sure.
isn't the whole point of an api wrapper to obfuscate off weird behaviour like that?

tests/hikari/utilities/test_event_stream.py
nekokatt: add `assert True, "expect this to always be reached without an AttributeError"` or
```py
try:
    ...
except AttributeError as ex:
    pytest.fail(ex)
```
so it is clear nothing is missed.

The latter is preferable since Pytest will distinguish between a failed test and an unhandled error that isn't an assertionerror.

nekokatt: ah i see. Can this mocked class be a fixture then?

Stick `@pytest.fixture(scope="module")` on it and it will only ever be generated once.

FasterSpeeding: `assert True, "expect this to always be reached without an AttributeError"` would annoy flake8 and if you insist then sure

hikari/utilities/event_stream.py
nekokatt: my point is, that RuntimeError should be caught.

```suggestion
            try:
                asyncio.ensure_future(self.close())
            except RuntimeError:
                # event loop has closed, nothing can be streamed now.
                ...
```

hikari/snowflakes.py
nekokatt: If that is the case, that is an argument for _not_ adding this.

The only reason it was added in the first place was because originally we had the REST component only allowing strings for IDs and the higher level wrapper allowing anything else.

I am more than happy to support completely removing string support for IDs as well.

tests/hikari/utilities/test_event_stream.py
nekokatt: create_autospec\*

tests/hikari/utilities/test_event_stream.py
FasterSpeeding: yeah it can

tests/hikari/impl/test_stateful_guild_chunker.py
FasterSpeeding: it's a passive listener, not active so that isn't quite the case and timing out is covered by the handles timeout test

hikari/impl/stateless_guild_chunker.py
FasterSpeeding: Is there any specific reason to use pass rather than `...` there either?

hikari/snowflakes.py
FasterSpeeding: I'll remove int from that then

tests/hikari/utilities/test_event_stream.py
FasterSpeeding: is there a need for that here?

hikari/impl/stateless_guild_chunker.py
nekokatt: not really. I tend to use `...` if it is detail that is implemented elsewhere or that should be overridden (i.e. in an abc).

I always read `...` as `insert something in place of this [somehow]`, whereas `pass` I assume means `no detail`, but that is just my own personal pattern. Either-or is fine.

hikari/snowflakes.py
nekokatt: uhm...  keep int, it is a shard ID!

tests/hikari/impl/test_stateful_guild_chunker.py
nekokatt: there is a thing in the helpers class that waits for tiny things like this that might be useful.

tests/hikari/utilities/test_event_stream.py
nekokatt: depends if you are using stuff on it or not ig

hikari/snowflakes.py
FasterSpeeding: Meant to say "str" there lol, sry

hikari/impl/shard.py
FasterSpeeding: I've moved the acquire after the validation, and added a bytes length check for nonce, if there's not anything else for this comment I'm missing then you can mark this as resolved

hikari/events/shard_events.py
nekokatt: No, but this overload takes an index.

Usually this would be a bad thing to do but the / operator prevents you referring to these by name, so it will remain compatible

nekokatt (approved):

PR #110, opened 2020-08-31
Freeze requirements and add update flag to pipelines

davfsa:
### Summary
Thanks to us having @dependabot we can safely freeze requirements allowing us to make sure that the specific package will work on the specific version. This also makes pipelines run the same version of the extensions on all computers 

### Checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -34,4 +34,4 @@ mypy==0.782
 # DEPENDENCY CHECKING #
 #######################
 
-safety==1.9.0
+safety~=1.9.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,7 +3,7 @@
 ################
 
 # Mocks (stdlib ones change between versions of Python, so are not consistent in behaviour like the backport is).
-mock~=4.0.2
+mock==4.0.2
 
 # Py.test stuff.
 pytest==6.0.1
@@ -12,17 +12,17 @@ pytest-cov==2.10.1
 pytest-randomly==3.4.1
 
 # Coverage testing.
-coverage~=5.2.1
+coverage==5.2.1
 
 # Other stuff
-async-timeout~=3.0.1  # Used for timeouts in some test cases.
+async-timeout==3.0.1  # Used for timeouts in some test cases.
 
 #################
 # DOCUMENTATION #
 #################
 
-pdoc3~=0.9.1
-sphobjinv~=2.0.1
+pdoc3==0.9.1
+sphobjinv==2.0.1
 
 #################
 # TYPE CHECKING #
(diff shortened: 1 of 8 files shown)


flake8-requirements.txt
```diff
@@ -1,27 +1,27 @@
-flake8~=3.8.3
+flake8==3.8.3
```
nekokatt, line 1: My fix that lets people be able to use flake8 on more devices should be in 3.8.4, so we should keep an eye out.

dev-requirements.txt
```diff
@@ -34,4 +34,4 @@ mypy==0.782
 # DEPENDENCY CHECKING #
 #######################
 
-safety~=1.9.0
+safety==1.9.0
```
nekokatt, line 37: Don't freeze safety. We want this to be as new as possible without having to manually update it via automated PRs

flake8-requirements.txt
```diff
@@ -1,27 +1,27 @@
...
 
 # Plugins
 # Ref: https://github.com/DmytroLitvinov/awesome-flake8-extensions
 
-flake8-bandit~=2.1.2                    # runs bandit
```
nekokatt, line 6: Don't freeze bandit please, we want this to always pull in the newest version regardless of whether dependabot has updated it or not.

pipelines/twemoji.nox.py
```diff
@@ -24,5 +24,5 @@
 @nox.session(reuse_venv=True)
 def twemoji_test(session: nox.Session):
     """Brute-force test all possible Twemoji mappings for Discord unicode emojis."""
-    session.install("-e", ".", "requests")
+    session.install("-U", "-e", ".", "requests")
```
nekokatt, line 27: -Ue can be one flag

nekokatt (changes requested):
Looks good, just revert changes to flake8-bandit and safety.

Those rely on being as new as possible to capture new vulnerabilities.

flake8-requirements.txt
```diff
@@ -1,27 +1,27 @@
...
-flake8-coding~=1.3.2                    # coding magic-comment detectiong
-flake8-comprehensions~=3.2.3            # comprehension checks
-flake8-deprecated~=1.3                  # deprecated call checks
-flake8-docstrings~=1.5.0                # pydocstyle support
-flake8-executable~=2.0.4                # shebangs
-flake8-fixme~=1.1.1                     # "fix me" counter
-flake8-functions~=0.0.4                 # function linting
-flake8-html~=0.4.1                      # html output
-flake8-if-statements~=0.1.0             # condition linting
+flake8-bandit==2.1.2                    # runs bandit
```
nekokatt, line 6: Don't freeze bandit please, we want this to always pull in the newest version regardless of whether dependabot has updated it or not.

```suggestion
flake8-bandit~=2.1.2                    # runs bandit
```

PR #111, opened 2020-08-31
Remove, narrow and fixup some guild related fields

FasterSpeeding:
### Summary
* Removed dead max presences and members values from gateway guild.
* Fixed consistency for approx active member field naming between models.
* Narrow the scope of "with_count" only rest guild attributes to make them always required

### Checklist

- [x ] I have made unittests according to the code I have added/modified/deleted.


hikari/utilities/constants.py
```diff
@@ -81,4 +81,7 @@
 TWEMOJI_PNG_BASE_URL: typing.Final[str] = "https://github.com/twitter/twemoji/raw/master/assets/72x72/"
 TWEMOJI_SVG_BASE_URL: typing.Final[str] = "https://github.com/twitter/twemoji/raw/master/assets/svg/"
 
+# Miscellaneous
+DEFAULT_MAX_PRESENCES: typing.Final[int] = 25000
```
nekokatt, line 85: I feel like we should ask Discord to provide this, because I feel uncomfortable with this potentially changing in the future without warning

PR #117, opened 2020-09-03
Added optional `pip install hikari[speedups]` to get compiled modules.

### LINKED ISSUE #116: ciso8601 needs visual c++ redist installed
We need to make ciso8601 optional, since it cannot be installed on Windows if you lack c++ redist currently...

nekokatt:
Without this, defaults are to use pure-Python implementations of stuff
where possible.

This fixes an issue where people without Visual C++ Redist 14.0 could not
install Hikari v2.0.0.dev74.

Fixes #116.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ _test_job: &test_job
     - ${PYTHON_COMMAND} -V
   script:
     - ${PYTHON_COMMAND} -m nox --sessions pytest
+    # Use --cov-append to make sure coverage covers speedup and non-speedup execution paths equally.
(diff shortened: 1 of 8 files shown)


davfsa (approved):

davfsa (approved):

davfsa:
I was wondering, is there a way to default it to try to install speedups and, if it fails, to fall back to the pure-python implementation?

nekokatt:
> I was wondering, is there a way to default it to try to install speedups and, if it fails, to fall back to the pure-python implementation?

I'll do this in the future maybe. Compiling extensions before failing can be really slow, so ideally it should be opt-in only.

davfsa (approved):

PR #119, opened 2020-09-04
Proper shutting down

davfsa:
- Fixed some other bugs I found

### Checklist


hikari/impl/shard.py
```diff
@@ -741,6 +746,7 @@ async def _run_once(self) -> bool:
                     # to always kill our connection if it dies.
                     # We return True if zombied
                     if await heartbeat_task:
+                        now = date.monotonic()
```
nekokatt, line 749: What is this for?

davfsa: `now` might be undefined, since it only gets defined in L730. This just makes sure its defined.

nekokatt (approved):

nekokatt (approved):

PR #121, opened 2020-09-04
Task/gateway simplification

nekokatt:
This isn't finished yet, and I know tests are failing, static typing is wrong, linters are screaming, and fundamental error handling and functionality is missing.


hikari/utilities/aio.py
```diff
@@ -36,6 +37,7 @@
 
 T_co = typing.TypeVar("T_co", covariant=True)
 T_inv = typing.TypeVar("T_inv")
+U_co = typing.TypeVar("U_co", covariant=True)
```
nekokatt, line 40: This is redundant, and needs removing.

hikari/utilities/aio.py
```diff
@@ -27,6 +27,7 @@
     "completed_future",
     "is_async_iterator",
     "is_async_iterable",
+    "first_completed",
```
nekokatt, line 30: This is missing functions.

hikari/api/shard.py
```diff
@@ -392,3 +257,6 @@ async def request_guild_members(
             When trying to request presences without the `GUILD_MEMBERS` or when trying to
             request the full list of members without `GUILD_PRESENCES`.
         """
+
+    def __await__(self) -> typing.Generator[None, typing.Any, typing.Any]:
```
nekokatt, line 261: Might replace this with explicit calls to join. This was me being partially lazy, as it is easier to do `asyncio.gather(*bot.shards.values())` than `asyncio.gather(*(s.join() for s in bot.shards.values()))`

tests/hikari/utilities/test_aio.py
```diff
@@ -192,3 +193,103 @@ async def __anext__(self):
                 return ...
 
         assert not aio.is_async_iterable(AsyncIterator)
+
+
+@pytest.mark.asyncio
+class TestFirstCompleted:
```
nekokatt, line 199: Need tests for all_of.

tests/hikari/impl/test_shard.py
```diff
@@ -1210,7 +1210,7 @@ async def test_when_opcode_is_HEARTBEAT_ACK(self, client, exit_error):
             "d": "some data",
         }
         client._receive_json = mock.AsyncMock(side_effect=[payload, exit_error])
-        client._last_heartbeat_sent = 5
+        client._last_sent = 5
```
nekokatt, line 1213: Refactoring mistake, but pretty much all of this file is now useless again, so meh. 

Will have to rewrite all of this as part of this PR, when I get time.

hikari/utilities/aio.py
```diff
@@ -89,3 +91,78 @@ def is_async_iterable(obj: typing.Any) -> bool:
...
+    ----------
+    *aws : typing.Awaitable[typing.Any]
+        Awaitables to wait for.
+    timeout : typing.Optional[float]
+        Optional timeout to wait for, or `builtins.None` to not use one.
+        If the timeout is reached, all awaitables are cancelled immediately.
+
+    Returns
+    -------
+    typing.Tuple[typing.Any, ...]
```
nekokatt, line 156: return types are wrong.

hikari/errors.py
```diff
@@ -87,22 +87,23 @@ class HikariWarning(RuntimeWarning):
     """
 
 
+@attr_extensions.with_copy
 @attr.s(auto_exc=True, slots=True, repr=False, weakref_slot=False)
```
FasterSpeeding, line 91: Isn't copy inherited here anyway?

hikari/impl/bot.py
```diff
@@ -23,589 +23,481 @@
...
             executor=self._executor,
             http_settings=self._http_settings,
             proxy_settings=self._proxy_settings,
-            token=token,
-            token_type=constants.BOT_TOKEN,  # nosec
             rest_url=rest_url,
-            version=rest_version,
+            token=token,
+            token_type=constants.BOT_TOKEN,
+            version=6,
```
FasterSpeeding, line 163: Random thought, would it not make sense for the rest client implementation to be tied to a version rather than having it take a version number that may break it between versions.

hikari/impl/bot.py
```diff
@@ -23,589 +23,481 @@
...
             entity_factory=self._entity_factory,
             executor=self._executor,
             http_settings=self._http_settings,
             proxy_settings=self._proxy_settings,
-            token=token,
-            token_type=constants.BOT_TOKEN,  # nosec
             rest_url=rest_url,
-            version=rest_version,
+            token=token,
+            token_type=constants.BOT_TOKEN,
```
FasterSpeeding, line 162: BOT_TOKEN seems a bit misleading/vague compared to something like BOT_TOKEN_PREFIX

hikari/impl/bot.py
```diff
@@ -23,589 +23,481 @@
...
     @property
     def executor(self) -> typing.Optional[concurrent.futures.Executor]:
-        # <<inherited docstring from traits.ExecutorAware>>
         return self._executor
 
     @property
     def heartbeat_latencies(self) -> typing.Mapping[int, float]:
-        # <<inherited docstring from traits.ShardAware>>
-        return {shard_id: shard.heartbeat_latency for shard_id, shard in self._shards.items()}
+        return {s.id: s.heartbeat_latency for s in self._shards.values()}
```
FasterSpeeding, line 196: Should we do `self._shards.copy()` to avoid an edge case error caused by someone calling a method or property that iterates through self._shards while the bot is rolling up and creating shards objects?

hikari/impl/bot.py
```diff
@@ -23,589 +23,481 @@
...
-    @property
-    def rest(self) -> rest_client_impl.RESTClientImpl:
-        # <<inherited docstring from traits.RESTAware>>
-        return self._rest
-
     @property
     def shards(self) -> typing.Mapping[int, gateway_shard.GatewayShard]:
-        # <<inherited docstring from traits.ShardAware>>
-        return self._shards
+        return types.MappingProxyType(self._shards)
```
FasterSpeeding, line 221: do we need to build a MappingProxyType every time we call this property? MappingProxyType should change as the mapping is changed upstream shouldn't it?

hikari/impl/bot.py
```diff
@@ -23,589 +23,481 @@
...
+                candidate_shard_id
+                for candidate_shard_id in range(
+                    window_start, window_start + requirements.session_start_limit.max_concurrency
+                )
+                if candidate_shard_id in shard_ids
+            ]
+
+            if not window:
+                continue
+            if self._shards:
```
FasterSpeeding, line 292: is there a reason this check is run for every window_start, would this not wipe the previously started shards for this session if there's more than one window?

hikari/impl/bot.py
```diff
@@ -23,589 +23,481 @@
...
-        return date.monotonic() - self._started_at_monotonic if self._started_at_monotonic else 0.0
+            for started_shard in started_shards:
+                self._shards[started_shard.id] = started_shard
 
-    async def start(self) -> None:
-        """Determine sharding settings if needed, and start all shards.
+        await self.dispatch(lifetime_events.StartedEvent(app=self))
 
-        This method will first log if any library updates are available.
+    async def join(self, until_close: bool = True) -> None:
```
FasterSpeeding, line 337: What's the behaviour around this if join is called while it's in the middle of starting shards, will this leave some shards in an active state then?

hikari/impl/bot.py
```diff
@@ -23,589 +23,481 @@
...
-                gatherer = asyncio.gather(*window.values())
-                waiter = asyncio.create_task(self._request_close_event.wait(), name="listen for bot closure events")
+        await self.dispatch(lifetime_events.StoppingEvent(app=self))
 
-                await asyncio.wait((gatherer, waiter), return_when=asyncio.FIRST_COMPLETED)
+        calls = [
+            ("rest", self._rest.close()),
+            ("chunker", self._chunker.close()),
+            ("voice handler", self._voice.close()),
+            *((f"shard {s.id}", s.close()) for s in self._shards.values()),
```
FasterSpeeding, line 462: what if shards are added while this is being iterated over?

hikari/impl/bot.py
```diff
@@ -624,381 +515,80 @@ async def wait_for(
...
-            await asyncio.wait([gatherer, waiter], return_when=asyncio.FIRST_COMPLETED)
-
-            if not waiter.done():
-                waiter.cancel()
-        finally:
-            _LOGGER.debug("gather terminated, shutting down shard(s)")
-            aborter = asyncio.shield(self.close())
+    @staticmethod
+    def _destroy_loop(loop: asyncio.AbstractEventLoop) -> None:
+        async def murder(future: asyncio.Future[typing.Any]) -> None:
```
FasterSpeeding, line 560: would this not be the "murderer" not just "murder"

hikari/impl/shard.py
```diff
@@ -53,14 +46,246 @@
...
+    async def send_json(
+        self,
+        data: data_binding.JSONObject,
+        compress: typing.Optional[int] = None,
+        *,
+        dumps: aiohttp.typedefs.JSONEncoder = json.dumps,
+    ) -> None:
+        pl = dumps(data)
+        self._log_debug_payload(pl, "sending")
+        await super().send_str(pl, compress)
```
FasterSpeeding, line 134: does this have to be a super call?

hikari/impl/shard.py
```diff
@@ -120,99 +345,13 @@ class GatewayShardImplV6(shard.GatewayShard):
...
         at all.
-
         If any of these _are_ specified, then any that are not specified will
         be set to sane defaults, which may change the previous status. This will
         only occur during startup, and is an artifact of how Discord manages
         these updates internally. All other calls to update the status of
         the shard will support partial updates.
     """
 
-    __slots__: typing.Sequence[str] = (
```
FasterSpeeding, line 131: Do we not need slots?

hikari/impl/shard.py
```diff
@@ -325,110 +436,28 @@ def intents(self) -> typing.Optional[intents_.Intents]:
...
+        return self._run_task is not None and not self._run_task.done()
 
     @property
     def shard_count(self) -> int:
         return self._shard_count
 
-    @property
-    def version(self) -> int:
-        return self._version
+    async def close(self) -> None:
```
FasterSpeeding, line 445: Should we not close the ratelimiters here plus chunker and other stuff or am i missing something?

hikari/utilities/aio.py
```diff
@@ -89,3 +88,87 @@ def is_async_iterable(obj: typing.Any) -> bool:
...
+    timeout : typing.Optional[float]
+        Optional timeout to wait for, or `builtins.None` to not use one.
+        If the timeout is reached, all awaitables are cancelled immediately.
+
+    !!! note
+        If more than one awaitable is completed before entering this call, then
+        the first future is always returned.
+    """
+    fs = list(map(asyncio.ensure_future, aws))
+    iterator = asyncio.as_completed(fs, timeout=timeout)
```
FasterSpeeding, line 121: this is probably anothing but would this look any cleaner using asyncio.wait?

hikari/impl/shard.py
davfsa: No, I have already fixed this in my testing branch :+1:

hikari/impl/bot.py
davfsa: Will get changed :+1:

hikari/errors.py
nekokatt: You were the one that added these to every exception type.

image

I was just following the consistency you were trying to enforce when you implemented this.

If this isn't valid, can we please open an issue to adjust this, as it isn't in scope of this PR.

hikari/impl/bot.py
nekokatt: yep, that's why bot doesn't expose it directly now.

hikari/impl/bot.py
nekokatt: this would only occur if they did it from another thread, and the application isn't marked as being threadsafe, so I am not sure.

Threadsafety should probably go in a separate issue as I doubt this is the only place this may cause a headache later.

hikari/impl/bot.py
nekokatt: how else would you suggest doing this? I'd rather not expose shards in a map directly as altering that can lead to other implications such as leaking of connections and resources, and preventing graceful shutdown.

Creating a mapping proxy type is trivial anyway, it exists on a C level as it is part of the conduit of how classes work, so I don't see an issue using it like this really.

It is on the level that if you were having performance issues, you shouldn't be repeatedly calling the property either.

hikari/errors.py
FasterSpeeding: I just added it to the base exception classes we use

hikari/impl/bot.py
nekokatt: it murders a future.

hikari/impl/shard.py
nekokatt: Had this discussion with Dav, I don't think it provides much benefit, but this has been re-introduced for consistency.

hikari/utilities/aio.py
nekokatt: not really, asyncio.wait produces two lists you have to manually purge. I was trying to change this a few minutes ago but this was simpler as far as mypy was concerned, and I feel this represents what I am trying to do with it a little clearer.

hikari/impl/bot.py
FasterSpeeding: meaning it's a future murderer

hikari/impl/shard.py
nekokatt: Fixed.

hikari/impl/bot.py
nekokatt: it will wait for any shards that are running.

We use it during windowing to detect if any shards have died while we back off for 5s.

hikari/impl/shard.py
```diff
@@ -23,23 +23,16 @@
 
 from __future__ import annotations
 
-__all__: typing.Final[typing.List[str]] = [
```
FasterSpeeding, line 26: don't we want to have a \__all__ here?

hikari/impl/bot.py
```diff
@@ -23,589 +23,721 @@
 
 from __future__ import annotations
 
-__all__: typing.Final[typing.List[str]] = ["BotApp"]
+__all__: typing.List[str] = ["BotApp", "LoggerLevel"]
```
FasterSpeeding, line 26: typing.Final?

hikari/impl/bot.py
```diff
@@ -23,589 +23,721 @@
...
-    from hikari.api import chunker as guild_chunker_
-    from hikari.events import base_events
+    from hikari.api import cache
+    from hikari.api import chunker
+    from hikari.api import rest as rest_
+    from hikari.api import shard
     from hikari.impl import event_manager_base
-    from hikari.utilities import event_stream
+
+LoggerLevel = typing.Union[
```
FasterSpeeding, line 70: This feels a bit funky to me, being PascalCase but without T at the end of it to identify it as solely a type variable

nekokatt: fixed.

hikari/impl/bot.py
nekokatt: fixed.

hikari/impl/shard.py
nekokatt: fixed.

FasterSpeeding (approved):
Lgtm

PR #122, opened 2020-09-06
Ensure ExponentialBackOff respects float limitations.

FasterSpeeding:
* ensure guild_chunker.close is awaited.
* Fix accidental call to pytest.skip instead of pytest.mark.skip in test_mapping.py.

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ jobs:
       os: linux
       arch: amd64
       install: "pip install nox"
+      if: env(DISCORD_EMOJI_MAPPING_URL) IS present
       env:
         - secure: "BjD5A9s8GCC/ighdYW143rLeKxs+ltNB6ibzMCEcKi+TffvMJ9oLbNM0q1jTs+Jw9YOzIr07d9hOTbc41IVQBqjwNzqBQo37dTp/JPLl1bI579gfOZJ5POg0AT+FX9VpM2qYVxNPYASX/2v03EVmtLEAVXdg37NnUzbx4H5BBjei9tcUOlhm7Mt0o5s4tx/XyNxfDNtUah4rg6FCvhaBza2oApiBxR/3dWEr4/pFLXXqh8fCtUfLtgv+zt8P4VZxgWn33hsfxuqqBattbK+aOaARHDvQe73p2xy9INWgmiKNcM1lgpiFGsgSa4HpfYo/oUqP/dLULUocloIPTqMHOpWcQSRHACd0UUkOv6ZiB63NahQOxAxOfPpIGLdf6JSb2huPNGgmg6gqewKHmRcqWghgJfnXoKKdKK3tFG8+jBbw0Rmi8NHg5xJVsfUDt8utkRceBb13WcirTdam9r/kWo5Ogy+jEdbDX5R8OFimq+uzDdsjnI+BuakphLScxW6wo+iibDOkIulsKk6xhgIXzCi1uT2KJkIBkX3zxNdRlMtZS7rC6uz3xgF4I9o756CiT85vlKBaQroBj/2fYfU+d2wYwVFtFVGopqVpTaSQidZaAVZzJNWTXov/aOAK/e55VwEssu22keS/p7g2+6dogUJjO+qswupB+dvcnVJlnNk="
       stage: test
--- a/hikari/impl/rate_limits.py
+++ b/hikari/impl/rate_limits.py
@@ -438,25 +438,41 @@ class ExponentialBackOff:
     Parameters
     ----------
     base : builtins.float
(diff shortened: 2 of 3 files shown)


hikari/impl/bot.py
```diff
@@ -640,7 +640,7 @@ async def close(self) -> None:
             return
 
         self._has_aborted = True
-        self._guild_chunker.close()
+        await self._guild_chunker.close()
```
davfsa, line 643: Mind reverting this? Im aware its a bug, but this will get fixed with #121 (just fixes the merge conflicts there might be)

davfsa (changes requested):
LGTM, just one little thing

nekokatt:
Per what Dav said, I am retargeting this onto the branch I am working on, as most of this integration has moved, so it will be pointless to apply this to master just to lose it again :)

Sorry to trample like this, but would you mind just rechecking this now I have retargeted it?

hikari/impl/bot.py
nekokatt: See line 473 :)
```
        calls = [
            ("rest", self._rest.close()),
            ("chunker", self._chunker.close()),
            ("voice handler", self._voice.close()),
            *((f"shard {s.id}", s.close()) for s in self._shards.values()),
        ]

        for coro in asyncio.as_completed([handle(*pair) for pair in calls]):
            await coro
```

hikari/impl/rate_limits.py
```diff
@@ -438,25 +438,41 @@ class ExponentialBackOff:
...
+    ------
+    ValueError
+        If an `builtins.int` that's too big to be represented as a
+        `builtins.float` is passed in place of a field that's annotated as
+        `builtins.float` or if `initial_increment` is invalid for the provided
+        back-off rules (e.g. too large).
     """
 
-    __slots__: typing.Sequence[str] = ("base", "increment", "maximum", "jitter_multiplier")
+    __slots__: typing.Sequence[str] = ("_is_frozen", "base", "increment", "maximum", "jitter_multiplier")
```
nekokatt, line 461: When do we hit a floating point issue?

Can we use `decimal.Decimal` here instead?

nekokatt:
Instead of working around float being unusable, we could just use `decimal.Decimal` instead.

I am curious as to how long this is backing off for before a constraint is hit. I would hope it is some big number, but in that case, what has gone so catastrophically wrong that we didn't give up already?

hikari/impl/rate_limits.py
FasterSpeeding: decimal.Decimal also has maximum size limits on certain impls and this won't necessarily be bigger than the maximum floating point number limit

tests/hikari/impl/test_rate_limits.py
```diff
@@ -401,6 +402,33 @@ def test_is_rate_limited_when_rate_limit_not_expired_only_returns_False(self, re
 
 
 class TestExponentialBackOff:
+    def test___init___raises_on_too_large_int_base(self):
+        base = int(sys.float_info.max) + int(sys.float_info.max * 1 / 100)
+        with pytest.raises(ValueError, match="int too large to be represented as a float"):
```
nekokatt, line 407: why / 100, / 200, / 300?

nekokatt (approved):

davfsa (approved):

.travis.yml
```diff
@@ -53,6 +53,7 @@ jobs:
       os: linux
       arch: amd64
       install: "pip install nox"
+      if: env(DISCORD_EMOJI_MAPPING_URL) IS present
```
davfsa, line 56: This has been fixed as of #124, you can revert this

davfsa (changes requested):

.travis.yml
nekokatt: no. it is secured, so it needs to be here.

securing the variable stops him being able to expose it.

Otherwise people could steal my passwords by just opening PRs with modifications that just send the encrypted variables to places.

davfsa (approved):

tests/hikari/impl/test_rate_limits.py
FasterSpeeding: No particular reason, just wanted a number that was different enough to actually be different after rounding and didn't want to duplicate data between test cases

nekokatt: but... why would it overflow if you did less than `float.max * random() / 300` of the maximum?

surely it needs to be >= float.max to do that?

FasterSpeeding: Because the result of that is being added onto the max?

nekokatt: ah ok, was being stupid, sorry

PR #129, opened 2020-09-07
Added top role property to member

ahnaf-zamil:
### Checklist

### Related issues
This doesn't work stateless without cache. This is why the nox tests couldn't be passed. Also I don't know how to write tests. So I need someone to check the code and see if it has issues.

### DIFF
--- a/hikari/guilds.py
+++ b/hikari/guilds.py
@@ -421,6 +421,21 @@ class Member(users.User):
         """
         return f"<@!{self.id}>" if self.nickname is not None else self.user.mention
 
+    @property
+    def top_role(self) -> Role:
+        """Returns the top role of the given member.
+
+        Returns
+        -------
+        hikari.guilds.Role
+            The highest role the member has.
+        """
+        roles = [r for r in self.app.cache.get_roles_view_for_guild(self.guild_id).values() if r.id in self.role_ids]
+        return max(
+            roles,
+            key=lambda r: r.position,
+        )
+
     def __str__(self) -> str:
         return str(self.user)
 
--- a/tests/hikari/test_guilds.py
+++ b/tests/hikari/test_guilds.py
@@ -105,6 +105,7 @@ class TestMember:
                 flags="flags",
                 mention="<@123>",
                 __str__=mock.Mock(return_value="davfsa#0001"),
+                top_role=mock.Mock(return_value="Owner"),
             ),
             nickname="davb",
         )
@@ -160,6 +161,9 @@ class TestMember:
     def test_mention_property_when_no_nickname(self, obj):
         obj.nickname = None
         assert obj.mention == "<@123>"
+
+    def test_top_role(self, obj):
+        assert obj.top_role.guild_id == "Owner"
 
 
 class TestPartialGuild:


hikari/guilds.py
```diff
@@ -421,6 +421,21 @@ def mention(self) -> str:
...
+    def top_role(self) -> Role:
+        """Returns the top role of the given member.
+
+        Returns
+        -------
+        hikari.guilds.Role
+            The highest role the member has.
+        """
+        roles = [r for r in self.app.cache.get_roles_view_for_guild(self.guild_id).values() if r.id in self.role_ids]
+        return max(
```
davfsa, line 434: Please run black `nox -s black` to fix this as the linter is complaning

tests/hikari/test_guilds.py
```diff
@@ -161,6 +162,9 @@ def test_mention_property_when_no_nickname(self, obj):
         obj.nickname = None
         assert obj.mention == "<@123>"
 
+    def test_top_role(self, obj):
+        assert obj.top_role.guild_id == "Owner"
```
davfsa, line 166: This is not the right way of testing this. You will have to mock `app.cache.get_roles_view_for_guild`, make it return a dict with the values you expect, make sure it gets called with the right guild_id and that the role return is the correct one. You will also have to write tests (and modify the code you wrote) for when an empty role view is returned (missing intents so we didnt cache it or he is using a stateless bot).

davfsa (changes requested):
Several things and you will have to make it take into account a missing intents cache and a stateless bot (which will return a EmptyCacheView) and return `None` instead of a role. Also, always remember to run `nox` before you commit to make sure all pipelines will pass

nekokatt:
You are getting the hang of how these components slot together, just a couple of points that I'd like to see fixed from me and Dav, if that is okay?

Thanks for spending the time to do this, it is greatly appreciated :-)

nekokatt (changes requested):

ahnaf-zamil:
> 
> 
> You are getting the hang of how these components slot together, just a couple of points that I'd like to see fixed from me and Dav, if that is okay?
> 
> Thanks for spending the time to do this, it is greatly appreciated :-)

Thanks, I just tried to do something to help. I am not very familiar with caching and testing in general so it's kinda hard for me to contribute to this project.

nekokatt:
> > You are getting the hang of how these components slot together, just a couple of points that I'd like to see fixed from me and Dav, if that is okay?
> > Thanks for spending the time to do this, it is greatly appreciated :-)
> 
> Thanks, I just tried to do something to help. I am not very familiar with caching and testing in general so it's kinda hard for me to contribute to this project.

if you want any help, ping one of us :+1:

tests/hikari/test_guilds.py
```diff
@@ -105,6 +105,7 @@ def obj(self):
                 flags="flags",
                 mention="<@123>",
                 __str__=mock.Mock(return_value="davfsa#0001"),
+                top_role=mock.Mock(return_value="Owner"),
```
davfsa, line 108: Dont mock something you are testing ;)

PR #133, opened 2020-09-07
Fixed a logic error in websocket init logic

nekokatt:
This caused exceptions to be discarded if the websocket had already cloesd, causing an infinite restart loop as the gateway shard would assume the connection closed without a reason when passing invalid intents, for example.


FasterSpeeding (dismissed):
LGTM

FasterSpeeding (approved):
lgtm

davfsa (approved):

PR #134, opened 2020-09-07
avoid the shard ratelimits delaying event dispatch

FasterSpeeding:
### Summary
shard.request_guild_members in places to avoid it blocking other stuff unnecessarily.

### Checklist

### DIFF
--- a/hikari/impl/stateful_event_manager.py
+++ b/hikari/impl/stateful_event_manager.py
@@ -168,7 +168,8 @@ class StatefulEventManagerImpl(event_manager_base.EventManagerBase):
         # to chunk small guilds.
         if (event.guild.is_large or not presences_declared) and members_declared:
             # We create a task here instead of awaiting the result to avoid any rate-limits from delaying dispatch.
-            asyncio.create_task(shard.request_guild_members(event.guild))
+            coroutine = shard.request_guild_members(event.guild)
+            asyncio.create_task(coroutine, name=f"{event.shard.id}:{event.guild.id} guild create members request")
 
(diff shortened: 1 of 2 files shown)


hikari/impl/stateful_event_manager.py
```diff
@@ -166,7 +167,8 @@ async def on_guild_create(self, shard: gateway_shard.GatewayShard, payload: data
         # payload if presence intents are also declared, so if this isn't the case then we also want
         # to chunk small guilds.
         if (event.guild.is_large or not presences_declared) and members_declared:
-            await shard.request_guild_members(event.guild)
+            # We create a task here instead of awaiting the result to avoid any rate-limits from delaying dispatch.
+            asyncio.create_task(shard.request_guild_members(event.guild))
```
nekokatt, line 171: can we give this a name

nekokatt (changes requested):

FasterSpeeding (commented):
Wanted to have similar behaviour in StatefulGuildChunkerImpl.request_guild_members where we don't wait for the request to go through but here we have to ensure any errors from the validation are raised and while we could do something like the following
```python
finished_coroutines, _ asyncio.wait((asyncio.create_task(shard.request_guild_members())), timeout=1)
if finished_coroutines:
    await finished_coroutines[0]
```
this feels jank so not sure if that should be included or not

hikari/impl/stateful_event_manager.py
FasterSpeeding: done

nekokatt (approved):

PR #136, opened 2020-09-08
Several bug fixes

davfsa:
- Fix some typehints in REST
  - Fix logging message in bot
  - If we are expecting a BINARY, but get something different, we now error out

### Checklist

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -149,9 +149,9 @@ class _V6GatewayTransport(aiohttp.ClientWebSocketResponse):
                 self._logger.error("connection closed with code %s (%s)", close_code, reason)
 
                 can_reconnect = close_code < 4000 or close_code in (
+                    errors.ShardCloseCode.UNKNOWN_ERROR,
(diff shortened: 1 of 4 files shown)


hikari/impl/shard.py
```diff
@@ -149,9 +149,9 @@ async def _receive_and_check(self, timeout: typing.Optional[float], /) -> str:
                 self._logger.error("connection closed with code %s (%s)", close_code, reason)
 
                 can_reconnect = close_code < 4000 or close_code in (
+                    errors.ShardCloseCode.UNKNOWN_ERROR,
```
nekokatt, line 152: Do we want to die on an unknown error? It could be a one-off?

nekokatt (approved):

hikari/impl/shard.py
nekokatt: Oh wait, misunderstood.

davfsa: This was already here. I just moved it up for it to have some sort of order. (4000, 4001, 4002, etc)

PR #141, opened 2020-09-08
Add `update_presence` to bot

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -40,6 +40,7 @@ from hikari import errors
 from hikari import intents as intents_
 from hikari import presences
 from hikari import traits
+from hikari import undefined
 from hikari import users
 from hikari.api import cache as cache_
 from hikari.api import chunker as chunker_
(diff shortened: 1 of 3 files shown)


hikari/impl/bot.py
```diff
@@ -59,6 +59,7 @@
 from hikari.utilities import date
 from hikari.utilities import event_stream
 from hikari.utilities import ux
```
nekokatt, line 61: please run isort if you havent already

I would assume `ux` comes after `un`, but might be wrong

hikari/impl/bot.py
```diff
@@ -823,6 +824,56 @@ async def wait_for(
...
+
+            This method is simply a facade to make performing this in bulk
+            simpler.
+        """
+        coros = [
+            s.update_presence(status=status, activity=activity, idle_since=idle_since, afk=afk)
+            for s in self._shards.values()
+        ]
+
+        await asyncio.gather(*coros)
```
nekokatt, line 875: Use aio.all_of here or you will get incomplete task spam possibly if something fails due to a cancelled error.

hikari/impl/bot.py
```diff
@@ -823,6 +824,56 @@ async def wait_for(
     ) -> event_dispatcher.EventT_co:
         return await self._events.wait_for(event_type, timeout=timeout, predicate=predicate)
 
+    async def update_presence(
```
FasterSpeeding, line 827: Would it not make sense for this method to maybe also be declared on shard aware for the traits?

davfsa: All the traits only include properties, not public methods, which is why I didnt think about adding it. @nekokatt?

nekokatt: It should be added.

davfsa: Added :+1:

PR #143, opened 2020-09-08
Removed constants module mess.

nekokatt:
### Checklist


hikari/impl/shard.py
```diff
@@ -87,6 +88,20 @@
 # Rate-limit for chunking requests (used to prevent saturating the entire
 # ratelimit window).
 _CHUNKING_RATELIMIT: typing.Final[typing.Tuple[float, int]] = (60.0, 60)
+# Supported gateway version
+_VERSION: int = 6
```
davfsa, line 92: This gets defined here but never used. Is that intentional?

davfsa (dismissed):
Looks good to me, just had a little question

hikari/impl/shard.py
nekokatt: nope, fixed.

davfsa (approved):

hikari/events/channel_events.py
```diff
@@ -585,6 +586,7 @@ class PrivatePinsUpdateEvent(PinsUpdateEvent, PrivateChannelEvent):
     # <<inherited docstring from ChannelEvent>>.
 
     last_pin_timestamp: typing.Optional[datetime.datetime] = attr.ib(repr=True)
+
```
FasterSpeeding, line 589: Are these gaps intentional?

nekokatt: Apparently so.

PR #149, opened 2020-09-09
Task/close revamp

davfsa:
### Checklist

### DIFF
--- a/hikari/errors.py
+++ b/hikari/errors.py
@@ -94,8 +94,8 @@ class HikariInterrupt(KeyboardInterrupt, HikariError):
     signum: int = attr.ib()
     """The signal number that was raised."""
 
-    description: typing.Optional[str] = attr.ib()
-    """Signal description."""
+    signame: str = attr.ib()
+    """The signal name that was raised."""
 
 
 @attr.s(auto_exc=True, slots=True, repr=False, weakref_slot=False)
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -198,6 +198,7 @@ class BotApp(traits.BotAware, event_dispatcher.EventDispatcher):
         "_cache",
         "_chunker",
         "_closing_event",
+        "_closed",
         "_debug",
         "_entity_factory",
         "_events",
@@ -237,6 +238,7 @@ class BotApp(traits.BotAware, event_dispatcher.EventDispatcher):
         # Settings and state
         self._banner = banner
         self._closing_event = asyncio.Event()
+        self._closed = False
         self._debug = debug
         self._executor = executor
         self._http_settings = http_settings if http_settings is not None else config.HTTPSettings()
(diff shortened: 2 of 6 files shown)


hikari/impl/bot.py
```diff
@@ -844,6 +864,15 @@ async def update_presence(
 
         await aio.all_of(*coros)
 
+    async def _set_close_flag(self, signame: str, signum: int) -> None:
```
nekokatt, line 867: if this is becoming a method, perhaps `async def handle` above should be too?

hikari/impl/rest.py
```diff
@@ -120,6 +120,8 @@ class BasicLazyCachedTCPConnectorFactory(rest_api.ConnectorFactory):
 
     def __init__(self, **kwargs: typing.Any) -> None:
         self.connector: typing.Optional[aiohttp.TCPConnector] = None
+        kwargs.setdefault("enable_cleanup_closed", True)
```
nekokatt, line 123: so um, didn't this cause a windows bug earlier

davfsa: Looks like not having it is worst, lol. After all this changes I try to replicate it and I went 50 bot restarts without an issue, so idk where nor how, but its fixed

hikari/impl/bot.py
```diff
@@ -526,12 +574,12 @@ def run(
             The number of shards to use in the entire distributed application.
             Defaults to `builtins.None` which results in the count being
             determined dynamically on startup.
-        status : hikari.presences.Status
-            The initial status to show for the user presence on startup.
-            Defaults to `hikari.presences.Status.ONLINE`.
         """
+        if shard_ids is not None and shard_count is None:
+            raise RuntimeError("'shard_ids' must be passed with 'shard_count'")
```
nekokatt, line 579: This should be a TypeError come to think of it...

hikari/impl/bot.py
davfsa: I moved this here to make `run` shorter as it was too populated. I also tried to move `handle`, but since it accesses variables that are on `run`, I decided not to. I can move it if you want

hikari/impl/bot.py
davfsa: Will change :+1:

hikari/impl/bot.py
```diff
@@ -448,15 +483,16 @@ def run(
         *,
         activity: typing.Optional[presences.Activity] = None,
         afk: bool = False,
+        idle_since: typing.Optional[datetime.datetime] = None,
```
nekokatt, line 486: i kinda did alphabetise these so it was easier to scan a list.

We should decide on how to do this consistently as me and you keep doing the complete opposite lol.

I'm happy to discuss which way we should follow going forwards.

hikari/impl/bot.py
```diff
@@ -448,15 +483,16 @@ def run(
...
+        status: presences.Status = presences.Status.ONLINE,
         asyncio_debug: typing.Optional[bool] = None,
         close_executor: bool = False,
```
nekokatt, line 489: While we are here, can we rename this to `close_passed_executor` so that it is disambiguated from the default executor for the event loop, which is closed explicitly in >= CPython 3.9 only?

hikari/impl/bot.py
```diff
@@ -23,6 +23,9 @@
 
 from __future__ import annotations
 
+import threading
```
nekokatt, line 26: The imports seem to have escaped.

The patch I sent earlier wasn't formatted as I didn't have half the stuff installed I needed, so you might want to just double check first.

hikari/impl/bot.py
davfsa: Yep :)

hikari/impl/bot.py
davfsa: Will do :+1:

hikari/impl/bot.py
davfsa: I generally just bundle them up with the other ones that act the same. Eg: activity, afk, idle_since and status all control the bot status that gets displayed, so i put them next to each other. I can revert this if you want

hikari/impl/bot.py
davfsa: Looks like they were fine :)

hikari/impl/bot.py
nekokatt: As long as it is consistent, it is fine. I just personally find alphabetised stuff easier to spot missing things in, etc.

hikari/impl/bot.py
```diff
@@ -543,18 +591,49 @@ def run(
             except AttributeError:
                 _LOGGER.debug("cannot set coroutine tracking depth for %s, no functionality exists for this", loop)
 
-        def signal_handler(signum: int) -> None:
-            raise errors.HikariInterrupt(signum, signal.strsignal(signum))
+        # Throwing this in the handler will lead to lots of fun OS specific shenanigans. So, lets just
+        # cache it for later, I guess.
+        interrupt: typing.Optional[errors.HikariInterrupt] = None
```
nekokatt, line 596: You could possibly achieve this using a hack with `types.SimpleNamespace` if you wanted to move the handler but you'd still need a way to pass it around so I think this is fine here. It just looks bad because of my essay.

davfsa:
All fixed :+1: Ready for another review :)

FasterSpeeding:
what's the behaviour for double keyboard interrupts on linux during startup for this? On windows it seems to just treat each interrupt as a single keyboard interrupt on startup even if you spam it and will wait until all the shards have started up to start shutting down

hikari/errors.py
```diff
@@ -94,8 +94,8 @@ class HikariInterrupt(KeyboardInterrupt, HikariError):
     signum: int = attr.ib()
     """The signal number that was raised."""
 
-    description: typing.Optional[str] = attr.ib()
-    """Signal description."""
+    signame: str = attr.ib()
+    """The signal name that was raised."""
```
nekokatt, line 98: Can we use the property format for the docstring

nekokatt (approved):

FasterSpeeding (commented):
(copied from earlier comment to make it an actual review)
what's the behaviour for double keyboard interrupts on linux during startup for this? On windows it seems to just treat each interrupt as a single keyboard interrupt on startup even if you spam it and will wait until all the shards have started up to start shutting down

nekokatt:
@FasterSpeeding

On Windows, CTRL-BREAK, and on Linux, CTRL \ are used to murder a process immediately.

Keyboard interrupts are a polite request to close soon if possible. Processes are not required to do them as is equal with C on signal term

PR #150, opened 2020-09-09
Add methods to BotAware along with moving docs

FasterSpeeding:
### Summary
Add methods around starting and stopping the bot app to the BotAware trait and move docs from impl

### Checklist

- [ x] I have run `nox` and all the pipelines have passed.
- [x ] I have made unittests according to the code I have added/modified/deleted.

### DIFF
--- a/hikari/traits.py
+++ b/hikari/traits.py
@@ -572,20 +572,3 @@ class BotAware(RESTAware, ShardAware, EventFactoryAware, DispatcherAware, typing
             Defaults to `hikari.presences.Status.ONLINE`.
         """
         raise NotImplementedError
-
-    async def terminate(self, close_executor: bool = False) -> None:
-        """Kill the application by shutting all components down.
-
-        Other Parameters
-        ----------------
-        close_executor : builtins.bool
-            Defaults to `builtins.False`. If `builtins.True`, any custom
-            `concurrent.futures.Executor` passed to the constructor will be
-            shut down when the application terminates. This does not affect the
-            default executor associated with the event loop, and will not
-            do anything if you do not provide a custom executor to the
-            constructor.
-        """
-        raise NotImplementedError
-
-    __slots__: typing.Sequence[str] = ()
--- a/hikari/traits.py
+++ b/hikari/traits.py
@@ -470,11 +470,7 @@ class BotAware(RESTAware, ShardAware, EventFactoryAware, DispatcherAware, typing
         *,
         activity: typing.Optional[presences.Activity] = None,
         afk: bool = False,
-        asyncio_debug: typing.Optional[bool] = None,
         close_executor: bool = False,
-        close_loop: bool = True,
-        coroutine_tracking_depth: typing.Optional[int] = None,
-        enable_signal_handlers: bool = True,
         idle_since: typing.Optional[datetime.datetime] = None,
         ignore_session_start_limit: bool = False,
         large_threshold: int = 250,
@@ -492,9 +488,6 @@ class BotAware(RESTAware, ShardAware, EventFactoryAware, DispatcherAware, typing
         afk : builtins.bool
             The initial AFK state to display in the bot user presence, or
             `builtins.False` (default) to not show any.
-        asyncio_debug : builtins.bool
-            Defaults to `builtins.False`. If `builtins.True`, then debugging is
-            enabled for the asyncio event loop in use.
         close_executor : builtins.bool
             Defaults to `builtins.False`. If `builtins.True`, any custom
             `concurrent.futures.Executor` passed to the constructor will be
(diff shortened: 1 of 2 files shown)


hikari/traits.py
```diff
@@ -446,4 +446,177 @@ async def update_presence(
...
+        close_executor: bool = False,
+        close_loop: bool = True,
+        coroutine_tracking_depth: typing.Optional[int] = None,
+        enable_signal_handlers: bool = True,
+        idle_since: typing.Optional[datetime.datetime] = None,
+        ignore_session_start_limit: bool = False,
+        large_threshold: int = 250,
+        status: presences.Status = presences.Status.ONLINE,
+        shard_ids: typing.Optional[typing.Set[int]] = None,
+        shard_count: typing.Optional[int] = None,
```
nekokatt, line 483: I don't agree that most of these should be present.

Half of this stuff is implementation specific, and implementations may not want to or may not be able to provide some of these.

I'd revert:
- asyncio_debug  (as their loop may not support this properly or may have other implications if it uses an existing loop)
- close_loop  (as their impl may use the event loop outside this)
- coroutine_tracking_depth (provisional feature in CPython, may go at any time)
- enable_signal_handlers (implementations may not bother with signals at all)

nekokatt (changes requested):

hikari/traits.py
FasterSpeeding: ok 👍

nekokatt: Under the rule of `def x(a, b, c)` being compatible with `def x(a, b, c, d=None, e=None, f=None), this should be fine with MyPy.

nekokatt: Also afaik unfortunately you'll have to duplicate this docstring because pdoc3 defaults :/

FasterSpeeding: Well if we're changing the args from protocol to impl it won't be duping anyway and pdoc3 is inheriting the docs here

FasterSpeeding: Removed the ones you listed, if you think any more should be removed then send a list

nekokatt: thats it

nekokatt (approved):

nekokatt:
Please can you rebase onto master.

Chat to @davfsa if unsure about changes

hikari/traits.py
```diff
@@ -446,4 +446,146 @@ async def update_presence(
...
+            Defaults to `builtins.False`. If `builtins.True`, any custom
+            `concurrent.futures.Executor` passed to the constructor will be
+            shut down when the application terminates. This does not affect the
+            default executor associated with the event loop, and will not
+            do anything if you do not provide a custom executor to the
+            constructor.
+        """
+        raise NotImplementedError
+
     __slots__: typing.Sequence[str] = ()
```
nekokatt, line 591:
```suggestion
```

hikari/traits.py
```diff
@@ -446,4 +446,146 @@ async def update_presence(
 class BotAware(RESTAware, ShardAware, EventFactoryAware, DispatcherAware, typing.Protocol):
     """Structural supertype for a component that is aware of all internals."""
 
+    async def join(self, until_close: bool = True) -> None:
```
nekokatt, line 449:
```suggestion
class BotAware(RESTAware, ShardAware, EventFactoryAware, DispatcherAware, typing.Protocol):
    """Structural supertype for a component that is aware of all internals."""
    
    __slots__: typing.Sequence[str] = ()

    async def join(self, until_close: bool = True) -> None:
```

nekokatt (changes requested):
Fixing conflicts with master and fixing a slots that fell out of the right place :)

hikari/traits.py
```diff
@@ -446,4 +446,129 @@ async def update_presence(
 class BotAware(RESTAware, ShardAware, EventFactoryAware, DispatcherAware, typing.Protocol):
     """Structural supertype for a component that is aware of all internals."""
 
-    __slots__: typing.Sequence[str] = ()
+    async def join(self, until_close: bool = True) -> None:
+        """Wait indefinitely until the application closes.
```
nekokatt, line 450:
```suggestion
class BotAware(RESTAware, ShardAware, EventFactoryAware, DispatcherAware, typing.Protocol):
    """Structural supertype for a component that is aware of all internals."""

    __slots__: typing.Sequence[str] = ()
    
    async def join(self, until_close: bool = True) -> None:
        """Wait indefinitely until the application closes.
```

nekokatt: Gah, GitHub won't let me do it. Would you mind just re-adding the slots in for me here? You'd moved them down below everything else and I was trying to undo this for you so you didn't have to waste time fixing stuff that wasn't your fault, but I guess I made it worse. Sorry...

nekokatt (approved):

PR #152, opened 2020-09-10
Remove the UnavailableGuild model and separate available and unavailable guilds in the cache

FasterSpeeding:
### Summary
* Remove the UnavailableGuild model
* Separate guild reading methods for guild into available and unavailable guild methods (with these changes reflected in model methods).

### Checklist

### DIFF
--- a/hikari/api/cache.py
+++ b/hikari/api/cache.py
@@ -149,8 +149,8 @@ class Cache(abc.ABC):
         """
 
     @abc.abstractmethod
-    def get_guild(self, guild_id: snowflakes.Snowflake, /) -> typing.Optional[guilds.GatewayGuild]:
-        """Get a guild object from the cache.
+    def get_available_guild(self, guild_id: snowflakes.Snowflake, /) -> typing.Optional[guilds.GatewayGuild]:
+        """Get the object of an available guild from the cache.
 
         Parameters
         ----------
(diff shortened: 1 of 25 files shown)


hikari/api/cache.py
```diff
@@ -166,8 +166,45 @@ def get_guild(self, guild_id: snowflakes.Snowflake, /) -> typing.Optional[guilds
...
     @abc.abstractmethod
-    def get_guilds_view(self) -> CacheView[snowflakes.Snowflake, guilds.GatewayGuild]:
-        """Get a view of the guild objects in the cache.
+    def get_unavailable_guild(self, guild_id: snowflakes.Snowflake) -> typing.Optional[guilds.GatewayGuild]:
+        """Get the object of a unavailable guild from the cache.
+
+        !!! note
+            Unlike `Cache.get_available_guild`, the objects returned by this
+            method will likely be out of date and not-accurate as they are
+            considered unavailable meaning that we aren't receiving gateway
```
nekokatt, line 175:
```suggestion
            considered unavailable meaning, that we are not receiving gateway
```

hikari/api/cache.py
```diff
@@ -166,8 +166,45 @@ def get_guild(self, guild_id: snowflakes.Snowflake, /) -> typing.Optional[guilds
...
+            A view of guild IDs to the guild objects found in the cache.
+        """
+
+    @abc.abstractmethod
+    def get_unavailable_guilds_view(self) -> CacheView[snowflakes.Snowflake, guilds.GatewayGuild]:
+        """Get a view of the unavailable guild objects in the cache.
+
+        !!! note
+            Unlike `Cache.get_available_guilds_view`, the objects returned by
+            this method will likely be out of date and not-accurate as they are
```
nekokatt, line 205:
```suggestion
            this method will likely be out of date and inaccurate as they are
```

hikari/events/channel_events.py
```diff
@@ -137,18 +137,34 @@ def guild_id(self) -> snowflakes.Snowflake:
         """
 
     @property
-    def guild(self) -> typing.Optional[guilds.GatewayGuild]:
-        """Get the cached guild that this event relates to, if known.
+    def available_guild(self) -> typing.Optional[guilds.GatewayGuild]:
```
nekokatt, line 140: I'm not sure I like this, I think `guild` is fine here.

I feel like this will possibly get confusing for people, as it forces their code to make assumptions about the gateway guild state (which may or may not correlate to the state on the REST API).

I would hope you can usually assume that if you got a channel event, it would still be available. But this kind of obfuscates how you access a guild here.

I think it would be safer to just do
```py
@property
def guild() -> typing.Optional[guilds.GatewayGuild]:
    return self.app.cache.get_available_guild(self.guild_id) or self.app.cache.get_unavailable_guild(self.guild_id)
```

hikari/events/message_events.py
```diff
@@ -136,7 +136,7 @@ def channel(self) -> typing.Optional[channels.GuildTextChannel]:
         return typing.cast("channels.GuildTextChannel", self.app.cache.get_guild_channel(self.channel_id))
 
     @property
-    def guild(self) -> typing.Optional[guilds.GatewayGuild]:
+    def available_guild(self) -> typing.Optional[guilds.GatewayGuild]:
```
nekokatt, line 139: Yeah, I think we should revert this change in particular, as it is going to cause a headache for people using this already for not much additional benefit. We should just be able to store a flag on the guild to say if it is available or not.

hikari/api/cache.py
FasterSpeeding: odd placement for a comma

hikari/events/channel_events.py
```diff
@@ -148,7 +148,7 @@ def guild(self) -> typing.Optional[guilds.GatewayGuild]:
             The gateway guild this event relates to, if known. Otherwise
             this will return `builtins.None`.
         """
-        return self.app.cache.get_guild(self.guild_id)
+        return self.app.cache.get_available_guild(self.guild_id) or self.app.cache.get_unavailable_guild(self.guild_id)
```
nekokatt, line 151: This is much better :)

nekokatt (approved):

PR #154, opened 2020-09-10
Fixed several bugs in shard/bot run logic.

nekokatt:
- Fixed bug where shard would not immediately close if awaiting startup in bot.
- Fixed bug where shards get requested to shutdown, but never joined, so the
    event loop task reaper is what kills them.
- Removed unused and undocumented shard version parameter.
- Fixed bug in 5 second shard start backoff where close waiter was being waited
    for, but within an `all_of`, which meant it would not do anything until
    any shard threw an exception, or the 5 second window was up.
- Fixed bug where a single shard being closed on startup may be missed if the
    shard is closed directly.
- Added more contextual debug logging to say what case caused the closure each time.

@davfsa, @FasterSpeeding please verify when you get time.

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -699,8 +699,7 @@ class GatewayShardImpl(shard.GatewayShard):
 
             self._logger.debug("preparing to send HEARTBEAT [s:%s, interval:%ss]", self._seq, heartbeat_interval)
 
-            # TODO: Could this error, or can I just assume any other receiver will die first?
-            await self._send_heartbeat()
+            task = asyncio.create_task(self._send_heartbeat(), name=f"shard {self.id} send heartbeat")
 
             try:
                 await asyncio.wait_for(self._closing.wait(), timeout=heartbeat_interval)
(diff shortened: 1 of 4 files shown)


hikari/impl/shard.py
```diff
@@ -742,9 +747,13 @@ async def _run(self) -> None:
                         pass
 
                 try:
+                    last_started_at = date.monotonic()
                     if not await self._run_once():
                         self._logger.debug("shard has shut down")
 
+                except errors.GatewayConnectionError as ex:
+                    self._logger.error("failed to connect to server, reason was %s, will retry shortly", ex.__cause__)
```
davfsa, line 753:
```suggestion
                    self._logger.error("failed to connect to server, reason was: %s. Will retry shortly", ex.__cause__)
```

davfsa (approved):
LGTM

hikari/impl/shard.py
```diff
@@ -840,16 +868,23 @@ async def _run_once(self) -> bool:
...
-                    # If the heartbeat died due to an error, it should be raised here. We expect the heartbeat
-                    # to always kill our connection if it dies.
-                    # We return True if zombied
+                    # If the heartbeat died due to an error, it should be raised here.
+                    # This will currently allow us to try to resume if that happens
+                    # We return True if zombied.
                     if await heartbeat_task:
                         now = date.monotonic()
                         self._logger.error(
                             "connection is a zombie, last heartbeat sent %ss ago",
```
davfsa, line 877:
```suggestion
                            "connection is a zombie, last heartbeat sent %.2fs ago",
```

PR #155, opened 2020-09-10
Task/debug is dead

nekokatt:
This must not be merged until #154 is merged, as it builds on top of what was done there.

Essentially, `Bot(debug=True)` no longer exists, and a new TRACE_HIKARI logging level can be used instead.

Fixed some REST-only bug too.

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -72,6 +72,7 @@ if typing.TYPE_CHECKING:
 
 LoggerLevelT = typing.Union[
     int,
+    typing.Literal["TRACE_HIKARI"],
     typing.Literal["DEBUG"],
     typing.Literal["INFO"],
     typing.Literal["WARNING"],
@@ -81,8 +82,8 @@ LoggerLevelT = typing.Union[
 """Type-hint for a valid logging level.
 
 This may be an `int` logging level (e.g. `logging.DEBUG`, `logging.CRITICAL`),
-or a capitalized string that matches one of `"DEBUG"`, `"INFO"`, `"WARNING"`,
-`"ERROR"`, or `"CRITICAL"`.
+or a capitalized string that matches one of `"TRACE_HIKARI"`, `"DEBUG"`,
+`"INFO"`, `"WARNING", `"ERROR"`, or `"CRITICAL"`.
 """
 
 _LOGGER: typing.Final[logging.Logger] = logging.getLogger("hikari")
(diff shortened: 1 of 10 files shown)


davfsa (approved):

hikari/impl/bot.py
```diff
@@ -870,7 +865,7 @@ async def _start_one_shard(
         end = date.monotonic()
 
         if new_shard.is_alive:
-            _LOGGER.info("Shard %s started successfully in %.1fms", shard_id, (end - start) * 1_000)
+            _LOGGER.debug("Shard %s started successfully in %.1fms", shard_id, (end - start) * 1_000)
```
davfsa, line 868: I think should stay in `info`, provides more feedback to the user

hikari/impl/bot.py
```diff
@@ -833,7 +829,7 @@ async def _set_close_flag(self, signame: str, signum: int) -> None:
         # without getting undefined behaviour. We do however have `asyncio.run_coroutine_threadsafe` which can
         # run a coroutine function on the event loop from a completely different thread, so this is the safest
         # solution.
-        _LOGGER.info("received interrupt %s (%s), will start shutting down shortly", signame, signum)
+        _LOGGER.debug("received interrupt %s (%s), will start shutting down shortly", signame, signum)
```
davfsa, line 832: This should also stay on info since it will provide feedback to windows users since `^C` doesn't get outputted to some terminals (cmd) when causing SIGINT

hikari/impl/rest.py
```diff
@@ -274,12 +291,18 @@ def acquire(self, token: str, token_type: str = _BEARER_TOKEN_PREFIX) -> rest_ap
...
         # We can achieve this using a lambda. This allows the entity factory to build models that
         # are also REST-aware
-        entity_factory = entity_factory_impl.EntityFactoryImpl(_RESTProvider(lambda: rest_client))
+        cache = stateless_cache.StatelessCacheImpl()
+        provider = _RESTProvider(
+            lambda: entity_factory,
+            self._executor,
+            lambda: cache,
+            lambda: rest_client,
+        )
```
davfsa, line 300: Any reason why these have to be provided as lambdas?

davfsa (commented):
LGTM, just some things I wanted to ask

hikari/impl/rest.py
nekokatt: because circular dependency

davfsa: Ah, ok :+1:

hikari/impl/bot.py
nekokatt: it will output that the bot is shutting down anyway

hikari/impl/bot.py
nekokatt: theres already a gateway log to say it is ready, which immediately triggers this

PR #165, opened 2020-09-12
Added message author attributes to events and fixed typing.

### LINKED ISSUE #162: Add author to message events.
Allow us to do `event.author` instead of `event.message.author` on message create events.

nekokatt:
Guild message events are now documented to provide News or Text channels in output rather than just
GuildTextChannel derivatives.

Fixes #162

### DIFF
--- a/hikari/events/message_events.py
+++ b/hikari/events/message_events.py
@@ -48,14 +48,13 @@ import attr
 
 from hikari import intents
 from hikari import snowflakes
+from hikari import undefined
 from hikari import users
 from hikari.events import base_events
 from hikari.events import shard_events
 from hikari.utilities import attr_extensions
 
 if typing.TYPE_CHECKING:
-    # Do NOT remove the users import here. It **is** required, even if PyCharm
-    # tries to assure you otherwise.
     from hikari import channels
     from hikari import guilds
     from hikari import messages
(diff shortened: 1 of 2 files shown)


hikari/events/message_events.py
```diff
@@ -292,14 +356,27 @@ class GuildMessageCreateEvent(GuildMessageEvent, MessageCreateEvent):
...
     def guild_id(self) -> snowflakes.Snowflake:
         # <<inherited docstring from GuildMessageEvent>>.
         # Always present in this event.
-        return typing.cast("snowflakes.Snowflake", self.message.guild_id)
+        guild_id = self.message.guild_id
+        assert isinstance(guild_id, snowflakes.Snowflake)
+        return guild_id
+
+    @property
+    def author(self) -> typing.Union[guilds.Member]:
```
tandemdude, line 370: The `Union` here seems redundant, surely it should `guilds.Member` instead. Unless you mean `typing.Union[guilds.Member, users.User]`

tandemdude (changes requested):

hikari/events/message_events.py
nekokatt: I think I meant the latter but now I think about it, that is redundant.

tandemdude:
```suggestion
    def author(self) -> guilds.Member:
```

tandemdude (approved):
LGTM now

PR #167, opened 2020-09-12
Implemented `user` member on typing events.

nekokatt:
- GuildTypingEvent has had the `member` attribute renamed to `user`
    (this is a breaking change).
- PrivateTypingEvent now has a property to get the cached user
    (if possible).


tandemdude (approved):
LGTM

PR #168, opened 2020-09-12
Migrate to V8 Gateway and REST API

nekokatt:
This change migrates the entire API from v6 REST and v6 Gateway to v8 REST and v8 Gateway. This PR will be updated with details as it develops. Some of these changes may be provisional, since at the time of writing, not many details have been released on what is being changed.

Merging this will introduce a number of breaking changes:

- Intents are now MANDATORY if you want to be able to connect to the gateway at all.
- MemberPresence has had some fields removed, namely:
    - role_ids
    - premium_since
    - nick(name)
- hikari.Bot is now named hikari.BotApp
- hikari.REST is now named hikari.RESTApp


davfsa (approved):
LGTM

PR #174, opened 2020-09-13
Added the ability to mark files as spoilers

davfsa:
### Checklist

### DIFF
--- a/hikari/files.py
+++ b/hikari/files.py
@@ -67,6 +67,7 @@ if typing.TYPE_CHECKING:
     import types
 
 _MAGIC: typing.Final[int] = 50 * 1024
+SPOILER_TAG: typing.Final[str] = "SPOILER_"
 
 ReaderImplT = typing.TypeVar("ReaderImplT", bound="AsyncReader")
 ReaderImplT_co = typing.TypeVar("ReaderImplT_co", bound="AsyncReader", covariant=True)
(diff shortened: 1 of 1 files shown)


hikari/files.py
```diff
@@ -817,22 +817,27 @@ class File(Resource[FileReader]):
...
     path: pathlib.Path
     """The path to the file."""
 
+    is_spoiler: bool
+    """Whether the file will be marked as a spoiler."""
+
     _filename: typing.Optional[str]
 
-    def __init__(self, path: Pathish, filename: typing.Optional[str] = None) -> None:
+    def __init__(self, path: Pathish, filename: typing.Optional[str] = None, spoiler: bool = False) -> None:
```
nekokatt, line 837:
```suggestion
    def __init__(self, path: Pathish, /, filename: typing.Optional[str] = None, *, spoiler: bool = False) -> None:
```

nekokatt:
What about bytes, webresources/urls?

Can embeds have spoilt images? If so, we'd want to adjust the logic in the embed serializer to upload any spoilt resources again so that we can inject a file name, right?

davfsa:
> What about bytes, webresources/urls?
Will add it to `Bytes`, websources and urls im not really sure can be spoilt (since they are text). Maybe adding them between `||` if the user selects to spoil them?

> Can embeds have spoilt images? If so, we'd want to adjust the logic in the embed serializer to upload any spoilt resources again so that we can inject a file name, right?
No, looks like they cant, they just render them as if they were not spoilers

hikari/files.py
davfsa: Fixed :+1:

PR #175, opened 2020-09-13
Add `top_role` property to member

### LINKED ISSUE #46: Suggestion member.get_top_role()

davfsa:
### Checklist

### Related issues
Closes #46

### DIFF
--- a/hikari/guilds.py
+++ b/hikari/guilds.py
@@ -426,6 +426,29 @@ class Member(users.User):
         """
         return f"<@!{self.id}>" if self.nickname is not None else self.user.mention
 
+    @property
+    def top_role(self) -> typing.Optional[Role]:
+        """Return the highest role the member has.
+
+        Returns
+        -------
+        typing.Optional[hikari.guilds.Role]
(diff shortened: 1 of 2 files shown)


hikari/guilds.py
```diff
@@ -426,6 +426,29 @@ def mention(self) -> str:
...
+        """
+        roles_view = self.app.cache.get_roles_view_for_guild(self.guild_id)
+
+        if roles_view:
+            roles = []
+            for role_id in self.role_ids:
+                if (role := roles_view.get(role_id)) is not None:
+                    roles.append(role)
+
+            if roles:
```
nekokatt, line 447: could this be
```py
roles = sorted(role for role in roles_view.values() if role.id in self.role_ids, key=lambda r: r.position)

try:
    return next(iter(roles))
except StopIteration:
    return None
```

nekokatt (commented):
Looks good, just want to check if we can simplify this or not.

hikari/guilds.py
davfsa: Works, so i implemented it :+1:

PR #178, opened 2020-09-13
Made flag enums lazily load, speeding up cache and event ops.

nekokatt:
IntFlags are now stored on models in private fields as raw ints.

These are cast to the flag type only when accessed, which speeds up startup by some time.


tandemdude (approved):
lgtm

PR #183, opened 2020-09-15
Add Competing activity type

FasterSpeeding:
For some reason discord hasn't documented the fields this introduces to the activity object so this may end up being a preliminary commit.

### Summary
Add "Competing" activity type

### Checklist

### DIFF
--- a/hikari/presences.py
+++ b/hikari/presences.py
@@ -85,6 +85,9 @@ class ActivityType(enum.IntEnum):
         Bots currently do not support setting custom statuses.
     """
 
+    COMPETING = 5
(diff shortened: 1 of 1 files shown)


davfsa (commented):
Can bots set their status as COMPETING? If they cant, we should probably document it as we did with custom statuses

nekokatt (approved):

nekokatt:
> Can bots set their status as COMPETING? If they cant, we should probably document it as we did with custom statuses

Merging this as I want to push a tag; will make an issue to look into this.

PR #186, opened 2020-09-15
Add ability to disable checking for updates

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -509,6 +509,7 @@ class BotApp(traits.BotAware, event_dispatcher.EventDispatcher):
         activity: typing.Optional[presences.Activity] = None,
         afk: bool = False,
         asyncio_debug: typing.Optional[bool] = None,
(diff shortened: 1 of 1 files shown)


hikari/traits.py
```diff
@@ -547,6 +548,9 @@ async def start(
         afk : builtins.bool
             The initial AFK state to display in the bot user presence, or
             `builtins.False` (default) to not show any.
+        check_for_updates : builtins.bool
+            Defaults to `builtins.True`. If `builtins.True`, will check for
```
nekokatt, line 552: I don't think we need to enforce all implementations require checking for updates.

Probably best to revert this file.

nekokatt (changes requested):
LGTM, but we should leave traits alone :)

hikari/traits.py
davfsa: Will revert, will write this documentation in bot.py instead :+1:

nekokatt (approved):

PR #190, opened 2020-09-16
Remove mapping.copy_mapping

davfsa:
### Checklist

### DIFF
--- a/hikari/utilities/mapping.py
+++ b/hikari/utilities/mapping.py
@@ -31,7 +31,6 @@ __all__: typing.Sequence[str] = [
     "MRIMutableMapping",
     "CMRIMutableMapping",
     "get_index_or_slice",
-    "copy_mapping",
 ]
 
 import abc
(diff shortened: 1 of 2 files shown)


nekokatt:
This won't work.

Python only defines `copy` on instances of `dict`. It isn't part of `typing.Mapping` nor `typing.MutableMapping`. @FasterSpeeding implemented this because our implementation only requires that you pass a mapping. You could use an `immutables.Map` in the future and still expect the code to work, even though that does not have `.copy()` on the interface.

This is why we have the try-catchall for this.

As you can see, the permissions overwrite field is defined like so:
```py
    permission_overwrites: typing.Mapping[snowflakes.Snowflake, PermissionOverwrite] = attr.ib(...)
```

This is why MyPy is now failing for you :^)

hikari/utilities/mapping.py
```diff
@@ -297,17 +296,3 @@ def get_index_or_slice(
...
-
-    .. deprecated::
-        `MappedCollection` should be preferred over this.
-    """
-    # dict.copy ranges from between roughly 2 times to 5 times more efficient than casting to a dict so we want to
-    # try to use this where possible.
-    try:
-        return mapping.copy()  # type: ignore[attr-defined, no-any-return]
-    except (AttributeError, TypeError):
-        raise NotImplementedError("provided mapping doesn't implement a copy method") from None
```
nekokatt, line 313: @FasterSpeeding while I agree we should be keeping this, I do wonder if we could fall back to returning something like `dict(mapping)` or `copy.copy` in this exception case, rather than dying?

Maybe with a `warnings.warn` to flag this up the first time, perhaps?

FasterSpeeding: if you want to fall back to it's original behaviour then keeping it around sounds good to deal with these specific cases sounds fine to me

PR #191, opened 2020-09-16
Added new logic to shard that should hopefully detect random socket c…

nekokatt:
…losures.

It appears Windows will send an EOF to the TCP socket if an adapter closes.
This seems to happen when connecting to a VPN or putting the computer into
hibernation/sleep mode. When this occurs, aiohttp assumes it means a normal
closure and gracefully closes, but we do not usually have a way to determine
this.

We can however see that aiohttp invokes ClientWebSocketResponse#close which
is the same call we make. The solution is to make a facading method that we
invoke that sets a flag determining that we did in fact want to close. If this
doesn't get set, we know aiohttp is the one to blame.

@FasterSpeeding please try and break this.

### DIFF
--- a/hikari/impl/rate_limits.py
+++ b/hikari/impl/rate_limits.py
@@ -441,8 +441,8 @@ class ExponentialBackOff:
     base : builtins.float
         The base to use. Defaults to `2.0`.
     maximum : builtins.float
-        The max value the backoff can be in a single iteration before an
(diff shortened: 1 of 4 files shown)


nekokatt:
Please verify and review.

nekokatt:
Assuming no further issues.

PR #195, opened 2020-09-16
Bugfix/windows os error

nekokatt:
Fixes asyncio windows error where "something that is not a socket was shutdown".

### DIFF
--- a/hikari/config.py
+++ b/hikari/config.py
@@ -23,8 +23,6 @@
 
 from __future__ import annotations
 
-import yarl
-
 __all__: typing.Final[typing.List[str]] = [
     "BasicAuthHeader",
     "ProxySettings",
@@ -37,6 +35,7 @@ import ssl as ssl_
 import typing
 
 import attr
+import yarl
 
 from hikari.utilities import attr_extensions
 from hikari.utilities import data_binding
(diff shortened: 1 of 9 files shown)


hikari/config.py
```diff
@@ -116,45 +203,182 @@ class HTTPTimeoutSettings:
...
     """
 
+    @acquire_and_connect.validator
+    @request_socket_connect.validator
+    @request_socket_read.validator
+    @total.validator
+    def _(self, attrib: attr.Attribute[typing.Optional[float]], value: typing.Optional[float]) -> None:
+        # This error won't occur until some time in the future where it will be annoying to
+        # try and determine the root cause, so validate it NOW.
+        if value is not None and (not isinstance(value, (float, int)) or value <= 0):  # type: ignore[unreachable]
```
nekokatt, line 256: typeshed is nice, it assumes the validators are always getting valid input
that defeats the object of them in the first place surely?

FasterSpeeding (approved):
lgtm

hikari/utilities/ux.py
```diff
@@ -22,6 +22,10 @@
 """User-experience extensions and utilities."""
 from __future__ import annotations
 
+import time
+
+from hikari.utilities import net
```
FasterSpeeding, line 27: more miss-placed imports

nekokatt: k, fix on other pr.

PR #196, opened 2020-09-16
Fixes edge case where messages in guilds had a None 'author' attribute

nekokatt:
Fixes edge case where messages in guilds had a None 'author' attribute if a webhook sent the message.

### Checklist

### DIFF
--- a/hikari/events/message_events.py
+++ b/hikari/events/message_events.py
@@ -214,6 +214,41 @@ class MessageCreateEvent(MessageEvent, abc.ABC):
             The user that sent the message.
         """
 
+    @property
+    def is_bot(self) -> bool:
+        """Return `builtins.True` if the message is from a bot.
+
+        Returns
+        -------
+        builtins.bool
+            `builtins.True` if from a bot, or `builtins.False` otherwise.
+        """
+        return self.message.author.is_bot
+
+    @property
+    def is_webhook(self) -> bool:
+        """Return `builtins.True` if the message was created by a webhook.
+
+        Returns
+        -------
+        builtins.bool
+            `builtins.True` if from a webhook, or `builtins.False` otherwise.
+        """
+        return self.message.webhook_id is not None
+
+    @property
+    def is_human(self) -> bool:
+        """Return `builtins.True` if the message was created by a human.
+
+        Returns
+        -------
+        builtins.bool
+            `builtins.True` if from a human user, or `builtins.False` otherwise.
+        """
+        # Not second-guessing some weird edge case will occur in the future with this,
+        # so I am being safe rather than sorry.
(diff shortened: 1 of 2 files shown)


hikari/events/message_events.py
```diff
@@ -214,6 +214,41 @@ def author(self) -> users.User:
...
+        """Return `builtins.True` if the message was created by a human.
+
+        Returns
+        -------
+        builtins.bool
+            `builtins.True` if from a human user, or `builtins.False` otherwise.
+        """
+        # Not second-guessing some weird edge case will occur in the future with this,
+        # so I am being safe rather than sorry.
+        return not self.message.author.is_bot and self.message.webhook_id is None
```
FasterSpeeding, line 250: If we're account for edge cases should we not consider checking for if it's a system user or are the definitions between what might account for a system users in the future a bit too blured for that to be accurate. (not too sure about this but what i do know is discord isn't consistent with setting the system user flag as per https://github.com/discord/discord-api-docs/issues/1994)

nekokatt: Hmm, isn't the system user still a bot/webhook?

hikari/events/message_events.py
```diff
@@ -367,16 +402,29 @@ def guild_id(self) -> snowflakes.Snowflake:
...
+            The member that sent the message, if known. This is a specialised
             implementation of `hikari.users.User`.
+
+            If the author was a webhook, then a `hikari.users.User` will be
+            returned instead, as webhooks do not have member objects.
         """
-        return typing.cast(guilds.Member, self.message.member)
+        member = self.message.member
+        if member is not None:
+            return member
```
FasterSpeeding, line 426: is that assignment to "member" necessary here (mypy should let you narrow the type of a model attribute within the scope of a function)?

hikari/events/message_events.py
FasterSpeeding: It should be but I feel like we could only be as sure of that as we could be sure that a webhook message would also be created by a bot. But this shouldn't matter too much, would be too hard to second guess what discord's going to do with system users anyway

hikari/events/message_events.py
nekokatt: partially do it to avoid any risk of raciness where something else could ever override that if a user uses a multithreaded application elsewhere. It isn't overly important, it is just slightly more efficient as it performs two less getattrs each time tbh.

FasterSpeeding (approved):
LGTM

hikari/events/message_events.py
nekokatt: We can resolve this another time when clarification is clearer I think.

PR #200, opened 2020-09-18
Fix inconsistecy in cache naming

davfsa:
### Checklist

### DIFF
--- a/hikari/api/cache.py
+++ b/hikari/api/cache.py
@@ -78,8 +78,8 @@ class Cache(abc.ABC):
     __slots__: typing.Sequence[str] = ()
 
     @abc.abstractmethod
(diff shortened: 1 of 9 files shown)


hikari/api/cache.py
```diff
@@ -78,7 +78,7 @@ class Cache(abc.ABC):
     __slots__: typing.Sequence[str] = ()
 
     @abc.abstractmethod
-    def get_dm(self, user_id: snowflakes.Snowflake, /) -> typing.Optional[channels.DMChannel]:
+    def get_dm(self, user_id: snowflakes.Snowflake, /) -> typing.Optional[channels.DM]:
```
nekokatt, line 81: I don't agree with this.

You've named it "DM" which means "direct message", when it is not a message at all, it is the channel a message is in.

It has "channel" in the name because it is a "direct message's channel".

If anything, `get_dm` should be renamed to `get_dm_channel` if we were being sensible about this

PR #202, opened 2020-09-19
Fix __all__ typehints

davfsa:
### Checklist


nekokatt:
why do we want to allow to mutate these now out of curiosity

davfsa:
> why do we want to allow to mutate these now out of curiosity

Mutating them at runtime doesn't change anything apart from the value of `__all__` itself, and I have found some linters (Pyright for example) that seem to complain when you set `__all__` to typing.Final. I can revert it if you want.

PR #203, opened 2020-09-20
Added all missing docs on REST

### LINKED ISSUE #42: Add missing documentation to REST
hikari.api.rest.RESTClient is missing a LOT of documentation still, and should be completed.

davfsa:
- Fix x_emoji typehints

### Checklist

### Related issues
Closes #42

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -86,9 +86,8 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
 
         Parameters
         ----------
-        channel : hikari.snowflakes.SnowflakeishOr
-            The channel to fetch. This may be a
-            `hikari.channels.PartialChannel` object, or the ID of an
+        channel : hikari.snowflakes.SnowflakeishOr[hikari.channels.PartialChannel]
+            The channel to fetch. This may be the object or the ID of an
             existing channel.
 
         Returns
@@ -122,6 +121,14 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If you are missing the `READ_MESSAGES` permission in the channel.
         hikari.errors.NotFoundError
             If the channel is not found.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
+            rate-limits automatically. This includes bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
         hikari.errors.InternalServerError
             If an internal error occurs on Discord while handling the request.
         """
@@ -150,7 +157,7 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         Parameters
         ----------
         channel : hikari.snowflakes.SnowflakeishOr[hikari.channels.GuildChannel]
-            The channel to edit. This may be a channel object, or the ID of an
+            The channel to edit. This may be the object or the ID of an
             existing channel.
         name : hikari.undefined.UndefinedOr[[builtins.str]
             If provided, the new name for the channel.
@@ -172,6 +179,7 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If provided, the new guild category for the channel.
         reason : hikari.undefined.UndefinedOr[builtins.str]
             If provided, the reason that will be recorded in the audit logs.
+            Maximum of 512 characters.
 
         Returns
         -------
@@ -188,6 +196,14 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If you are missing permissions to edit the channel.
         hikari.errors.NotFoundError
             If the channel is not found.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
+            rate-limits automatically. This includes bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
         hikari.errors.InternalServerError
             If an internal error occurs on Discord while handling the request.
         """  # noqa: E501 - Line too long
@@ -226,6 +242,14 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             channel.
         hikari.errors.NotFoundError
             If the origin or target channel is not found.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
+            rate-limits automatically. This includes bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
         hikari.errors.InternalServerError
             If an internal error occurs on Discord while handling the request.
         """
@@ -237,7 +261,7 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         Parameters
         ----------
         channel : hikari.snowflakes.SnowflakeishOr[hikari.channels.PartialChannel]
-            The channel to delete. This may be a channel object, or the ID of an
+            The channel to delete. This may be the object or the ID of an
             existing channel.
 
         Raises
@@ -248,6 +272,14 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If you are missing the `MANAGE_CHANNEL` permission in the channel.
         hikari.errors.NotFoundError
             If the channel is not found.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
+            rate-limits automatically. This includes bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
         hikari.errors.InternalServerError
             If an internal error occurs on Discord while handling the request.
 
@@ -301,11 +333,11 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         Parameters
         ----------
         channel : hikari.snowflakes.SnowflakeishOr[hikari.channels.GuildChannel]
-            The channel to edit a permission overwrite in. This may be a
-            channel object, or the ID of an existing channel.
+            The channel to edit a permission overwrite in. This may be the
+            object, or the ID of an existing channel.
         target : typing.Union[hikari.users.PartialUser, hikari.guilds.PartialRole, hikari.channels.PermissionOverwrite, hikari.snowflakes.Snowflakeish]
-            The channel overwrite to edit. This may be a overwrite object, or the ID of an
-            existing channel.
+            The channel overwrite to edit. This may be the object or the ID of an
+            existing overwrite.
         target_type : hikari.undefined.UndefinedOr[typing.Union[hikari.channels.PermissionOverwriteType, builtins.str]]
             If provided, the type of the target to update. If unset, will attempt to get
             the type from `target`.
@@ -315,6 +347,7 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If provided, the new vale of all disallowed permissions.
         reason : hikari.undefined.UndefinedOr[builtins.str]
             If provided, the reason that will be recorded in the audit logs.
+            Maximum of 512 characters.
 
         Raises
         ------
@@ -330,6 +363,14 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         hikari.errors.NotFoundError
             If the channel is not found or the target is not found if it is
             a role.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
+            rate-limits automatically. This includes bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
         hikari.errors.InternalServerError
             If an internal error occurs on Discord while handling the request.
         """  # noqa: E501 - Line too long
@@ -347,8 +388,8 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         Parameters
         ----------
         channel : hikari.snowflakes.SnowflakeishOr[hikari.channels.GuildChannel]
-            The channel to delete a permission overwrite in. This may be a
-            channel object, or the ID of an existing channel.
+            The channel to delete a permission overwrite in. This may be the
+            object, or the ID of an existing channel.
         target : typing.Union[hikari.users.PartialUser, hikari.guilds.PartialRole, hikari.channels.PermissionOverwrite, hikari.snowflakes.Snowflakeish]
             The channel overwrite to delete.
 
@@ -360,6 +401,14 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If you are missing the `MANAGE_PERMISSIONS` permission in the channel.
         hikari.errors.NotFoundError
             If the channel is not found or the target is not found.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
+            rate-limits automatically. This includes bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
         hikari.errors.InternalServerError
             If an internal error occurs on Discord while handling the request.
         """  # noqa: E501 - Line too long
@@ -389,6 +438,14 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If you are missing the `MANAGE_CHANNEL` permission in the channel.
         hikari.errors.NotFoundError
             If the channel is not found in any guilds you are a member of.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
+            rate-limits automatically. This includes bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
         hikari.errors.InternalServerError
             If an internal error occurs on Discord while handling the request.
         """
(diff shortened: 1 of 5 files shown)


hikari/api/rest.py
```diff
@@ -1289,6 +1284,11 @@ def fetch_reactions_for_emoji(
             If the channel or message is not found.
         hikari.errors.InternalServerError
             If an internal error occurs on Discord while handling the request.
+
+        !!! note
+            The exceptions on this endpoint will only be raised once the
+            result is awaited or interacted with. Invoking this function
```
FasterSpeeding, line 1290: might be better to say "iterated over" here as "interacting" with an object may just mean looking at attributes

hikari/api/rest.py
```diff
@@ -1346,10 +1346,10 @@ async def fetch_webhook(
         Parameters
         ----------
         webhook : hikari.snowflakes.SnowflakeishOr[hikari.webhooks.Webhook]
-            The webhook to fetch. This may be a `hikari.webhooks.Webhook`
-            or the ID of an existing webhook.
+            The webhook to fetch. This may be a webhook object or the ID
```
FasterSpeeding, line 1349: Could these bits not follow a more brief less repetitive format along the lines of "the object or ID of an existing webook"

hikari/api/rest.py
```diff
@@ -1766,6 +1766,8 @@ async def edit_my_user(
 
         Raises
         ------
+        hikari.errors.BadRequestError
```
FasterSpeeding, line 1769: Might be worth it documenting the behaviour here with the sub-ratelimits which iirc exist and would lead to the ratelimited error being raised

FasterSpeeding: Those should also be documented for places like edit channel and execute webhook if they aren't already

hikari/api/rest.py
```diff
@@ -2011,23 +2022,105 @@ def fetch_audit_log(
         user: undefined.UndefinedOr[snowflakes.SnowflakeishOr[users.PartialUser]] = undefined.UNDEFINED,
         event_type: undefined.UndefinedOr[audit_logs.AuditLogEventType] = undefined.UNDEFINED,
     ) -> iterators.LazyIterator[audit_logs.AuditLog]:
-        ...
+        """Fetch the guilds audit log.
```
FasterSpeeding, line 2025: `guild's`

hikari/api/rest.py
```diff
@@ -2011,23 +2022,105 @@ def fetch_audit_log(
...
+
+        Returns
+        -------
+        hikari.emojis.KnownCustomEmoji
+            The requested emoji.
+
+        Raises
+        ------
+        hikari.errors.NotFoundError
+            If the guild or the emoji are not found.
```
FasterSpeeding, line 2091: or if you aren't in the guild maybe? not sure if the nuanced difference between a guild not being found vs a guild not being found in the group of guilds you have access to would be worth explicitly documenting both cases just felt like pointing that out.

hikari/api/rest.py
```diff
@@ -2041,46 +2134,211 @@ async def create_emoji(
...
+
+        Parameters
+        ----------
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            The guild to create the emoji on. This can be a
+            guild object or the ID of an existing guild.
+        name : builtins.str
+            The name for the emoji.
+        image : hikari.files.Resourceish
+            The image for the emoji.
```
FasterSpeeding, line 2147: Might be worth it documenting file limitations here.

hikari/api/rest.py
```diff
@@ -2041,46 +2134,211 @@ async def create_emoji(
...
+        Parameters
+        ----------
+        name : builtins.str
+            The new guilds name.
+
+        Returns
+        -------
+        hikari.api.special_endpoints.GuildBuilder
+            The guild builder to use. This will allow to create a guild
+            later with `hikari.api.special_endpoints.GuildBuilder.create`.
```
FasterSpeeding, line 2267: is awaiting the object not alsoa valid way to create the guild?

hikari/api/rest.py
```diff
@@ -2114,7 +2372,69 @@ async def edit_guild(
...
+        icon : hikari.undefined.UndefinedOr[hikari.files.Resourceish]
+            If provided, the new guild icon.
+        owner : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishOr[hikari.users.PartialUser]]
+            If provided, the new guild owner.
+
+            !!! warn
+                You need to be the owner of the server to use this.
+        splash : hikari.undefined.UndefinedNoneOr[hikari.files.Resourceish]
+            If provided, the new guild splash.
+        banner : hikari.undefined.UndefinedNoneOr[hikari.files.Resourceish]
```
FasterSpeeding, line 2406: Document file restrictions?

hikari/api/rest.py
```diff
@@ -2494,21 +3467,92 @@ async def fetch_guild_voice_regions(
...
+            If an internal error occurs on Discord while handling the request.
+        """
 
     @abc.abstractmethod
     async def fetch_guild_invites(
         self,
         guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
     ) -> typing.Sequence[invites.InviteWithMetadata]:
-        ...
+        """Fetch the guilds invites.
```
FasterSpeeding, line 3502: `guild's`

hikari/api/rest.py
```diff
@@ -2494,21 +3467,92 @@ async def fetch_guild_voice_regions(
...
+            If an internal error occurs on Discord while handling the request.
+        """
 
     @abc.abstractmethod
     async def fetch_integrations(
         self,
         guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
     ) -> typing.Sequence[guilds.Integration]:
-        ...
+        """Fetch the guilds integrations.
```
FasterSpeeding, line 3532: `guild's`

hikari/api/rest.py
```diff
@@ -2518,11 +3562,55 @@ async def delete_integration(
         *,
         reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
     ) -> None:
-        ...
+        """Fetch the guilds integrations.
```
FasterSpeeding, line 3565: `guild's`

hikari/api/rest.py
```diff
@@ -2160,28 +2501,116 @@ async def create_guild_text_channel(
...
+            The channels name.
+        position : hikari.undefined.UndefinedOr[builtins.int]
+            If provided, the position of the channel (relative to the
+            category, if any).
+        topic : hikari.undefined.UndefinedOr[builtins.str]
+            If provided, the channels topic.
+        nsfw : hikari.undefined.UndefinedOr[builtins.bool]
+            If provided, whether to mark the channel as NSFW.
+        rate_limit_per_user : hikari.undefined.UndefinedOr[builtins.int]
+            If provided, the rate limit per user.
```
FasterSpeeding, line 2521: This doesn't really tell us what this field actually is unit wise or what it's accepted range is.
Another thought, wouldn't it make sense to also accept timedeltas for fields like this (especially as that's how we should be storing time deltas on models)?

hikari/api/rest.py
```diff
@@ -2160,28 +2501,116 @@ async def create_guild_text_channel(
...
+        permission_overwrites : hikari.undefined.UndefinedOr[typing.Sequence[hikari.channels.PermissionOverwrite]]
+            If provided, the permission overwrites for the channel.
+        category : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishOr[hikari.channels.GuildCategory]]
+            The category to create the channel under. This may be a
+            `hikari.channels.GuildCategory` or the ID of an existing
+            category.
+        reason : hikari.undefined.UndefinedOr[builtins.str]
+            If provided, the reason that will be recorded in the audit logs.
```
FasterSpeeding, line 2529: Do we document the length limit for reason anywhere?

hikari/api/rest.py
```diff
@@ -2195,7 +2624,49 @@ async def create_guild_voice_channel(
...
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            The guild to create the channel in. This may be a
+            guild object or the ID of an existing guild.
+        name : builtins.str
+            The channels name.
+        position : hikari.undefined.UndefinedOr[builtins.int]
+            If provided, the position of the channel (relative to the
+            category, if any).
+        user_limit : hikari.undefined.UndefinedOr[builtins.int]
+            If provided, the user limit for the channel.
```
FasterSpeeding, line 2640: Could be more specific around this being a limit for how many users can be in the channel.

hikari/api/rest.py
```diff
@@ -2195,7 +2624,49 @@ async def create_guild_voice_channel(
...
+        bitrate : hikari.undefined.UndefinedOr[builtins.int]
+            If provided, the bitrate for the channel.
```
FasterSpeeding, line 2642: Valid range?

hikari/api/rest.py
```diff
@@ -2209,29 +2680,138 @@ async def create_guild_category(
...
-        ...
+        """Create a category in a guild.
+
+        Parameters
+        ----------
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            The guild to create the channel in. This may be a
+            guild object or the ID of an existing guild.
+        name : builtins.str
+            The channels name.
```
FasterSpeeding, line 2691: document the valid range (or anything else i'm missing) maybe?

hikari/api/rest.py
```diff
@@ -2209,29 +2680,138 @@ async def create_guild_category(
...
+        """Reposition the channels in a guild.
+
+        Parameters
+        ----------
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            The guild to reposition the channels in. This may be a
+            guild object or the ID of an existing guild.
+        positions : typing.Mapping[builtins.int, hikari.snowflakes.SnowflakeishOr[hikari.channels.GuildChannel]]
+            A mapping of the position relative to the category (if any)
+            to the channel.
```
FasterSpeeding, line 2733: to the category (if any) to the channel? (might be worth saying something along the lines of to the object or ID of the channels to repositions)

hikari/api/rest.py
```diff
@@ -2209,29 +2680,138 @@ async def create_guild_category(
...
+        """Fetch a guild member.
+
+        Parameters
+        ----------
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            The guild to get the member from. This may be a
+            guild object or the ID of an existing guild.
+        user : hikari.snowflakes.SnowflakeishOr[hikari.users.PartialUser]
+            The user to get the member for. This may be a
+            `hikari.users.PartialUser` or the ID of an existing user.
```
FasterSpeeding, line 2762: This extra link seems unnecessary.

hikari/api/rest.py
```diff
@@ -2209,29 +2680,138 @@ async def create_guild_category(
...
+        hikari.errors.InternalServerError
+            If an internal error occurs on Discord while handling the request.
+        """
 
     @abc.abstractmethod
     def fetch_members(
         self, guild: snowflakes.SnowflakeishOr[guilds.PartialGuild]
     ) -> iterators.LazyIterator[guilds.Member]:
-        ...
+        """Fetch reactions for an emoji from a message.
```
FasterSpeeding, line 2783: this description doesn't fit this endpoint

hikari/api/rest.py
```diff
@@ -2250,7 +2830,55 @@ async def edit_member(
...
+        roles : hikari.undefined.UndefinedOr[typing.Collection[hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialRole]]]
+            If provided, the new roles for the member. Requires the
+            `MANAGE_ROLES` permission.
+        mute : hikari.undefined.UndefinedOr[builtins.bool]
+            If provided, the new server mute state for the member.
+            Requires the `MUTE_MEMBERS` permission.
+        deaf : hikari.undefined.UndefinedOr[builtins.bool]
+            If provided, the new server deaf state for the member.
+            Requires the `DEAFEN_MEMBERS` permission.
+        voice_channel : hikari.undefined.UndefinedOr[builtins.bool]
```
FasterSpeeding, line 2856: why is this bool?

hikari/api/rest.py
```diff
@@ -2260,7 +2888,30 @@ async def edit_my_nick(
         *,
         reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
     ) -> None:
-        ...
+        """Edit the associated token's member nick.
```
FasterSpeeding, line 2891: Something along the lines of "Edit the associated token's nickname in the provided guild." might be a bit easier to read.

hikari/api/rest.py
```diff
@@ -2330,21 +3107,91 @@ async def fetch_ban(
         guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
         user: snowflakes.SnowflakeishOr[users.PartialUser],
     ) -> guilds.GuildMemberBan:
-        ...
+        """Fetch a users ban.
```
FasterSpeeding, line 3110: `"user's ban"` and might be worth saying something along the line of "Fetch the state of a user's ban." as "Fetch a user's ban" is a bit ambiguous.

hikari/api/rest.py
```diff
@@ -2330,21 +3107,91 @@ async def fetch_ban(
...
+            The guild to fetch the ban from. This may be a
+            guild object or the ID of an existing guild.
+        user : hikari.snowflakes.SnowflakeishOr[hikari.users.PartialUser]
+            The user to fetch the ban of. This may be a
+            user object or the ID of an existing user.
+
+        Returns
+        -------
+        hikari.guilds.GuildMemberBan
+            The requested ban.
```
FasterSpeeding, line 3124: requested ban state maybe

hikari/api/rest.py
```diff
@@ -2359,15 +3206,76 @@ async def create_role(
...
+        ----------
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            The guild to create the role in. This may be a
+            guild object or the ID of an existing guild.
+        name : hikari.undefined.UndefinedOr[builtins.str]
+            If provided, the name for the role.
+        permissions : hikari.undefined.UndefinedOr[hikari.permissions.Permissions]
+            If provided, the permissions for the role.
+        color : hikari.undefined.UndefinedOr[hikari.colors.Colorish]
+            If provided, the roles color.
```
FasterSpeeding, line 3221: `"role's"` (cause posessive)

hikari/api/rest.py
davfsa: Sometimes there are, sometimes they aren't. I'm pretty sure I documented them all

hikari/api/rest.py
davfsa: No, you have to call the create function to create it

FasterSpeeding: The shouldn't the 
```
        !!! note
            The exceptions on this endpoint will only be raised once the result
            is awaited or interacted with. Invoking this function itself will
            not raise any of the above types.
```

be more specific for this endpoint?

hikari/api/rest.py
davfsa: No, since discord doesnt document it themselves (if there is, havent seen it).

hikari/api/rest.py
davfsa: Forgot to remove it

hikari/api/rest.py
FasterSpeeding: https://github.com/discord/discord-api-docs/commit/e0b3bf3e6499cd9121e9a77af36fff798820bdf9

hikari/api/rest.py
davfsa: Bans dont have states, they are just entries. Maybe I will change it to `Fetch a ban` or something along the lines

hikari/api/rest.py
davfsa: Oh, didnt see that, will add

hikari/api/rest.py
FasterSpeeding: I mean an entry is just a state but we use the word "ban" as both a noun and a verb which might get confusing (think of "ban_user" as opposed to "create_ban" in some places but in other places fetch_ban)

hikari/api/rest.py
```diff
@@ -727,31 +727,31 @@ async def create_message(
...
-            If specified, the message attachments. These can be resources, or
+            If provided, the message attachments. These can be resources, or
             strings consisting of paths on your computer or URLs.
         tts : hikari.undefined.UndefinedOr[builtins.bool]
-            If specified, whether the message will be TTS (Text To Speech).
+            If provided, whether the message will be TTS (Text To Speech).
         nonce : hikari.undefined.UndefinedOr[builtins.str]
-            If specified, a nonce that can be used for optimistic message
+            If provided, a nonce that can be used for optimistic message
             sending.
```
nekokatt, line 741:
```suggestion
            An arbitrary identifier to associate with the message. This
            can be used to identify it later in received events. If provided,
            this must be less than 32 bytes. If not provided, then 
            a null value is placed on the message instead. All users can
            see this value.
```

hikari/api/rest.py
```diff
@@ -727,31 +727,31 @@ async def create_message(
...
         mentions_everyone : hikari.undefined.UndefinedOr[builtins.bool]
-            If specified, whether the message should parse @everyone/@here
+            If provided, whether the message should parse @everyone/@here
             mentions.
         user_mentions : hikari.undefined.UndefinedOr[typing.Union[typing.Collection[hikari.snowflakes.SnowflakeishOr[hikari.users.PartialUser]], builtins.bool]]
-            If specified, and `builtins.True`, all mentions will be parsed.
-            If specified, and `builtins.False`, no mentions will be parsed.
+            If provided, and `builtins.True`, all mentions will be parsed.
```
nekokatt, line 746: I don't like this wording. Parsed to...what?

We should say something like `If provided and builtins.True, all types of mentions will be detected. If builtins.False, all mention types will ignored if appearing in the message body.` or something a bit clearer, imho.

hikari/api/rest.py
```diff
@@ -882,7 +882,7 @@ async def edit_message(
             a `hikari.channels.TextChannel` or the ID of an existing
             channel.
         message : hikari.snowflakes.SnowflakeishOr[hikari.messages.Message]
-            The message to edit. This may be a `hikari.messages.Message`
+            The message to edit. This may be a message object
```
nekokatt, line 885: define "message object"? Does that include partial messages? (The answer is yes).

hikari/api/rest.py
```diff
@@ -1117,8 +1117,7 @@ async def delete_my_reaction(
         ----------
         channel : hikari.snowflakes.SnowflakeishOr[hikari.channels.TextChannel]
             The channel where the message to delete the reaction from is.
-            This may be a `hikari.channels.TextChannel` or the ID of an
-            existing channel.
+            This may be a channel object or the ID of an existing channel.
```
nekokatt, line 1120: It can't be a voice channel, even if that exists.

hikari/api/rest.py
davfsa: I dont think `edit_my_user` has any sub-ratelimits. iirc, its only `edit_channel` that has them

FasterSpeeding: edit_my_user is actually an outlier in the fact that it will return a 400 instead of a 429 if you change fields too quickly

nekokatt:
Lgtm, but a few little things:

- Can we change `pre-emtively` to `preemptively`? (my bad, i cant spell).
- Can we update that ratelimit paragraph to read this:

```suggestion
Usually, Hikari will handle and retry on hitting rate-limits automatically. 
This includes bucket-specific rate-limits and global rate-limits. In some
rare edge cases, however, Discord implements other undocumented rules
for rate-limiting, such as limits per attribute. These cannot be detected or
handled normally by Hikari due to their undocumented nature, and will
trigger this exception if they occur.
```

nekokatt (approved):

nekokatt:
Tests should be fixed on master, this isn't due to what you changed, so I am happy to force merge.

PR #205, opened 2020-09-20
Remove blocked off integration endpoints

davfsa:
Looks like discord blocked off some integration endpoints to prevent bots from using them.

### Checklist

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -2511,33 +2511,12 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         ...
 
     @abc.abstractmethod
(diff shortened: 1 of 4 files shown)


FasterSpeeding (approved):
lgtm

nekokatt:
So this is blocked off for bots, but we also support OAuth2 apps, cant integrations use these too?

FasterSpeeding:
> So this is blocked off for bots, but we also support OAuth2 apps, cant integrations use these too?
Well these endpoints are all guild bound and outside the scope of discord's oauth2 flow, the oauth2 equivalent of these is fetch_my_connections.

PR #211, opened 2020-09-21
Improve `fetch_ban` to return None if the user ban is not found

### LINKED ISSUE #209: Differentiate not found entry from not found guild or user in `fetch_ban`
### Summary
Differentiate not found entry from not found guild or user in `fetch_ban`

### Problem
Unable to differentiate if the user is not found or if you just provided an unknown user/guild due to how discord implements this

### Ideal implementation
TBD

davfsa:
Also works on v8 of the REST

### Checklist

### Related issues
Close #209

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -2329,7 +2329,7 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         self,
         guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
         user: snowflakes.SnowflakeishOr[users.PartialUser],
-    ) -> guilds.GuildMemberBan:
+    ) -> typing.Optional[guilds.GuildMemberBan]:
         ...
 
     @abc.abstractmethod
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -2095,9 +2095,20 @@ class RESTClientImpl(rest_api.RESTClient):
         self,
         guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
         user: snowflakes.SnowflakeishOr[users.PartialUser],
-    ) -> guilds.GuildMemberBan:
+    ) -> typing.Optional[guilds.GuildMemberBan]:
         route = routes.GET_GUILD_BAN.compile(guild=guild, user=user)
-        raw_response = await self._request(route)
+        try:
+            raw_response = await self._request(route)
+        except errors.NotFoundError as ex:
+            # Discord raises an error when the ban is not found.
+            # We simplify this for the user and only raise the
+            # NotFoundError if it is something else not being found.
+            json_body = typing.cast(data_binding.JSONObject, data_binding.load_json(ex.raw_body))
+            if json_body["code"] == 10026:
+                return None
+
+            raise
+
         response = typing.cast(data_binding.JSONObject, raw_response)
         return self._entity_factory.deserialize_guild_member_ban(response)
 
--- a/tests/hikari/impl/test_rest.py
+++ b/tests/hikari/impl/test_rest.py
@@ -43,6 +43,7 @@ from hikari import users
 from hikari.impl import entity_factory
 from hikari.impl import rest
 from hikari.impl import special_endpoints
+from hikari.utilities import data_binding
 from hikari.utilities import net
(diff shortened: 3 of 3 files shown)


hikari/impl/rest.py
```diff
@@ -2095,9 +2095,20 @@ async def fetch_ban(
...
         route = routes.GET_GUILD_BAN.compile(guild=guild, user=user)
-        raw_response = await self._request(route)
+        try:
+            raw_response = await self._request(route)
+        except errors.NotFoundError as ex:
+            # Discord raises an error when the ban is not found.
+            # We simplify this for the user and only raise the
+            # NotFoundError if it is something else not being found.
+            json_body = typing.cast(data_binding.JSONObject, data_binding.load_json(ex.raw_body))
+            if json_body["code"] == 10026:
```
nekokatt, line 2107: We should define an enum with this code in if we add any more in the future.

hikari/impl/rest.py
```diff
@@ -2095,9 +2095,20 @@ async def fetch_ban(
...
+    ) -> typing.Optional[guilds.GuildMemberBan]:
         route = routes.GET_GUILD_BAN.compile(guild=guild, user=user)
-        raw_response = await self._request(route)
+        try:
+            raw_response = await self._request(route)
+        except errors.NotFoundError as ex:
+            # Discord raises an error when the ban is not found.
+            # We simplify this for the user and only raise the
+            # NotFoundError if it is something else not being found.
+            json_body = typing.cast(data_binding.JSONObject, data_binding.load_json(ex.raw_body))
```
nekokatt, line 2106: Should we be doing this for other endpoints too, then?

I wonder if this is a little inconsistent with other endpoints, is all.

I wonder whether it would be worth adding an optional `type` field to `NotFoundError` instead so we can flag to the user what is missing.

Maybe something like uh...

```py
class NotFoundError(...):
    type: typing.Optional[typing.Type[typing.Any]]
```
and then
```py
if json_body["code"] == xxxx:
    raise NotFoundError(guilds.GuildMemberBan)
```
I'm not sure.

On the other hand, your way makes sense in my head. I'm just not sure if we should raise in some cases but not others. I'll have a ponder.

nekokatt: Actually, I think we should approach this differently.

We should get Discord to verify the basic format for exceptions, so we know it is behaviour we can use without risking it breaking. I'll assume we can only expect a code in it and should ignore anything else.

We should then add the JSON error code enum to errors.py. The code should then be an optional field on any HTTP exception. This will allow users to instead check the exact nature of the error and this will then be applicable to all endpoints.

I think that is the best way forwards with this. It would also possibly allow us to tidy up the str/repr for http errors.

hikari/api/rest.py
```diff
@@ -2329,7 +2329,7 @@ async def fetch_ban(
         self,
         guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
         user: snowflakes.SnowflakeishOr[users.PartialUser],
-    ) -> guilds.GuildMemberBan:
+    ) -> typing.Optional[guilds.GuildMemberBan]:
```
nekokatt, line 2332: You'd need to update your docs in the other PR then right?

nekokatt (changes requested):
I think we should instead parse the error codes in the HTTP error handler if possible and add an enum of all the JSON error codes, remembering that some may not be defined (best effort, fall back to raw int if unable to resolve?)

hikari/impl/rest.py
davfsa: Could do, but since this is the only one where this happens. One thing I thought about was maybe having a `code` property in the errors that will return the code. Will have a think 👍

hikari/api/rest.py
davfsa: Yea, at the time I did this commit I couldnt rebase to master. Its the reason I left it as a draft 😄

hikari/impl/rest.py
nekokatt: Special cases arent special enough to break the rules, right?

davfsa:
Blocked by #213

davfsa:
Have decided to go another way with this. Will open another pr with the changes :+1:

PR #213, opened 2020-09-21
Improve error str and added code + message

davfsa:
### Checklist

### Related issues
Unblocks #211

### DIFF
--- a/hikari/errors.py
+++ b/hikari/errors.py
@@ -198,19 +198,6 @@ class HTTPClientClosedError(HTTPError):
     """The error message."""
 
 
-_message_default_factory = attr.Factory(
-    lambda self: f"{self.status}: {self.raw_body}" if self.reason is None else self.reason, takes_self=True
-)
-
-
-@attr.s(auto_exc=True, slots=True, repr=False, weakref_slot=False)
-class HTTPErrorResponse(HTTPError):
-    """Base exception for an erroneous HTTP response."""
-
-    url: str = attr.ib()
-    """The URL that produced this error message."""
-
-
 @attr.s(auto_exc=True, slots=True, repr=False, weakref_slot=False)
 class HTTPResponseError(HTTPError):
     """Base exception for an erroneous HTTP response."""
@@ -227,27 +214,30 @@ class HTTPResponseError(HTTPError):
     raw_body: typing.Any = attr.ib()
     """The response body."""
 
-    reason: typing.Optional[str] = attr.ib(default=None)
-    """The error reason. If `builtins.None`, will generate one automatically."""
-
-    message: str = attr.ib(default=_message_default_factory, init=False)
+    message: str = attr.ib(default="")
     """The error message."""
 
+    code: int = attr.ib(default=0)
+    """The error code."""
+
     def __str__(self) -> str:
-        try:
-            raw_body = self.raw_body.decode("utf-8")
-        except (AttributeError, UnicodeDecodeError):
-            raw_body = str(self.raw_body)
-
-        chomped = len(raw_body) > 200
-
         if isinstance(self.status, http.HTTPStatus):
             name = self.status.name.replace("_", " ").title()
             name_value = f"{name} {self.status.value}"
         else:
             name_value = str(self.status).title()
 
-        return f"{name_value}: {raw_body[:200]}{'...' if chomped else ''} for {self.url}"
+        if self.message:
+            body = self.message
+        else:
+            try:
+                body = self.raw_body.decode("utf-8")
+            except (AttributeError, UnicodeDecodeError, TypeError, ValueError):
+                body = str(self.raw_body)
+
+        chomped = len(body) > 200
+
+        return f"{name_value}: '{body[:200]}{'...' if chomped else ''}' for {self.url}"
 
 
 @attr.s(auto_exc=True, slots=True, repr=False, weakref_slot=False)
@@ -266,9 +256,6 @@ class BadRequestError(ClientHTTPResponseError):
     status: http.HTTPStatus = attr.ib(default=http.HTTPStatus.BAD_REQUEST, init=False)
     """The HTTP status code for the response."""
 
-    message: str = attr.ib(default=_message_default_factory, init=False)
-    """The error message."""
-
 
 @attr.s(auto_exc=True, slots=True, repr=False, weakref_slot=False)
 class UnauthorizedError(ClientHTTPResponseError):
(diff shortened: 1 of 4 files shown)


hikari/errors.py
```diff
@@ -227,27 +214,31 @@ class HTTPResponseError(HTTPError):
...
+            body = self.message
+        else:
+            try:
+                body = self.raw_body.decode("utf-8")
+            except (AttributeError, UnicodeDecodeError):
+                body = str(self.raw_body)
+
+        chomped = len(body) > 200
+
+        return f"{name_value}: '{body[:200]}{'...' if chomped else ''}' for {self.url}"
```
nekokatt, line 241: smart

nekokatt (dismissed):

hikari/errors.py
```diff
@@ -227,27 +214,31 @@ class HTTPResponseError(HTTPError):
...
     """The error message."""
 
-    def __str__(self) -> str:
-        try:
-            raw_body = self.raw_body.decode("utf-8")
-        except (AttributeError, UnicodeDecodeError):
-            raw_body = str(self.raw_body)
-
-        chomped = len(raw_body) > 200
+    code: typing.Union[int] = attr.ib(default=0)
```
FasterSpeeding, line 221: union int?

davfsa: Oh, my bad, I used to have another thing there, will fix :sweat_smile:

hikari/errors.py
```diff
@@ -227,27 +214,31 @@ class HTTPResponseError(HTTPError):
     raw_body: typing.Any = attr.ib()
     """The response body."""
 
-    reason: typing.Optional[str] = attr.ib(default=None)
-    """The error reason. If `builtins.None`, will generate one automatically."""
-
-    message: str = attr.ib(default=_message_default_factory, init=False)
+    # We want to make it optional here, so ignore mypy complaining
+    message: typing.Optional[str] = attr.ib(default=None)  # type: ignore[assignment]
```
nekokatt, line 218: This is bad. It breaks typesafety by messing up the contract on the supertype.

We really shouldn't be doing this, cause you'll get an HttpError and expect the message to be a string when you catch it, and youll get None instead.

nekokatt (dismissed):

hikari/errors.py
davfsa: Didn't think about that, will set it to `""` instead of None as a default

nekokatt: Why not default to a generic message?

davfsa: This is done in `__str__` anyway, so really no need to complicate it even further and make initializing a error slower :)

hikari/errors.py
```diff
@@ -227,27 +214,30 @@ class HTTPResponseError(HTTPError):
...
         else:
             name_value = str(self.status).title()
 
-        return f"{name_value}: {raw_body[:200]}{'...' if chomped else ''} for {self.url}"
+        if self.message:
+            body = self.message
+        else:
+            try:
+                body = self.raw_body.decode("utf-8")
+            except (AttributeError, UnicodeDecodeError):
```
FasterSpeeding, line 235: This is probably a non-issue but seeing as raw_body is typed as any, we don't have any guarantee that if it has a "decode" method it'll be compatible with that call signature, was wondering if it might be worth it also catching TypeError and ValueError here to cover every case

hikari/utilities/net.py
```diff
@@ -39,18 +39,24 @@ async def generate_error_response(response: aiohttp.ClientResponse) -> errors.HT
...
+        cls = errors.BadRequestError
+    elif response.status == http.HTTPStatus.UNAUTHORIZED:
+        cls = errors.UnauthorizedError
+    elif response.status == http.HTTPStatus.FORBIDDEN:
+        cls = errors.ForbiddenError
+    elif response.status == http.HTTPStatus.NOT_FOUND:
+        cls = errors.NotFoundError
+
+    if cls is not None:
+        json_body = await response.json()
```
FasterSpeeding, line 54: Will this be reached and error out if we hit a cloudflare error? (might be worth it defaulting the expected json body fields to "Unknown"

hikari/utilities/net.py
```diff
@@ -39,18 +39,24 @@ async def generate_error_response(response: aiohttp.ClientResponse) -> errors.HT
...
+        # False positive
+        return cls(real_url, response.headers, raw_body, json_body["message"], json_body["code"])  # type: ignore[call-arg]
```
FasterSpeeding, line 56: which rest api version are we targetting here? I'm not sure about v7 and v8 but I know for sure that not all sane error response will have "message" and "code" in their body on v6 (a good example of this is when you try to pass an invalid embed on create message or execute webhook)

hikari/utilities/net.py
davfsa: This wont get reached since a cloudfare error will not return any of the status there is above

hikari/utilities/net.py
davfsa: The docs state that all the errors will include a `code` and `message` along with the status. Will have a look at those exceptions you said

FasterSpeeding: The docs sate "an also return more detailed error codes" where emphasis should be put on the usage of "can" as opposed to "will", there's no real guarantee there

hikari/errors.py
davfsa: Will add them for good measure :+1:

hikari/utilities/net.py
davfsa: Ok, will change so that it doesnt assume that they send `message` nor `code` and will also fix the other thing you pointed out with cloudfare

hikari/utilities/net.py
FasterSpeeding: Well, you're also covering 400 here which in-theory shouldn't ever be hit if we have enough checks on our end but in rare cases you can get a 400 from services in-between us and and Discord. I can think of two cases of this off the top of my head, when you try to include a payload in a GET request (google will response here) and If you made a request that was too large (I'm sure you get this if the url is too long and that'd likely be caused by including too many elements in a query string but haven't tested to see how this behaves with payload size).

FasterSpeeding (approved):
LGTM

FasterSpeeding (approved):
Assuming those tests pass, LGTM 2 electric-boogaloo

FasterSpeeding (approved):
All work and no play makes Johnny Look Good To Me

nekokatt (approved):

PR #201, opened 2020-09-19
Task/enum

nekokatt:
FIxes #151, reimplements enum to be faster.

```
BasicPyEnum.__call__('25') 0.43657039100071415 µs
BasicHikariEnum.__call__('25') 0.19885178900040046 µs
BasicPyEnum._value2member_map_['25'] 0.09891081800014945 µs
BasicHikariEnum._value2member_map['25'] 0.1033127640002931 µs
BasicPyEnum.__getitem__['z'] 0.16736914299872296 µs
BasicHikariEnum.__getitem__['z'] 0.15685381999901438 µs
BasicPyEnum.__call__ profile
         2000003 function calls in 0.744 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.249    0.249    0.744    0.744 <string>:1(<module>)
  1000000    0.274    0.000    0.495    0.000 enum.py:283(__call__)
  1000000    0.221    0.000    0.221    0.000 enum.py:562(__new__)
        1    0.000    0.000    0.744    0.744 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


BasicHikariEnum.__call__ profile
         1000003 function calls in 0.340 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.183    0.183    0.340    0.340 <string>:1(<module>)
  1000000    0.157    0.000    0.157    0.000 enums.py:115(__call__)
        1    0.000    0.000    0.340    0.340 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


BasicPyEnum.__getitem__ profile
         1000003 function calls in 0.331 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.165    0.165    0.331    0.331 <string>:1(<module>)
  1000000    0.166    0.000    0.166    0.000 enum.py:348(__getitem__)
        1    0.000    0.000    0.331    0.331 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


BasicHikariEnum.__getitem__ profile
         1000003 function calls in 0.318 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
[...]

PR #214, opened 2020-09-22
Faster enum impl

### LINKED ISSUE #151: Write enum meta replacement
EnumMeta in the standard lib is slow, and during deserialization we invoke this a lot.

We should have a custom impl such that;

```py
if TYPE_CHECKING:
    EnumMeta = enum.EnumMeta
    Enum = enum.Enum
    ...
else:
    class EnumMeta(...):
        ...

    class Enum(...):
        ...
```
These should have a similar interface but be optimised where possible for speed.

Benchmarking is a must.

### LINKED ISSUE #182: Flexibility around enum values.
### Summary
The idea is to possibly avoid discord from breaking older versions of the lib when they instantaneously release new enum values with zero warning by trying to cast the value to the enum and falling back to using the raw value rather than letting the value error propagate leading to the event being dropped as this will effect stuff like cache reliability and may lead to events being dropped over arguably insignificant values.

### Problem
Discord doesn't know how versioning work.

### Ideal implementation
enum fields (excluding int flags as these don't use range check validation) will be a union of the enum or the raw value we get from discord (either string or integer) and will fall back to the raw value if they couldn't be casted.

### Checklist

### LINKED ISSUE #212: Stop using enums for non discrete values
This relates to #182, which is partially touched by #208.

We currently use enum types to represent constants. This is not ideal since Discord are... let's say not very consistent... at documenting what their API produces. They will add new fields and values without warning. This defeats the purpose of using an enum type to represent these in a way that is typesafe.

Instead, I propose we make a mechanism to handle this in a similar way. 

One idea I have had is the idea being we create a class via a metaclass we define that gets called with a string value. It can then either return an existing value if cached (store values in a dict, I guess, since primitiveish types cant be weakreffed nicely), or return the value it is called with. This would be efficient in terms of memory since stuff like strings could be interned to prevent making multiple copies of the same string in memory, possibly.

We may have to define this differently for ints and strings, since ints get interned differently (automatically for the first few thousand integers around zero), whereas strings need explicit sys.intern-ing for that to work.

If this is not MyPy compatible, we can always fall back to instead having a non-instantiating class that simply acts as a namespace to hold common values in.

This would not need to apply to Flag enum types like Intents and Permissions which already behave like this.

Another possibility is using the Flag type for values like the GuildFeature type in some cases, and latebinding it somehow. This would require the hikari.utilities.flags module to be changed to work with all flag types, not just IntFlag; use of enums is slower we need to remember, and unlike an int flag, there is the complexity of storing non-late-bound values in a list until we convert them.

If anyone has any feedback, feel free to pop a comment. I am just tossing ideas around here.

davfsa:
- If a value is failed to be found, will just return the inputed value. This way, if Discord adds new fields, it wont break.

Co-authored-by: @nekokatt

### Checklist

### Related issues
Closes #182
Closes #151

# Performance metrics
```
BasicPyEnum.__call__('25') 0.43657039100071415 µs
BasicHikariEnum.__call__('25') 0.19885178900040046 µs
BasicPyEnum._value2member_map_['25'] 0.09891081800014945 µs
BasicHikariEnum._value2member_map['25'] 0.1033127640002931 µs
BasicPyEnum.__getitem__['z'] 0.16736914299872296 µs
BasicHikariEnum.__getitem__['z'] 0.15685381999901438 µs
BasicPyEnum.__call__ profile
         2000003 function calls in 0.744 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.249    0.249    0.744    0.744 <string>:1(<module>)
  1000000    0.274    0.000    0.495    0.000 enum.py:283(__call__)
  1000000    0.221    0.000    0.221    0.000 enum.py:562(__new__)
        1    0.000    0.000    0.744    0.744 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


BasicHikariEnum.__call__ profile
         1000003 function calls in 0.340 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.183    0.183    0.340    0.340 <string>:1(<module>)
  1000000    0.157    0.000    0.157    0.000 enums.py:115(__call__)
        1    0.000    0.000    0.340    0.340 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


BasicPyEnum.__getitem__ profile
         1000003 function calls in 0.331 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.165    0.165    0.331    0.331 <string>:1(<module>)
  1000000    0.166    0.000    0.166    0.000 enum.py:348(__getitem__)
        1    0.000    0.000    0.331    0.331 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


BasicHikariEnum.__getitem__ profile
         1000003 function calls in 0.318 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.159    0.159    0.318    0.318 <string>:1(<module>)
  1000000    0.158    0.000    0.158    0.000 enums.py:141(__getitem__)
        1    0.000    0.000    0.318    0.318 {built-in method builtins.exec}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```


nekokatt (approved):

FasterSpeeding:
Do we have some performance metrics on how faster this is?

davfsa:
> Do we have some performance metrics on how faster this is compared to the stdlib enum class?

Edited the top comment to include them

hikari/impl/entity_factory.py
```diff
@@ -424,10 +421,7 @@ def deserialize_audit_log(self, payload: data_binding.JSONObject) -> audit_log_m
                 user_id = snowflakes.Snowflake(raw_user_id)
 
             action_type: typing.Union[audit_log_models.AuditLogEventType, int]
-            try:
-                action_type = audit_log_models.AuditLogEventType(entry_payload["action_type"])
-            except ValueError:
-                action_type = entry_payload["action_type"]
+            action_type = audit_log_models.AuditLogEventType(entry_payload["action_type"])
```
FasterSpeeding, line 424: For the relevant audit log fields we were likely typing them as `Union[Enum, int]` on the model, I don't see that being removed anywhere and unless I missed it so would the Union with int now be unnecessary?

hikari/impl/entity_factory.py
```diff
@@ -1084,10 +1078,7 @@ def deserialize_guild_member_ban(self, payload: data_binding.JSONObject) -> guil
     def _set_partial_guild_attributes(payload: data_binding.JSONObject) -> _PartialGuildFields:
         features = []
         for feature in payload["features"]:
-            try:
-                features.append(guild_models.GuildFeature(feature))
-            except ValueError:
-                features.append(feature)
+            features.append(guild_models.GuildFeature(feature))
```
FasterSpeeding, line 1081: For guild.features we were likely typing them as `Union[Enum, str]` on the model, I don't see that being removed anywhere and unless I missed it so would the Union with str now be unnecessary?

hikari/utilities/enums.py
```diff
@@ -0,0 +1,236 @@
...
+
+class _EnumNamespace(typing.Dict[str, typing.Any]):
+    __slots__: typing.Sequence[str] = ("base", "names_to_values", "values_to_names")
+
+    def __init__(self, base: typing.Type[typing.Any]) -> None:
+        super().__init__()
+        self.base = base
+        self.names_to_values: typing.Dict[str, typing.Any] = {}
+        self.values_to_names: typing.Dict[str, typing.Any] = {}
+        self["__doc__"] = "An enumeration."
```
FasterSpeeding, line 43: I'm not exactly sure how this works in the inheritance tree but just making sure that this doesn't override any set enums on the classes that inherit from this right?

hikari/impl/entity_factory.py
davfsa: No. The enum implementation handles it in a way that if we cast a value that is not found in the enum, it returns the raw value:
https://github.com/nekokatt/hikari/blob/1407e6c7565dd97c78b61e5cbbac41e7741cc8fd/hikari/utilities/enums.py#L117-L122

This is done so that enums dont break if Discord decides to suddenly change something

hikari/impl/entity_factory.py
davfsa: Same as above

hikari/impl/entity_factory.py
FasterSpeeding: So we need those unions everywhere then?

hikari/utilities/enums.py
davfsa: No, it doesnt affect it at the slightest. 
image

It will only show that doc if the subclass doesn't overwrite the doc

hikari/impl/entity_factory.py
davfsa: We have to since we cant guarantee the user that the value will be in the enum if they use a older version of the lib that doesn't include something that Discord added. This just helps bots from randomly breaking due to Discord changing stuff

FasterSpeeding: But we do want our models to be typed accurately don't we? Seems weird to leave those unions in 2 arbitrary places if the behaviour is now present across all our models

davfsa: Oh, thought all the other ones already had that, will fix :+1:

FasterSpeeding:
lgtm, just gonna have to fix that merge conflict

nekokatt:
This kinda links in to what I said in #201 or one of the related issues about whether we are misusing enums for sone stuff.

Lgtm though.

nekokatt (approved):

nekokatt:
Just fix da merge conflicts and we can see.

I'd be interested to see how this affects the startup overhead with asyncio_debug=1

davfsa:
@nekokatt rebased :+1:

FasterSpeeding (approved):
Assuming you managed to add all of those unions, looks good to me

hikari/api/rest.py
```diff
@@ -338,7 +338,7 @@ async def edit_permission_overwrites(
         target : typing.Union[hikari.users.PartialUser, hikari.guilds.PartialRole, hikari.channels.PermissionOverwrite, hikari.snowflakes.Snowflakeish]
             The channel overwrite to edit. This may be the object or the ID of an
             existing overwrite.
-        target_type : hikari.undefined.UndefinedOr[typing.Union[hikari.channels.PermissionOverwriteType, builtins.str]]
```
FasterSpeeding, line 341: Should these have not been left?

davfsa: The consistency across all REST endpoint is force the args that take an enum dont also take the raw value. If you want to change this it might be worth to make another pr for it, since this one is getting kinda big.

TLDR; consistency

FasterSpeeding: Oh thought that was a model not a method (was looking on mobile)

davfsa: All good :)

PR #215, opened 2020-09-22
Comparing applications.TeamMember with guilds.Member and users.User

### LINKED ISSUE #194: Team members cannot be compared with users.
Suppose a situation where:
```py
app = await bot.rest.fetch_application()

event: GuildMessageCreateEvent = ...

assert event.author_id in app.team.members
assert event.author_id in app.team.members.keys()
```

You will find surprisingly that the following does not hold true
```py
assert event.author in app.team.members.values()
```

This is incorrect and should be fixed.

tomxey:
### Summary
This PR modifies `__hash__` and `__eq__` methods of `applications.TeamMember` and `guilds.Member` so that they are always compared according to their `user` field (other fields don't matter anymore).
Custom `__hash__` and `__eq__` had to be written, resigning from the ones generated by the `@attr` decorator(or maybe I'm wrong, I don't have experience with attrs pakage).

I'm not sure if this is the best solution as now other fields of `TeamMember` and `Member` are always ignored for every comparison, is this desired?

Another way to make the `in` operator work (see linked issue) would be to define methods like `__contains__`, `__getitem__` and so on... on the `applications.Team` class or on some new class like `applications.TeamMembers`.
These methods could then support different object types passed as arguments.

@nekokatt What do you think?

### Checklist

### Related issues

This PR fixes #194

### DIFF
--- a/hikari/applications.py
+++ b/hikari/applications.py
@@ -265,31 +265,37 @@ class TeamMembershipState(enum.IntEnum):
 
 
 @attr_extensions.with_copy
-@attr.s(eq=True, hash=True, init=True, kw_only=True, slots=True, weakref_slot=False)
+@attr.s(eq=False, hash=False, init=True, kw_only=True, slots=True, weakref_slot=False)
 class TeamMember:
     """Represents a member of a Team."""
 
-    app: traits.RESTAware = attr.ib(repr=False, eq=False, hash=False, metadata={attr_extensions.SKIP_DEEP_COPY: True})
+    app: traits.RESTAware = attr.ib(repr=False, metadata={attr_extensions.SKIP_DEEP_COPY: True})
     """The client application that models may use for procedures."""
 
-    membership_state: TeamMembershipState = attr.ib(eq=False, hash=False, repr=False)
+    membership_state: TeamMembershipState = attr.ib(repr=False)
     """The state of this user's membership."""
 
-    permissions: typing.Sequence[str] = attr.ib(eq=False, hash=False, repr=False)
+    permissions: typing.Sequence[str] = attr.ib(repr=False)
     """This member's permissions within a team.
 
     At the time of writing, this will always be a sequence of one `builtins.str`,
     which will always be `"*"`. This may change in the future, however.
     """
 
-    team_id: snowflakes.Snowflake = attr.ib(eq=True, hash=True, repr=True)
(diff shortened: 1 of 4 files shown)


nekokatt:
> Custom __hash__ and __eq__ had to be written, resigning from the ones generated by the <at>attr decorator(or maybe I'm wrong, I don't have experience with attrs pakage).

Sounds good to me. I don't believe there is any other way to do it 👍 

I will take a look, bare with me as I am on mobile

hikari/applications.py
```diff
@@ -265,32 +265,38 @@ def __str__(self) -> str:
...
+    user: users.User = attr.ib(repr=True)
     """The user representation of this team member."""
 
     def __str__(self) -> str:
         return str(self.user)
 
+    def __hash__(self) -> int:
+        return hash(self.user)
+
+    def __eq__(self, other: object) -> bool:
```
nekokatt, line 297: Instead of `object`, we should use `typing.Any` here.

Also iirc attrs overrides eq and hashcode per model anyway, so the way you have implemented this is exactly what we need!

tests/hikari/integration/test_users_comparison.py
```diff
@@ -0,0 +1,113 @@
...
+
+import mock
+
+from hikari import applications
+from hikari import guilds
+from hikari import snowflakes
+from hikari import users
+
+
+def assert_objects_equal(a, b):
```
nekokatt, line 31: I've wanted to start splitting into unit tests and integration tests, so thanks for setting this structure up!

nekokatt:
Apart from the type hint that should ideally be `typing.Any`, all looks great.

Thanks for taking the time to put this together! Drop me a comment when ready and I'll approve it when I get home!

davfsa (commented):
LGTM :+1:

hikari/applications.py
tomxey: It seems like `Any` wants to propagate itself further into the return type.
`error: Returning Any from function declared to return "bool"  [no-any-return]
            return self.user == other`
Maybe `object` type is the correct choice? After all I'm treating it like an `object`, I'm not executing any methods on it that are not part of `object` interface.

nekokatt: We can leave it as-is for now, this may be a MyPy bug, so I'll have a look around when I find time

Issue #197, opened 2020-09-16
Cache methods should take Snowflakeish instead of Snowflake


Tmpod:
### Summary

Cache methods, both for retrieval and updating (in mutable cache), should have their snowflake arguments typed as `Snowflakeish` instead of just `Snowflake`.

### Problem

The API is currently inconsistent, since, for example, the REST components  are typed with `Snowflakeish`, while the cache is typed with `Snowflake`. It's also more convenient to be able to use whatever we have at hand from these three quite interchangeable formats, `int`, `str` and `Snowflake`. Additionally, it would also address the fact that currently the cache methods can take ints all fine, despite not being typed as such.

### Ideal implementation

Of course the first step would be to change the methods' signatures on the base classes, and on the implementations. Then I suppose one would have to change the implementations to cast the given Snowflakeish to just one of the three types, possibly Snowflake being the best candidate, since there would be no need to change the cache's collections' typing and also because this type offers more utility methods which are, well, useful :P  One downside would be memory consumption, which would naturally be higher in Snowflake than in the builtins.

### Checklist

- [x] I have searched the issue tracker and have made sure it's not a duplicate.

Tmpod:
One problem that Espy just thought about in a little chat on DCA was that allowing `Snowflakeish` would mean allowing `str`, meaning tons of `int` casts everywhere which would introduce a fair amount of overhead. This union type works great in REST because sooner or later the ID will have to be cast into a string in order to be sent in the JSON payload, but this doesn't quite fit this cache issue.   
Not sure what the best course of action would be.

nekokatt:
the eventual plan is to have model methods/properties to access this info in a more controlled way.

While I appreciate that this is not ideal for now, it should probably be best to manually cast to Snowflake before touching the cache, if you need MyPy compat. I will discuss with @FasterSpeeding the implications of changing the cache interface to allow int as well. I assume there isn't much that needs changing other than the typehints, since Snowflake inherits from int anyway, so the `__eq__` and friends, and `__hash__` overloads are interoperable.

I _do_ eventually have plans to deprecate the use of str in Snowflakeish, since unless you are talking directly to the Discord API on a lower level than that which we provide, then you shouldn't need to use str anyway. I am not aware of any part of hikari that requires this anymore. str is a relic from when the core REST client provided almost zero transformation to/from the API JSON response/request shape. I will look to make an issue for this.

[...]

PR #222, opened 2020-09-24
Snowflakeish types are no longer compatible with 'str' directly.

nekokatt:
'str' objects must now be cast to a Snowflake or int to be used with compatible functions.

Fixes some points in #197.


tandemdude (approved):
lgtm

PR #225, opened 2020-09-24
Make sure the id of @everyone is always in member.role_ids

FasterSpeeding:
### Summary
Make sure the id of @everyone is always in member.role_ids

### Checklist

### DIFF
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -987,6 +987,10 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
 
         role_ids = [snowflakes.Snowflake(role_id) for role_id in payload["roles"]]
 
+        # If Discord ever does start including this here without warning we don't want to duplicate the entry.
+        if guild_id not in role_ids:
+            role_ids.insert(0, guild_id)
+
         joined_at = date.iso8601_datetime_string_to_datetime(payload["joined_at"])
 
         raw_premium_since = payload.get("premium_since")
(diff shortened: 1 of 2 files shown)


hikari/impl/entity_factory.py
```diff
@@ -987,6 +987,9 @@ def deserialize_member(
 
         role_ids = [snowflakes.Snowflake(role_id) for role_id in payload["roles"]]
 
+        if guild_id not in role_ids:
```
davfsa, line 990: Is it possible for them to send it?

FasterSpeeding: they could decide to start including it someday without warning, that's just protection against weird behaviour

davfsa: :+1:

tests/hikari/impl/test_entity_factory.py
```diff
@@ -1430,7 +1430,7 @@ def test_deserialize_member(self, entity_factory_impl, mock_app, member_payload,
         assert member.guild_id == 76543325
         assert member.user == entity_factory_impl.deserialize_user(user_payload)
         assert member.nickname == "foobarbaz"
-        assert member.role_ids == [11111, 22222, 33333, 44444]
+        assert member.role_ids == [76543325, 11111, 22222, 33333, 44444]
```
davfsa, line 1433: Maybe a test for when its there already?

hikari/impl/entity_factory.py
nekokatt: why aren't we using an ID table here?

nekokatt (approved):

hikari/impl/entity_factory.py
tomxey: Is order here meaningful? Could `set` be used here instead of `list`? No duplicate checks would be necessary then.

nekokatt: yeah, I agree with @tomxey on this. If we used IDTable here, that implements the set interface, but is implemented as a snowflake array internally as a bisect binary tree.

nekokatt (dismissed):
We should use IDTable internally.

PR #227, opened 2020-09-26
Refactoring message event hierarchy; other tidyups.

### LINKED ISSUE #226: Refactor Message Events
We need to refactor the inheritance model for message events.

It would probably be a good idea to remove the message object from deletion events entirely, as this is confusing.

nekokatt:
Until this gets merged, master will not build as it fixes issues there as well.

Ideally it would be nice to get this through ASAP.

I will outline the other tidyups before the main change, as they are only
brief.

- PRIVATE\* intents are now named DM* intents to match terminology across
  the rest of the library.
- hikari.utilities.mappings is renamed to hikari.utilities.collections.
- IDTable from hikari.utilities.cache is moved to hikari.utilities.collections,
  has been refactored and fully tested, and altered to be slightly more
  efficient.
- IDTable is now named SnowflakeSet.

Cache mappings have been renamed to be a little clearer as to what
they really are doing.

- DMChannelMRUMutableMapping is now named DMChannelCacheMapping.
- MappedCollection is now named ExtendedMutableMapping.
- DictionaryCollection is now named FreezableDict.
- MRIMutableMapping is now named TimedCacheMap.
- CMRIMutableMapping is now named LimitedCapacityCacheMap.

Message hierarchy changes
-------------------------

Due to the fundamental differences between create/update/delete events and
information they contain for messages, I have concluded that the complexity
created by having these staggered classes is not overly helpful in writing
robust error-safe code. It is definitely not worth the extra overhead, as
there is almost no cases where this is overly useful where you could use
this in such a way that using two event listeners would not be more
terse and idiomatic.

Therefore, GuildMessageEvent and DMMessageEvent no longer exist. The new
hierarchy is as follows instead:

```
  Event
   |
   |-- ShardEvent
   '    |
   '    |-- MessageEvent
        '    |
        '    |-- MessageCreateEvent
             |    |-- GuildMessageCreateEvent
             |    '-- DMMessageCreateEvent
             |
             |-- MessageUpdateEvent
             |    |-- GuildMessageUpdateEvent
             |    '-- DMMessageUpdateEvent
             |
             '-- MessageDeleteEvent
                  |-- GuildMessageDeleteEvent
                  '-- DMMessageDeleteEvent
```

** Bulk deletion events **

Bulk deletion events have been merged with deletion events.
The reason for doing this is because Discord already provides
little to no useful information for singular message deletions
other than the message id, the guild id, and the channel id.
This means that no stateless information about what was in the
message nor who made it or who even deleted it originally
can be discovered without polling message history before the
event can occur (which is basically impossible), or by
maintaining a long-lived message cache (which is not yet
supported fully on this API). This would also need heavy audit
log polling to be somewhat effective.

This is pretty much a limitation of how Discord operates,
unfortunately.

This means the only reasonable use case for this event currently
is for tracking any kind of message deletion. The bulk delete
event shares practically the same shape as the singular deletion
event, and anyone listening on deletion events for handling of
things like pagination will therefore almost always want to
listen to both events.

The interface for this event model is now to provide three
attributes to handle detecting which messages were deleted:

  - `message_ids`: will be a `SnowflakeSet` of each ID that
    was deleted. In singular message deletion cases, this
    will always have a single item.
  - `message_id`: to match with the `MessageEvent` interface,
    and to provide a shorthand for cases where people really
    do want to handle only a singular message deletion, rather
    than an operation where you would just iterate across
    the SnowflakeSet. For bulk deletions this will return the
    first deleted message (implementation detail makes this
    the smallest magnitude snowflake that is sent).
  - `is_bulk`: true if a bulk deletion, false if not.

Additionally, since it is undocumented, bulk deletions in DMs
will now result in a NotImplementedError should they occur
(which unless Discord break their API contract, should not
happen).

It is worth noting message deletion events no longer have a
`message` attribute either, as this was pretty much a complete
waste of space given nothing could be in it in the first place
other than the id, channel_id and guild_id.

** Message Update Events **

Now that we have ditched the previous structure for messages,
we can make several attributes such as `author` and `guild_id`
simpler on the PartialMessage model that only the
MessageUpdateEvent uses. You now only have to `None`-check
the `guild_id`, and the `author` should always be present.

** Message Create Events **

This interface is mostly the same, other than having some
property definitions refactored.

### Checklist

### Related issues

Close #226


davfsa (dismissed):
Looks really good, like the new hierarchy :+1:

davfsa:
FYI, edited the pr to close #226 when merged :)

hikari/impl/entity_factory.py
```diff
@@ -1406,19 +1406,19 @@ def deserialize_invite_with_metadata(self, payload: data_binding.JSONObject) ->
...
+        member: undefined.UndefinedOr[guild_models.Member] = undefined.UNDEFINED
         if "guild_id" in payload:
             guild_id = snowflakes.Snowflake(payload["guild_id"])
 
-        member: undefined.UndefinedOr[guild_models.Member] = undefined.UNDEFINED
-        if "member" in payload:
-            assert author is not undefined.UNDEFINED
-            assert guild_id is not undefined.UNDEFINED
-            member = self.deserialize_member(payload["member"], user=author, guild_id=guild_id)
+            # TODO: will discord always send this with a guild? If so, we should default this to none I guess.
```
FasterSpeeding, line 1419: They will not for message updates that are triggered by embed workers

hikari/impl/entity_factory.py
```diff
@@ -1406,19 +1406,19 @@ def deserialize_invite_with_metadata(self, payload: data_binding.JSONObject) ->
     ##################
 
     def deserialize_partial_message(self, payload: data_binding.JSONObject) -> message_models.PartialMessage:
-        author: undefined.UndefinedOr[user_models.User] = undefined.UNDEFINED
-        if "author" in payload:
-            author = self.deserialize_user(payload["author"])
+        assert (
+            "author" in payload
```
FasterSpeeding, line 1410: This isn't included in message updates triggered by embed workers. 
See the following example
```json
{
    "id": "759533107606585354",
    "embeds": [
        {
            "video": {
                "width": 1280,
                "url": "https://www.youtube.com/embed/Eak9JO1zO2U",
                "height": 720
            },
            "url": "https://www.youtube.com/watch?v=Eak9JO1zO2U",
            "type": "video",
            "title": "Conduite \u00e0 tenir la PLS : Position Lat\u00e9rale de S\u00e9curit\u00e9 - Protectio...",
            "thumbnail": {
                "width": 1280,
                "url": "https://i.ytimg.com/vi/Eak9JO1zO2U/maxresdefault.jpg",
                "proxy_url": "https://images-ext-1.discordapp.net/external/IMPARnTHG9KAVzy1MrY3P6ZxrGkAXuIDNTVN3lFHfLI/https/i.ytimg.com/vi/Eak9JO1zO2U/maxresdefault.jpg",
                "height": 720
            },
            "provider": {
                "url": "https://www.youtube.com",
                "name": "YouTube"
            },
            "description": "Position Lat\u00e9rale de S\u00e9curit\u00e9 : PLS. \nD\u00e9monstration comment\u00e9e en temps r\u00e9el par la Protection Civile de Paris\n\nD\u00e9couvrez sur notre site Internet toutes nos formations www.protectioncivile.org\n\n\nPlus de 480 b\u00e9n\u00e9voles au sein de la Protection Civile de Paris ; pourquoi pas vous ...",
            "color": 16711680,
            "author": {
                "url": "https://www.youtube.com/channel/UCtM2N8EnpnlX4vnONCkSmWQ",
                "name": "Protection Civile Paris Seine"
            }
        }
    ],
    "channel_id": "574921092926537729",
    "guild_id": "574921006817476608"
}
```

nekokatt: really

ffs.

nekokatt: Fixed.

hikari/impl/entity_factory.py
nekokatt: Fixed.

hikari/events/message_events.py
```diff
@@ -274,383 +188,431 @@ def is_webhook(self) -> bool:
...
+    # <<inherited docstring from MessageCreateEvent>>
+
+    shard: shard_.GatewayShard = attr.ib(metadata={attr_extensions.SKIP_DEEP_COPY: True})
+    # <<inherited docstring from ShardEvent>>
+
     @property
-    def channel_id(self) -> snowflakes.Snowflake:
-        # <<inherited docstring from MessageEvent>>.
-        return self.message.channel_id
+    def author(self) -> users.User:
```
FasterSpeeding, line 232: if this was created in a guild then shouldn't this always be a member object?

hikari/events/message_events.py
```diff
@@ -274,383 +188,431 @@ def is_webhook(self) -> bool:
...
+        """Member or user that sent the message.
+
+        Returns
+        -------
+        typing.Union[hikari.users.User, hikari.guilds.Member]
+            The user that sent the message. If the member is cached
```
FasterSpeeding, line 238: wouldn't that use message_payload["member"] not the cache?

nekokatt: could not confirm. Although that seems like a sensible assumption.

Can you pop a suggested fix on this? Im not on a pc rn

hikari/events/message_events.py
```diff
@@ -274,383 +188,431 @@ def is_webhook(self) -> bool:
...
+        """
+        return self.message.id
 
 
-@base_events.requires_intents(intents.Intents.GUILD_MESSAGES)
-@attr_extensions.with_copy
 @attr.s(kw_only=True, slots=True, weakref_slot=False)
-class GuildMessageUpdateEvent(GuildMessageEvent, MessageUpdateEvent):
-    """Event triggered when a message is updated in a guild channel."""
+@base_events.requires_intents(intents.Intents.GUILD_MESSAGES)
```
FasterSpeeding, line 474: what happened to with_copy here?

nekokatt: I'll readd that.

davfsa (approved):

PR #229, opened 2020-09-26
Improved handling of bucket ratelimits in edge cases.

nekokatt:
Running two bots on the same token should now not error if you get ratelimited due to tracking issues.

The example case we found was running two instances of the same bot that responded to pings with a specific reaction under the same token. 

-- This should now simply log a message to say that this has happened.


FasterSpeeding (dismissed):
lgtm

davfsa (approved):

FasterSpeeding (approved):
lgtm

PR #230, opened 2020-09-26
Added support for color detection via COLORTERM env var.

nekokatt:
Tmpod said he would beat me if I didn't add this, and that man has a very hard hand, from experience.


davfsa (approved):

FasterSpeeding (approved):
lgtm

PR #234, opened 2020-09-27
Remove DM channels from the cache and dm channel events that were removed from v8

FasterSpeeding:
### Summary
Remove DM channels from the cache and dm channel events that were removed from v8

### Checklist


nekokatt:
How do we DM users now then and respond to DMs?
make two API calls per message we want to send to ensure the channel is open?

FasterSpeeding:

davfsa:

DM'ing users will have to be creating the channel first to make sure that it exists and then sending a message there. Responding should be as easy as `Message.reply()` since the channel_id is correct.

PR #269, opened 2020-10-04
Make flake8 and mypy run on examples

davfsa:
### Checklist

### DIFF
--- a/hikari/files.py
+++ b/hikari/files.py
@@ -127,7 +127,7 @@ This may be one of:
 - `aiohttp.StreamReader`
 """
 
-Resourceish = typing.Union["Resource[AsyncReader]", Pathish, Rawish]
+Resourceish = typing.Union["Resource", Pathish, Rawish]
 """Type hint representing a file or path to a file/URL/data URI.
 
 This may be one of:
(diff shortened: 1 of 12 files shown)


nekokatt:
We should do mypy also.

davfsa:
@nekokatt done :+1:

hikari/files.py
```diff
@@ -127,7 +127,7 @@
 - `aiohttp.StreamReader`
 """
 
-Resourceish = typing.Union["Resource[AsyncReader]", Pathish, Rawish]
+Resourceish = typing.Union["Resource", Pathish, Rawish]
```
FasterSpeeding, line 130: what are the implications of letting the Resource be treated as `Resource[typing.Any]` here as opposed to `Resource[AsyncReader]`

FasterSpeeding (approved):
looks fine to me

hikari/files.py
davfsa: I changed this because `URL` is a `Resource[WebReader]` and passing it as an attachment made mypy complain. I just removed the `AsyncReader` part.

Further more, `Resource` is defined like this:
```py
class Resource(typing.Generic[ReaderImplT], abc.ABC):
    ...
```
and `ReaderImplT` being: 
```py
ReaderImplT = typing.TypeVar("ReaderImplT", bound="AsyncReader")
```
which is already bound to `AsyncReader`, so I think it should be fine. Then again, if you think it should be changed to something else, then I will change it :smile:

PR #272, opened 2020-10-05
Added batch files to execute examples from cmd

davfsa:
I have tested this on a Windows 10 virtual machine, but would highly appreciate if you could have another look @FasterSpeeding. 

Should be as easy as opening cmd, going to the examples dir and executing `{script}.bat TOKEN` (just like the bash ones)

### Checklist

### DIFF
--- a/examples/image_resources/image_resources.bat
+++ b/examples/image_resources/image_resources.bat
@@ -1,0 +1,50 @@
+:: Copyright (c) 2020 Nekokatt
+::
+:: Permission is hereby granted, free of charge, to any person obtaining a copy
+:: of this software and associated documentation files (the "Software"), to deal
+:: in the Software without restriction, including without limitation the rights
+:: to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+:: copies of the Software, and to permit persons to whom the Software is
+:: furnished to do so, subject to the following conditions:
+::
+:: The above copyright notice and this permission notice shall be included in all
+:: copies or substantial portions of the Software.
+::
+:: THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+:: IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+:: FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+:: AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


nekokatt:
Would a powershell script be more sensible?

davfsa:
> Would a powershell script be more sensible?

I thought about it, but having a further look into it, powershell scripts are can be disabled sometimes on Windows, and might require you to change the policy to enable it ([execution policies docs](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-7)).

The reason why I went with the batch file is because you can also execute them from powershell (not only cmd).

nekokatt:
Fair. Get someone to verify on Windows once and it looks good to merge.

GitHub mobile is being dumb and not allowing me to approve

examples/image_resources/image_resources.bat
```diff
@@ -0,0 +1,51 @@
...
+:: OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+:: SOFTWARE.
+
+:: Script to set the PYTHONPATH variable up to point to the root of this repository.
+
+@echo off
+
+if "%~1"=="" (
+    echo Please pass a bot token as the first parameter to this script!
+    :: This is used so that if the script is executed by double clicking it,
```
FasterSpeeding, line 27: These in-statement comments are being interpreted as a drive letter on my system in both of these, leading to it sending out a unnecessary warning before exiting with status 1
```
>image_resources.bat
Please pass a bot token as the first parameter to this script!
The system cannot find the drive specified.
Press any key to continue . . .
```

nekokatt: comments in batch are `REM blah blah`

davfsa: Hmm interesting.... This is weird because there are a lot of comments before this and those didnt spam anything. Further more, that is the way to mark as comments in batch (as I learned today).

Could you try removing those lines (that one and the following one) and see if it still errors?

FasterSpeeding: replacing "::" with "REM" on those two lines in both of the bat files got rid of the warning.

davfsa: > comments in batch are REM blah blah

Most of resources I found said that both `rem` and `:: ` were fine and I went with `:: ` as it made it more readable

davfsa: > replacing "::" with "REM" on those two lines in both of the bat files got rid of the warning.

Ig I will change all the comments then

nekokatt: might be the indent, im not sure. Batch is archaic.

davfsa: Well, should be fixed :+1: @FasterSpeeding mind trying again?

FasterSpeeding (approved):
lgtm

PR #275, opened 2020-10-06
Implemented default max-retry of 60s.

### LINKED ISSUE #264: Add a max retry value for hikari's rest ratelimit handling
### Summary
Add a max_retry value for Hikari's rest ratelimit handling. This Could in-theory replace the check for whether a ratelimit was triggered by the ratelimit system that's exposed over headers or a separate system which can be found [here](https://github.com/nekokatt/hikari/blob/5429e110e64f1a737dce939111a208b962cfd44d/hikari/impl/rest.py#L717)

### Problem
The guild role create endpoint ratelimit has a reset-after of 2 days, this could lead to Hikari implicitly backing off for said amount of time if they were to hit the limit which would likely lead to unwanted and confusing results.
image

### Ideal implementation
Have a check for retry after time is greater than the limit which was passed through while initiating the bot/rest client and raise a ratelimited error if it is (this will likely be accepted as a int, float, timedelta or None where None represents no limit for ppl who are feeling more frisky).

### Checklist

nekokatt:
This means if any REST rate limit occurs that lasts more than 60 seconds,
you will instead get a hikari.errors.RateLimitTooLongError get raised
instead which contains details of how long to back off for should you
wish to retry later.

You can disable this functionality by passing `max_rate_limit=float("inf")`
to the `hikari.impl.bot.BotApp` and `hikari.impl.rest.RESTApp` constructors
if you have a reason to not want this. The assumption is that in 99% of
cases, waiting for sustained periods of time would be confusing to the
user and lead to a poorer user experience overall, which is why this has
been implemented.

I still need to fix the tests.

Fixes #264

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -121,9 +121,15 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If you are missing the `READ_MESSAGES` permission in the channel.
         hikari.errors.NotFoundError
             If the channel is not found.
+        hikari.errors.RateLimitTooLongError
+            Raised in the event that a rate limit occurs that is
+            longer than `max_rate_limit` when making a request.
+        hikari.errors.RateLimitTooLongError
+            Raised in the event that a rate limit occurs that is
+            longer than `max_rate_limit` when making a request.
         hikari.errors.RateLimitedError
             Usually, Hikari will handle and retry on hitting
-            rate-limits automatically. This includes bucket-specific
+            rate-limits automatically. This includes most bucket-specific
             rate-limits and global rate-limits. In some rare edge cases,
             however, Discord implements other undocumented rules for
             rate-limiting, such as limits per attribute. These cannot be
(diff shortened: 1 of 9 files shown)


hikari/api/rest.py
```diff
@@ -121,9 +121,15 @@ async def fetch_channel(
...
+        hikari.errors.RateLimitTooLongError
+            Raised in the event that a rate limit occurs that is
+            longer than `max_rate_limit` when making a request.
+        hikari.errors.RateLimitTooLongError
+            Raised in the event that a rate limit occurs that is
+            longer than `max_rate_limit` when making a request.
         hikari.errors.RateLimitedError
             Usually, Hikari will handle and retry on hitting
-            rate-limits automatically. This includes bucket-specific
+            rate-limits automatically. This includes most bucket-specific
```
davfsa, line 132: Shouldnt this change also be made on all the other endpoints documentation for `hikari.errors.RateLimitedError`?

davfsa (commented):
Looks really nice, just one little thing :+1:

davfsa:
@nekokatt all pipelines should pass now :+1:

nekokatt:
Assuming no other feedback as this has been open long enough now, so merging.

PR #278, opened 2020-10-06
And more testing

davfsa:
### Checklist

### Related issues
#47

### DIFF
--- a/hikari/internal/aio.py
+++ b/hikari/internal/aio.py
@@ -176,8 +176,9 @@ async def all_of(
                     await f
                 except asyncio.CancelledError:
                     pass
+
+        gatherer.cancel()
         try:
-            gatherer.cancel()
-            await gatherer
+            await gatherer  # pragma: no cover - I tried everything to make this work with coverage, but no luck :(
         except asyncio.CancelledError:
             pass
--- a/hikari/internal/aio.py
+++ b/hikari/internal/aio.py
@@ -179,6 +179,8 @@ async def all_of(
 
         gatherer.cancel()
         try:
-            await gatherer  # pragma: no cover - I tried everything to make this work with coverage, but no luck :(
+            # coverage.py will complain that this is not fully covered, as the
+            # "except" block will always be hit. This is intentional.
+            await gatherer  # pragma: no cover
         except asyncio.CancelledError:
             pass
--- a/hikari/internal/time.py
+++ b/hikari/internal/time.py
@@ -110,20 +110,20 @@ def iso8601_datetime_string_to_datetime(datetime_str: str) -> datetime.datetime:
     return datetime.datetime.fromisoformat(datetime_str)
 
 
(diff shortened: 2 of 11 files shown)


hikari/internal/aio.py
```diff
@@ -176,8 +176,9 @@ async def all_of(
                     await f
                 except asyncio.CancelledError:
                     pass
+
+        gatherer.cancel()
         try:
-            gatherer.cancel()
-            await gatherer
+            await gatherer  # pragma: no cover - I tried everything to make this work with coverage, but no luck :(
```
nekokatt, line 182: odd, are you able to hit this on a breakpoint?

If so, theres a 0ytest coverage flag to make it more accurate, albeit slower, maybe try that

nekokatt (approved):

FasterSpeeding (dismissed):

hikari/internal/aio.py
davfsa: It does hit, the problem is that it only marks it partially covered.

davfsa: image
This is the problem I don't know how to solve

nekokatt: Oh, I see the issue.

This will always fail to complete, thus it is saying there is no coverage for the success case, because there cannot be a success case.

One solution would be to use `with contextlib.suppress(asyncio.CancelledError)`, but that is significantly slower than what we have here, and since we may use this function in performance critical places, we should probably avoid that.

nekokatt, line 182:
```suggestion
            # coverage.py will complain that this is not fully covered, as the
            # "except" block will always be hit. This is intentional.
            await gatherer  # pragma: no cover
```

hikari/errors.py
```diff
@@ -317,9 +317,6 @@ class RESTErrorCode(int, enums.Enum):
     MESSAGE_TOO_OLD = 50_034
     """A message provided was too old to bulk delete."""
 
-    TWO_FACTOR_AUTHENTICATION_REQUIRED = 60_003
-    """2FA is required to use this endpoint."""
```
nekokatt, line 321: Why has this been removed?

https://github.com/discord/discord-api-docs/pull/1671 documents this, Discord just haven't bothered to merge it.

Please can we revert this.

nekokatt (changes requested):
See the two comments I added.

hikari/errors.py
davfsa: I thought that bots could not get this error, which is why I removed it. Will revert

davfsa:
@nekokatt all solved

hikari/errors.py
nekokatt: Sure, but the REST only client allows logging in via users over OAuth2 :)

Also I think bots may be able to get this in some edge cases possibly.

nekokatt (approved):

PR #280, opened 2020-10-07
Moved default Travis CI build stages to use CPython 3.9.0.

nekokatt:
CPython 3.9.0 was released on Monday (two days ago, 5th October 2020).


nekokatt:
Blocked until travis push a cpython3.9 bundle for linux amd64 and arm64

PR #281, opened 2020-10-07
Changed max version in setup.py and added Python 3.10 TROVE classifier.

nekokatt:
This is blocked until
`Programming Language :: Python :: 3.10`
is deployed to https://pypi.org/classifiers/ and https://test.pypi.org/classifiers/. Until this happens, deployments will be rejected.


nekokatt:
See https://github.com/pypa/trove-classifiers/pull/51 for my request to add this classifier.

This is also partially blocked by #282, since until that build can pass, we cannot verify that we have working support.

PR #293, opened 2020-10-10
Implement user DM flow

### LINKED ISSUE #237: Implement user DM flow for sending new messages.

_Originally posted by @FasterSpeeding in https://github.com/nekokatt/hikari/pull/234#issuecomment-699677059_

davfsa:
### Checklist

### Related issues
Fixes #237

### DIFF
--- a/hikari/users.py
+++ b/hikari/users.py
@@ -39,7 +39,12 @@ from hikari.internal import enums
 from hikari.internal import routes
 
 if typing.TYPE_CHECKING:
+    from hikari import channels
+    from hikari import embeds
+    from hikari import guilds
+    from hikari import messages
     from hikari import traits
+    from hikari import users
 
 
 @typing.final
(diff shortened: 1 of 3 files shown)


hikari/users.py
```diff
@@ -120,6 +125,8 @@ class PartialUser(snowflakes.Unique, abc.ABC):
     API call.
     """
 
+    _dm_channel: typing.Optional[channels.DMChannel] = None
```
FasterSpeeding, line 128: I'm not sure how I feel about this solution, seems to be inconsistent for the interface style approach we're using with the PartialUser here as we're now introducing a variable to a class which is supposedly to solely be an interface. Could we not have send be an abstractmethod where how it's implemented comes down to the implementation's detail? Also would this actually work or would ._dm_channel end up being treated as a read only class attribute on an instance of this object due to lsotting?
(btw this is what i was talking about earlier, was just not in the right place to write this up all afternoon)

davfsa: @FasterSpeeding rewrote it :+1:

davfsa:
Closing this for now while I think about a better way to implement this

PR #312, opened 2020-10-13
Fix max_rate_limit

davfsa:
- Improved logic and RateLimitTooLongError error message

### Checklist

### DIFF
--- a/hikari/errors.py
+++ b/hikari/errors.py
@@ -478,6 +478,16 @@ class RateLimitTooLongError(HTTPError):
     period: float = attr.ib()
     """How long the rate limit window lasts for from start to end."""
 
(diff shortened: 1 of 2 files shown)


hikari/errors.py
```diff
@@ -478,6 +478,16 @@ class RateLimitTooLongError(HTTPError):
     period: float = attr.ib()
     """How long the rate limit window lasts for from start to end."""
 
+    message: str = attr.ib(init=False)
+    """The error message."""
+
+    @message.default
+    def _(self) -> str:
+        return (
```
FasterSpeeding, line 486: if we're defaulting this to a specific case then should we also specify the message for different case where this is raised such as out of bucket ratelimits to avoid confusion?

PR #332, opened 2020-10-29
Enum tests

davfsa:
- Remove unreachable and duplicate code
   - Fixed some logic

### Checklist

### Related issues
#47


hikari/internal/enums.py
```diff
@@ -122,8 +109,15 @@ def __call__(cls, value: typing.Any) -> typing.Any:
...
         return cls._name_to_member_map_[name]
 
+    def __contains__(cls, item: typing.Any) -> bool:
+        try:
+            cls._value_to_member_map_[item]
+            return True
+        except KeyError:
+            return False
+
     def __iter__(cls) -> typing.Iterator[str]:
```
FasterSpeeding, line 119: would iterating values here not make it an iterator of Any not str

hikari/internal/enums.py
```diff
@@ -122,8 +109,15 @@ def __call__(cls, value: typing.Any) -> typing.Any:
     def __getitem__(cls, name: str) -> typing.Any:
         return cls._name_to_member_map_[name]
 
+    def __contains__(cls, item: typing.Any) -> bool:
+        try:
+            cls._value_to_member_map_[item]
```
FasterSpeeding, line 114: how does this compare to just returning `item in cls._value_to_member_map_`?

davfsa: Same thing, just one call less, but will change if it makes it cleaner

hikari/internal/enums.py
davfsa: Yep, will fix typehint

davfsa:
@FasterSpeeding all fixed

FasterSpeeding (approved):

PR #335, opened 2020-10-29
Dont suppress errors raised by run_once

### LINKED ISSUE #330: Don't reconnect on repeatable gateway error codes (caused by the identify's configuration)
### Summary

After receiving the intent Gateway Close Event Codes 4013 (Invalid intent(s)) and 4014 (Disallowed intent(s)) trying to reconnect will lead to the bot entering a loop of trying to reconnect then receiving the same close codes which should not be the case.

This likely also applies to 4012 (Invalid API version) and the two sharding related codes 4010 (Invalid shard) and 4011 (Sharding required) which will all likely only be thrown when devs enter their values for sharding and the gateway version which are cases which it will probably be worth still covering.

### Fix

Shards should throw in these scenarios rather than retrying.

### Checklist

davfsa:
### Summary
Errors raised by `run_once` were being suppressed causing server gateway closure errors from not propagating 

### Checklist

### Related issues
Fixes #330

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -976,7 +976,7 @@ class BotApp(traits.BotAware, event_dispatcher.EventDispatcher):
             _LOGGER.debug("Shard %s started successfully in %.1fms", shard_id, (end - start) * 1_000)
             return new_shard
 
-        raise errors.GatewayError(f"Shard {shard_id} shut down immediately when starting")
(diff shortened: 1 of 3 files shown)


FasterSpeeding (approved):

davfsa:
Force merging this one since its only failing due to coverage

PR #339, opened 2020-11-03
Add `old_x` attribute to update events

### LINKED ISSUE #198: Add an `old_x` attribute to update events
### Summary

There should be an attribute in update related events that point to the latest "old" version of said entity in cache, None otherwise.

### Problem

Currently it's quite impracticable to know what changed with `XUpdateEvent` events, since the cache is updated before the event is dispatched to the listeners/subscribers. To achieve the desired effect, one would have to change the whole cache implementation so that the cache is updated after the events are dispatched, which is quite cumbersome.

Having the "possible" older version of the updated entity is quite useful because you're able to know what part of it changed and so correctly trigger whatever logic you want. Without this, it becomes much harder to execute some routine whenever, let's say, a member is muted, or a guild has its logo changed.

### Ideal implementation

I believe it would probably be best to simply add a field/attribute to those event classes that contain the cached entity before it is updated, possibly named `old_x` (`old_channel`, `old_role`, etc).   

One issue I see with this is an extra hit on cache, which can possibly be more precious than this, but I also believe if someone is needing that level of optimisation, they probably know their way around quite well and have written their own implementation of stuff.

### Checklist

Reliku:
### Summary

Messages are now cached, needed for MessageUpdateEvent.
Init StatefulCacheImpl requests a new argument max_messages - maximum count of messages in the cache.

Added `old_x` attribute in next `XUpdateEvent`:
    GuildChannelUpdateEvent, GuildUpdateEvent, EmojisUpdateEvent, PresenceUpdateEvent, MemberUpdateEvent, GuildMessageUpdateEvent, DMMessageUpdateEvent, RoleUpdateEvent, OwnUserUpdateEvent, VoiceStateUpdateEvent.

### Checklist

### Related issues
Closes #198

### DIFF
--- a/hikari/api/cache.py
+++ b/hikari/api/cache.py
@@ -1587,12 +1587,12 @@ class MutableCache(Cache, abc.ABC):
         """
 
     @abc.abstractmethod
-    def clear_messages(self) -> CacheView[snowflakes.Snowflake, messages.PartialMessage]:
+    def clear_messages(self) -> CacheView[snowflakes.Snowflake, messages.Message]:
         """Remove all message objects from the cache.
 
         Returns
         -------
-        CacheView[hikari.snowflakes.Snowflake, hikari.messages.PartialMessage]
+        CacheView[hikari.snowflakes.Snowflake, hikari.messages.Message]
             A view of message objects that were removed from the cache.
 
         Raises
@@ -1602,7 +1602,7 @@ class MutableCache(Cache, abc.ABC):
         """
 
     @abc.abstractmethod
-    def delete_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
+    def delete_message(self, message_id: snowflakes.Snowflake, /) -> typing.Optional[messages.Message]:
         """Remove a message object from the cache.
 
         Parameters
@@ -1612,7 +1612,7 @@ class MutableCache(Cache, abc.ABC):
 
         Returns
         -------
-        typing.Optional[hikari.messages.PartialMessage]
+        typing.Optional[hikari.messages.Message]
             The object of the message that was removed from the cache if found,
             else `builtins.None`.
 
@@ -1623,7 +1623,7 @@ class MutableCache(Cache, abc.ABC):
         """
 
     @abc.abstractmethod
-    def get_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
+    def get_message(self, message_id: snowflakes.Snowflake, /) -> typing.Optional[messages.Message]:
         """Get a message object from the cache.
 
         Parameters
@@ -1633,27 +1633,27 @@ class MutableCache(Cache, abc.ABC):
 
         Returns
         -------
-        typing.Optional[hikari.messages.PartialMessage]
+        typing.Optional[hikari.messages.Message]
             The object of the message found in the cache or `builtins.None`.
         """
 
     @abc.abstractmethod
-    def get_messages_view(self) -> CacheView[snowflakes.Snowflake, messages.PartialMessage]:
+    def get_messages_view(self) -> CacheView[snowflakes.Snowflake, messages.Message]:
         """Get a view of all the message objects in the cache.
 
         Returns
         -------
-        CacheView[hikari.snowflakes.Snowflake, hikari.messages.PartialMessage]
+        CacheView[hikari.snowflakes.Snowflake, hikari.messages.Message]
             A view of message objects found in the cache.
         """
 
     @abc.abstractmethod
-    def set_message(self, message: messages.PartialMessage) -> None:
+    def set_message(self, message: messages.Message, /) -> None:
         """Add a message object to the cache.
 
         Parameters
         ----------
-        message : hikari.messages.PartialMessage
+        message : hikari.messages.Message
             The object of the message to add to the cache.
 
         Raises
@@ -1664,18 +1664,18 @@ class MutableCache(Cache, abc.ABC):
 
     @abc.abstractmethod
     def update_message(
-        self, message: messages.PartialMessage
-    ) -> typing.Tuple[typing.Optional[messages.PartialMessage], typing.Optional[messages.PartialMessage]]:
+        self, message: typing.Union[messages.PartialMessage, messages.Message], /
+    ) -> typing.Tuple[typing.Optional[messages.Message], typing.Optional[messages.Message]]:
         """Update a message in the cache.
 
         Parameters
         ----------
-        message : hikari.messages.PartialMessage
+        message : typing.Union[hikari.messages.PartialMessage, hikari.messages.Message]
             The object of the message to update in the cache.
 
         Returns
         -------
-        typing.Tuple[typing.Optional[hikari.messages.PartialMessage], typing.Optional[hikari.messages.PartialMessage]]
+        typing.Tuple[typing.Optional[hikari.messages.Message], typing.Optional[hikari.messages.Message]]
             A tuple of the old cached message object if found (else `builtins.None`)
             and the new cached message object if it could be cached (else
             `builtins.None`).
--- a/hikari/api/cache.py
+++ b/hikari/api/cache.py
@@ -1587,33 +1587,14 @@ class MutableCache(Cache, abc.ABC):
         """
 
     @abc.abstractmethod
-    def delete_message(
-        self, message_id: snowflakes.Snowflake
-    ) -> None:
+    def delete_message(self, message_id: snowflakes.Snowflake) -> None:
         """Remove a message object from the cache.
-        
+
         Parameters
         ----------
         message_id : snowflakes.Snowflake
             The ID of the messages to remove the cache.
-        
-        Raises
-        ------
-        builtins.NotImplementedError
-            When called on a stateless cache implementation.
-        """
-    
-    @abc.abstractmethod
-    def delete_messages(
-        self, message_ids: typing.Sequence[snowflakes.Snowflake]
-    ) -> None:
-        """Remove a messages object from the cache.
-        
-        Parameters
-        ----------
-        message_ids : typing.Sequence[snowflakes.Snowflake]
-            The IDs of the messages to remove the cache.
-        
+
         Raises
         ------
         builtins.NotImplementedError
@@ -1621,9 +1602,22 @@ class MutableCache(Cache, abc.ABC):
         """
 
     @abc.abstractmethod
-    def get_message(
-        self, message_id: snowflakes.Snowflake
-    ) -> typing.Optional[messages.PartialMessage]:
+    def delete_messages(self, message_ids: typing.AbstractSet[snowflakes.Snowflake]) -> None:
+        """Remove a messages object from the cache.
+
+        Parameters
+        ----------
+        message_ids : typing.Sequence[snowflakes.Snowflake]
+            The IDs of the messages to remove the cache.
+
+        Raises
+        ------
+        builtins.NotImplementedError
+            When called on a stateless cache implementation.
+        """
+
+    @abc.abstractmethod
+    def get_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
         """Get a message object from the cache.
 
         Parameters
@@ -1638,16 +1632,14 @@ class MutableCache(Cache, abc.ABC):
         """
 
     @abc.abstractmethod
-    def set_message(
-        self, message: messages.PartialMessage
-    ) -> None:
+    def set_message(self, message: messages.PartialMessage) -> None:
         """Add a message object to the cache.
 
         Parameters
         ----------
         message : messages.PartialMessage
             The object of the message to add to the cache.
-        
+
         Raises
         ------
         builtins.NotImplementedError
@@ -1664,7 +1656,7 @@ class MutableCache(Cache, abc.ABC):
         ----------
         message : messages.PartialMessage
             The object of the message to update in the cache.
-        
+
         Returns
         -------
         typing.Tuple[typing.Optional[messages.PartialMessage], typing.Optional[messages.PartialMessage]]
--- a/hikari/api/cache.py
+++ b/hikari/api/cache.py
@@ -34,11 +34,11 @@ if typing.TYPE_CHECKING:
     from hikari import emojis
     from hikari import guilds
     from hikari import invites
+    from hikari import messages
     from hikari import presences
     from hikari import snowflakes
     from hikari import users
     from hikari import voices
-    from hikari import messages
 
 _KeyT = typing.TypeVar("_KeyT", bound=typing.Hashable)
 _ValueT = typing.TypeVar("_ValueT")
@@ -1586,15 +1586,32 @@ class MutableCache(Cache, abc.ABC):
             When called on a stateless cache implementation.
         """
 
-    @abc.abstractclassmethod
+    @abc.abstractmethod
     def delete_message(
-        self, message_ids: typing.Union[typing.AbstractSet[snowflakes.Snowflake], snowflakes.Snowflake]
+        self, message_id: snowflakes.Snowflake
+    ) -> None:
+        """Remove a message object from the cache.
+        
+        Parameters
+        ----------
+        message_id : snowflakes.Snowflake
+            The ID of the messages to remove the cache.
+        
+        Raises
+        ------
+        builtins.NotImplementedError
+            When called on a stateless cache implementation.
+        """
+    
+    @abc.abstractmethod
+    def delete_messages(
+        self, message_ids: typing.Sequence[snowflakes.Snowflake]
     ) -> None:
         """Remove a messages object from the cache.
         
         Parameters
         ----------
-        message_ids : typing.Union[typing.AbstractSet[snowflakes.Snowflake], snowflakes.Snowflake]
+        message_ids : typing.Sequence[snowflakes.Snowflake]
             The IDs of the messages to remove the cache.
         
         Raises
@@ -1603,7 +1620,7 @@ class MutableCache(Cache, abc.ABC):
             When called on a stateless cache implementation.
         """
 
-    @abc.abstractclassmethod
+    @abc.abstractmethod
     def get_message(
         self, message_id: snowflakes.Snowflake
     ) -> typing.Optional[messages.PartialMessage]:
@@ -1620,7 +1637,7 @@ class MutableCache(Cache, abc.ABC):
             The object of the message found in the cache or `builtins.None`.
         """
 
-    @abc.abstractclassmethod
+    @abc.abstractmethod
     def set_message(
         self, message: messages.PartialMessage
     ) -> None:
@@ -1637,7 +1654,7 @@ class MutableCache(Cache, abc.ABC):
             When called on a stateless cache implementation.
         """
 
-    @abc.abstractclassmethod
+    @abc.abstractmethod
     def update_message(
         self, message: messages.PartialMessage
     ) -> typing.Tuple[typing.Optional[messages.PartialMessage], typing.Optional[messages.PartialMessage]]:
--- a/hikari/api/cache.py
+++ b/hikari/api/cache.py
@@ -38,6 +38,7 @@ if typing.TYPE_CHECKING:
     from hikari import snowflakes
     from hikari import users
     from hikari import voices
+    from hikari import messages
 
 _KeyT = typing.TypeVar("_KeyT", bound=typing.Hashable)
 _ValueT = typing.TypeVar("_ValueT")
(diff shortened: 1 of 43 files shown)


hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,27 @@ def update_voice_state(
...
         self.set_voice_state(voice_state)
         return cached_voice_state, self.get_voice_state(voice_state.guild_id, voice_state.user_id)
+
+    def delete_message(self, message_id: snowflakes.Snowflake) -> None:
+        if message_id in self._message_entries:
+            self._message_entries.pop(message_id)
+
+    def delete_messages(self, message_ids: typing.AbstractSet[snowflakes.Snowflake]) -> None:
+        for message_id in message_ids:
+            self._message_entries.pop(message_id)
```
davfsa, line 1615: To keep it consistent, these should return the message/s that are deleted

hikari/impl/stateful_event_manager.py
```diff
@@ -200,7 +203,11 @@ async def on_guild_ban_remove(self, shard: gateway_shard.GatewayShard, payload:
 
     async def on_guild_emojis_update(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
         """See https://discord.com/developers/docs/topics/gateway#guild-emojis-update for more info."""
-        event = self._app.event_factory.deserialize_guild_emojis_update_event(shard, payload)
+        old = [
+            self._app.entity_factory.deserialize_known_custom_emoji(emoji, guild_id=Snowflake(payload["guild_id"]))
+            for emoji in payload["emojis"]
+        ]
```
davfsa, line 209: Why are the old ones the ones that we get 🤔

tests/hikari/impl/test_stateless_event_manager.py
```diff
@@ -95,7 +95,7 @@ async def test_on_channel_update(self, event_manager, shard, payload):
         await event_manager.on_channel_update(shard, payload)
 
         event_manager.dispatch.assert_awaited_once_with(event)
-        event_manager._app.event_factory.deserialize_channel_update_event.assert_called_once_with(shard, payload)
+        event_manager._app.event_factory.deserialize_channel_update_event.assert_called_once_with(shard, payload, None)
```
davfsa, line 98: These tests should be changed to also make sure they fetch from the cache. This will require mocking `._cache` and then making sure they are getting called with what you expect

hikari/impl/bot.py
```diff
@@ -270,6 +270,7 @@ def __init__(
         banner: typing.Optional[str] = "hikari",
         chunking_limit: int = 200,
         enable_cache: bool = True,
+        max_messages: int = 300,
```
davfsa, line 273: You will need to change the docstring of `BotApp` to also include this change.

hikari/api/event_factory.py
```diff
@@ -697,6 +747,8 @@ def deserialize_voice_state_update_event(
             The shard that emitted this event.
         payload : hikari.internal.data_binding.JSONObject
             The dict payload to parse.
+        old_state: typing.Optional[hikari.voices.VoiceState]
+            The VoiceState object or 'builtins.None'.
```
davfsa, line 751: `builtins.None` should be enclosed in "`" not "'". Also, specifying that it can be None here is redundant as it is already in the typehint. (This also applies to all the other ones)

hikari/events/channel_events.py
```diff
@@ -279,6 +279,14 @@ class GuildChannelUpdateEvent(GuildChannelEvent, ChannelUpdateEvent):
     shard: gateway_shard.GatewayShard = attr.ib(metadata={attr_extensions.SKIP_DEEP_COPY: True})
     # <<inherited docstring from ShardEvent>>.
 
+    old_channel: typing.Optional[channels.GuildChannel] = attr.ib(repr=True)
+    """Old guild channel object from cache or 'builtins.None'.
+
+    Returns
+    -------
+    hikari.channels.GuildChannel
+    """
```
davfsa, line 288: This doesnt need a return since its an attribute. Also, a better wording for the docstring is 
```suggestion
    """The old guild channel object.
    
    This will be `builtins.None` if it's missing from the cache.
    """
```
Again, this applies for all the other ones

hikari/api/event_factory.py
```diff
@@ -657,6 +702,8 @@ def deserialize_own_user_update_event(
             The shard that emitted this event.
         payload : hikari.internal.data_binding.JSONObject
             The dict payload to parse.
+        old_user: typing.Optional[hikaei.users.OwnUser]
```
davfsa, line 705:
```suggestion
        old_user: typing.Optional[hikari.users.OwnUser]
```

hikari/api/cache.py
```diff
@@ -1584,3 +1585,87 @@ def update_voice_state(
...
+        ----------
+        message : messages.PartialMessage
+            The object of the message to update in the cache.
+
+        Returns
+        -------
+        typing.Tuple[typing.Optional[messages.PartialMessage], typing.Optional[messages.PartialMessage]]
+            A tuple of the old cached message object if found (else `builtins.None`
+            and the new cached message object if it could be cached (else
+            `builtins.None`).
```
davfsa, line 1665:
```suggestion
        typing.Tuple[typing.Optional[messages.PartialMessage], typing.Optional[messages.PartialMessage]]
            A tuple of the old cached message object if found (else `builtins.None`)
            and the new cached message object if it could be cached (else
            `builtins.None`).
```

davfsa (changes requested):
First off, thanks for taking the time to work on this. Just some little nickpicks

hikari/impl/stateless_cache.py
```diff
@@ -390,3 +391,20 @@ def update_voice_state(
...
         raise self._no_cache()
+
+    def delete_message(self, message_id: snowflakes.Snowflake) -> None:
+        raise self._no_cache()
+
+    def delete_messages(self, message_ids: typing.AbstractSet[snowflakes.Snowflake]) -> None:
+        raise self._no_cache()
+
+    def get_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
+        return None
```
FasterSpeeding, line 402: don't we need a get messages view method here?

hikari/impl/stateless_cache.py
```diff
@@ -390,3 +391,20 @@ def update_voice_state(
         self, voice_state: voices.VoiceState, /
     ) -> typing.Tuple[typing.Optional[voices.VoiceState], typing.Optional[voices.VoiceState]]:
         raise self._no_cache()
+
```
FasterSpeeding, line 394: likely a clear messages one for the sake of keeping the interface somewhat consistent as well

hikari/impl/stateful_cache.py
```diff
@@ -111,9 +114,11 @@ class StatefulCacheImpl(cache.MutableCache):
         cache_utility.GenericRefWrapper[emojis.CustomEmoji],
     ]
     _user_entries: collections.ExtendedMutableMapping[snowflakes.Snowflake, cache_utility.GenericRefWrapper[users.User]]
+    _message_entries: collections.ExtendedMutableMapping[snowflakes.Snowflake, messages.PartialMessage]
     _intents: intents_.Intents
+    max_messages: int
 
-    def __init__(self, app: traits.RESTAware, intents: intents_.Intents) -> None:
+    def __init__(self, app: traits.RESTAware, intents: intents_.Intents, max_messages: int) -> None:
```
FasterSpeeding, line 121: This might not be the best way to handle configuration for the cache as in the future there will be more configuration such as all resources being individually togglable and the unlikely but still to be considered approach to configuration the caching approach for specific resources. Might want to make a dataclass for the cache config preliminarily but i'm still thinking over the best approach for cache config personally

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,27 @@ def update_voice_state(
         cached_voice_state = self.get_voice_state(voice_state.guild_id, voice_state.user_id)
         self.set_voice_state(voice_state)
         return cached_voice_state, self.get_voice_state(voice_state.guild_id, voice_state.user_id)
+
+    def delete_message(self, message_id: snowflakes.Snowflake) -> None:
+        if message_id in self._message_entries:
+            self._message_entries.pop(message_id)
+
+    def delete_messages(self, message_ids: typing.AbstractSet[snowflakes.Snowflake]) -> None:
```
FasterSpeeding, line 1613: is there any reason this has to be a set specifically, could this just be an iterable of snowflakes?

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,27 @@ def update_voice_state(
...
+        for message_id in message_ids:
+            self._message_entries.pop(message_id)
+
+    def get_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
+        return self._message_entries.get(message_id)
+
+    def set_message(self, message: messages.PartialMessage) -> None:
+        if len(self._message_entries) >= self.max_messages:
```
FasterSpeeding, line 1621: rather than specifically handling this logic here it might make more sense to use collections.LimitedCapacityCacheMap here for the time being (this approach would also allow us to slot in different caching strategies instead of binding the logic here to a single approach).

hikari/impl/stateful_event_manager.py
```diff
@@ -28,6 +28,7 @@
 import asyncio
 import typing
 
+from hikari import Snowflake
```
FasterSpeeding, line 31: For lib consistency we never want to import classes directly, we'd likely do (from hikari import snowflakes) then access snowflakes.Snowflake. Also did `nox -s isort` really order these invites like this?

hikari/events/guild_events.py
```diff
@@ -265,6 +265,14 @@ class GuildUpdateEvent(GuildEvent):
     shard: gateway_shard.GatewayShard = attr.ib(metadata={attr_extensions.SKIP_DEEP_COPY: True})
     # <<inherited docstring from ShardEvent>>.
 
+    old_guild: typing.Optional[guilds.Guild] = attr.ib()
+    """Old guild object from cache or 'builtins.None'.
+
+    Returns
+    -------
+    hikari.guilds.Guild
```
FasterSpeeding, line 273: would we not usually have descriptions for the returned field even if it just essentially dupes the top bit?

hikari/impl/stateful_cache.py
FasterSpeeding, line 1613: is there any benefit to having a special case multi-delete method here rather than just looping through and deleting each one specifically in the place where this is being called?

tests/hikari/impl/test_stateless_event_manager.py
davfsa: Nvm, just realized this is the stateless one. :sweat_smile:.

But we would still need the event_factory tests to make sure the field gets set

hikari/events/guild_events.py
```diff
@@ -265,6 +266,12 @@ class GuildUpdateEvent(GuildEvent):
     shard: gateway_shard.GatewayShard = attr.ib(metadata={attr_extensions.SKIP_DEEP_COPY: True})
     # <<inherited docstring from ShardEvent>>.
 
+    old_guild: typing.Optional[guilds.Guild] = attr.ib()
+    """The old guild object.
+
+    This will be `builtins.None` if guild missing from the cache.
+    """
```
davfsa, line 273: Just a nickpick to make everything look better and cause less confusion. This can also be applied to all the other `old_x` fields

```suggestion
    """The old guild object.

    This will be `builtins.None` if the guild missing from the cache.
    """
```

hikari/events/guild_events.py
```diff
@@ -390,6 +397,12 @@ class EmojisUpdateEvent(GuildEvent):
     guild_id: snowflakes.Snowflake = attr.ib()
     # <<inherited docstring from GuildEvent>>.
 
+    old_emojis: typing.Optional[cache.CacheView[snowflakes.Snowflake, emojis_.KnownCustomEmoji]] = attr.ib()
```
davfsa, line 400: For consistency this should not be a CacheView. You should make this a Mapping instead (will require changing when its passed here)

hikari/impl/event_factory.py
```diff
@@ -240,14 +255,19 @@ def deserialize_guild_ban_remove_event(
         )
 
     def deserialize_guild_emojis_update_event(
-        self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject
+        self,
+        shard: gateway_shard.GatewayShard,
+        payload: data_binding.JSONObject,
+        old_emojis: typing.Optional[cache.CacheView[snowflakes.Snowflake, emojis_models.KnownCustomEmoji]],
```
davfsa, line 261: Same as the other comment, mapping should go here

davfsa (changes requested):
Some nickpicks and more comments

hikari/impl/stateful_event_manager.py
```diff
@@ -200,7 +203,12 @@ async def on_guild_ban_remove(self, shard: gateway_shard.GatewayShard, payload:
 
     async def on_guild_emojis_update(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
         """See https://discord.com/developers/docs/topics/gateway#guild-emojis-update for more info."""
-        event = self._app.event_factory.deserialize_guild_emojis_update_event(shard, payload)
+        old = []
+        for emoji_id in self._cache.get_emojis_view_for_guild(snowflakes.Snowflake(payload["guild_id"])):
+            if emoji := self._cache.get_emoji(emoji_id):
+                old.append(emoji)
```
davfsa, line 209: This is really inecessary this could simply be this:
```suggestion
        emojis_view = self._cache.get_emojis_view_for_guild(snowflakes.Snowflake(payload["guild_id"]))
        old = list(emojis_view.values())
```

davfsa (changes requested):

davfsa:
@Reliku linting pipeline is failing. Run `nox -s reformat-code` to fix it.

PS: Could you please go back and look at the comments we made? There are some stuff you still haven't fixed (mostly tests)

davfsa (changes requested):
Tests look really good so far. Will have another look when the TODO's are implemented :+1:

davfsa (approved):
Looks really nice now :+1: 

Really good job and thanks for keeping up with us :smile:. I'm going to merge this into a separate branch when @FasterSpeeding gives the green light too so that I can work on the cache configuration, which you already started on

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,32 @@ def update_voice_state(
         cached_voice_state = self.get_voice_state(voice_state.guild_id, voice_state.user_id)
         self.set_voice_state(voice_state)
         return cached_voice_state, self.get_voice_state(voice_state.guild_id, voice_state.user_id)
+
+    def clear_messages(self) -> cache.CacheView[snowflakes.Snowflake, messages.PartialMessage]:
```
FasterSpeeding, line 1609: Random question but what's the point of this cache store, what's the use case for storing diff state updates for messages (which is what a PartialMessage is) alongside full messages rather than just full messages? (also wouldn't this potentially be replacing full message objects with message diff update objects?)

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,32 @@ def update_voice_state(
...
+        messages = self._message_entries
+        self._message_entries = collections.LimitedCapacityCacheMap(limit=self.max_messages)
+
+        return cache_utility.StatefulCacheMappingView(messages)
+
+    def delete_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
+        return self._message_entries.pop(message_id) if message_id in self._message_entries else None
+
+    def get_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
+        return self._message_entries.get(message_id)
```
FasterSpeeding, line 1622: you should be copying the message and user object attached to it here (copy.copy, not deepcopy)

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,32 @@ def update_voice_state(
...
+    def clear_messages(self) -> cache.CacheView[snowflakes.Snowflake, messages.PartialMessage]:
+        if not self._message_entries:
+            return cache_utility.EmptyCacheView()
+
+        messages = self._message_entries
+        self._message_entries = collections.LimitedCapacityCacheMap(limit=self.max_messages)
+
+        return cache_utility.StatefulCacheMappingView(messages)
+
+    def delete_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
```
FasterSpeeding, line 1618: this argument should be positional only

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,32 @@ def update_voice_state(
...
+        return self._message_entries.pop(message_id) if message_id in self._message_entries else None
+
+    def get_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
```
FasterSpeeding, line 1621: this argument should be positional only

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,32 @@ def update_voice_state(
...
+        return self._message_entries.get(message_id)
+
+    def get_messages_view(self) -> cache.CacheView[snowflakes.Snowflake, messages.PartialMessage]:
+        cached_messages = self._message_entries.freeze()
+        return cache_utility.StatefulCacheMappingView(cached_messages)
+
+    def set_message(self, message: messages.PartialMessage) -> None:
```
FasterSpeeding, line 1628: this argument should be positional only

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,32 @@ def update_voice_state(
...
+        self._message_entries[message.id] = message
+
+    def update_message(
```
FasterSpeeding, line 1631: this argument should be positional only

hikari/api/cache.py
```diff
@@ -1584,3 +1585,103 @@ def update_voice_state(
...
+            A view of message objects that were removed from the cache.
+
+        Raises
+        ------
+        builtins.NotImplementedError
+            When called on a stateless cache implementation.
+        """
+
+    @abc.abstractmethod
+    def delete_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
```
FasterSpeeding, line 1605: this argument should be positional only

hikari/api/cache.py
```diff
@@ -1584,3 +1585,103 @@ def update_voice_state(
...
+            else `builtins.None`.
+
+        Raises
+        ------
+        builtins.NotImplementedError
+            When called on a stateless cache implementation.
+        """
+
+    @abc.abstractmethod
+    def get_message(self, message_id: snowflakes.Snowflake) -> typing.Optional[messages.PartialMessage]:
```
FasterSpeeding, line 1626: this argument should be positional only

hikari/api/cache.py
```diff
@@ -1584,3 +1585,103 @@ def update_voice_state(
...
+        """Get a view of all the message objects in the cache.
+
+        Returns
+        -------
+        CacheView[hikari.snowflakes.Snowflake, hikari.messages.PartialMessage]
+            A view of message objects found in the cache.
+        """
+
+    @abc.abstractmethod
+    def set_message(self, message: messages.PartialMessage) -> None:
```
FasterSpeeding, line 1651: this argument should be positional only

hikari/api/cache.py
```diff
@@ -1584,3 +1585,103 @@ def update_voice_state(
...
+
+        Raises
+        ------
+        builtins.NotImplementedError
+            When called on a stateless cache implementation.
+        """
+
+    @abc.abstractmethod
+    def update_message(
+        self, message: messages.PartialMessage
```
FasterSpeeding, line 1667: this argument should be positional only

hikari/api/event_factory.py
```diff
@@ -70,7 +77,10 @@ def deserialize_channel_create_event(
         """
 
     def deserialize_channel_update_event(
-        self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject
+        self,
+        shard: gateway_shard.GatewayShard,
+        payload: data_binding.JSONObject,
+        old_channel: typing.Optional[channel_models.GuildChannel],
```
FasterSpeeding, line 83: it might be more consistent to have the old_x arguments be keyword only

hikari/impl/stateful_cache.py
```diff
@@ -125,7 +130,9 @@ def __init__(self, app: traits.RESTAware, intents: intents_.Intents) -> None:
         # found attached to cached presence activities.
         self._unknown_custom_emoji_entries = collections.FreezableDict()
         self._user_entries = collections.FreezableDict()
+        self._message_entries = collections.LimitedCapacityCacheMap(limit=max_messages)
         self._intents = intents
+        self.max_messages = max_messages
```
FasterSpeeding, line 135: We likely don't want to save the max messages as a public attribute like this as this'd mislead some people to think that modifying it will modify the amount of messages this cache will store

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1605,32 @@ def update_voice_state(
         cached_voice_state = self.get_voice_state(voice_state.guild_id, voice_state.user_id)
         self.set_voice_state(voice_state)
         return cached_voice_state, self.get_voice_state(voice_state.guild_id, voice_state.user_id)
+
+    def clear_messages(self) -> cache.CacheView[snowflakes.Snowflake, messages.PartialMessage]:
+        if not self._message_entries:
+            return cache_utility.EmptyCacheView()
+
+        messages = self._message_entries
```
FasterSpeeding, line 1613: might want to do 
```python
cached_messages = self._message_entries.copy
self._message_entries.clear()
```
if you want to preserve the max_messages without tracking it

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1602,60 @@ def update_voice_state(
...
+        return cache_utility.StatefulCacheMappingView(cached_messages)
+
+    def set_message(self, message: messages.Message, /) -> None:
+        self._message_entries[message.id] = copy.copy(message)
+
+    def update_message(
+        self, message: typing.Union[messages.PartialMessage, messages.Message], /
+    ) -> typing.Tuple[typing.Optional[messages.Message], typing.Optional[messages.Message]]:
+        cached_message = self.get_message(message.id)
+        msg = copy.copy(cached_message)
```
FasterSpeeding, line 1637: copied_message might be a nicer name than "msg" for this variable btw and we only have to copy this if message is PartialMessage so might want to copy it in that block of code.

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1602,60 @@ def update_voice_state(
...
+
+        if msg:
```
FasterSpeeding, line 1639: would a structure along the lines of 

```python
if message is messages.Message
   set message
else:
   assume partial and check then set each attribute
   set the copied message

return cached_message, self.get_message
```

possible be nicer to read

hikari/impl/stateful_cache.py
```diff
@@ -1598,3 +1602,60 @@ def update_voice_state(
...
+                ]
+
+                for key in keys:
+                    new_value = getattr(message, key)
+                    if new_value is not undefined.UNDEFINED:
+                        setattr(msg, key, new_value)
+
+                message = msg
+
+            self.set_message(message)
```
FasterSpeeding, line 1659: if it's a full message object then wouldn't we want to set it regardless of whether the message was previously cached or not to be more inline with how this cache handles "update" methods

davfsa:
@FasterSpeeding This pr still waits a review. I would like to get this one merged soon so that it can be in the next deploy :+1:

FasterSpeeding (approved):

PR #347, opened 2020-11-10
Implement a lock to help ratelimiting handling

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -66,7 +66,7 @@ from hikari import urls
 from hikari import users
 from hikari.api import cache as cache_
 from hikari.api import rest as rest_api
-from hikari.impl import buckets
+from hikari.impl import buckets as buckets_
 from hikari.impl import entity_factory as entity_factory_impl
 from hikari.impl import rate_limits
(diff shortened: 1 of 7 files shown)


FasterSpeeding (dismissed):

hikari/impl/rest.py
```diff
@@ -551,74 +553,74 @@ async def _request(
...
-                await asyncio.gather(self.buckets.acquire(compiled_route), self.global_rate_limit.acquire())
-
-                uuid = time.uuid()
-
-                if _LOGGER.getEffectiveLevel() <= ux.TRACE:
-                    _LOGGER.log(
-                        ux.TRACE,
-                        "%s %s %s\n%s",
-                        uuid,
+            async with self._lock:
```
FasterSpeeding, line 558: could we be more specific about when we aquire this lock?

davfsa: I decided to implement a per bucket lock to stop 2 requests going towards the same bucket at the same time. Doing that rn, will update pr when im done :+1:

FasterSpeeding (approved):

PR #350, opened 2020-11-14
Add logic and abstact method for search members endpoint

### LINKED ISSUE #323: Search Guild Members REST Endpoint
~~https://discord.com/developers/docs/resources/guild#list-guild-members~~
https://github.com/discord/discord-api-docs/pull/1577

FasterSpeeding:
### Checklist

### Related issues
Closes #323

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -3607,6 +3607,52 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         """
 
     @abc.abstractmethod
+    async def search_members(
+        self,
+        guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
+        name: str,
+    ) -> typing.Sequence[guilds.Member]:
+        """Search the members in a guild by nickname and username.
+
+        Parameters
+        ----------
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            The object or ID of the guild to search members in.
+        name : str
+            The query to match username(s) and nickname(s) against.
+
+        Returns
+        -------
+        typing.Sequence[hikari.guilds.Member]
+            A sequence of the members who matched the provided `name`.
+
+        Raises
+        ------
+        hikari.errors.UnauthorizedError
+            If you are unauthorized to make the request (invalid/missing token).
+        hikari.errors.NotFoundError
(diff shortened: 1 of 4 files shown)


hikari/impl/rest.py
```diff
@@ -1997,6 +1997,21 @@ def fetch_members(
...
 
+    async def search_members(
+        self,
+        guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
+        name: str,
+    ) -> typing.Sequence[guilds.Member]:
+        route = routes.GET_GUILD_MEMBERS_SEARCH.compile(guild=guild)
+        query = data_binding.StringMapBuilder()
+        query.put("query", name)
+        query.put("limit", 1000)
```
davfsa, line 2008: Shouldnt we also allow the user to pass the limit?

hikari/api/rest.py
```diff
@@ -3606,6 +3606,52 @@ def fetch_members(
...
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
+        hikari.errors.InternalServerError
+            If an internal error occurs on Discord while handling the request.
+
+        !!! note
+            Unlike `RESTClient.fetch_members` this endpoint isn't paginated and
+            therefore will return all the members in one go rather than needing
+            to be asynchronously iterated over.
```
davfsa, line 3652: While you are at this, could you also add some documentation to `fetch_members` that it requires intents to be used?

hikari/impl/rest.py
FasterSpeeding: we don't allow this for any "limited" endpoints right now and to allow it here seems inconsistent

davfsa: All the other endpoints are chunked, this is the only one that isnt (by the looks of it). So I don't see why not

FasterSpeeding: plus we don't have any precedent for specifying defaults for meta fields like that so either we inconsistently make the actual default for it a part of the interface which would be inconsistent with how we currently handle defaults or we make it implementation detail essentially making the default useless as it may be anything from 1 to 1000 dependent on implementation detail

davfsa (approved):

PR #366, opened 2020-11-21
Implement user DM flow

### LINKED ISSUE #237: Implement user DM flow for sending new messages.

davfsa:
### Checklist

### Related issues
Closes #237

### DIFF
--- a/hikari/guilds.py
+++ b/hikari/guilds.py
@@ -439,6 +439,9 @@ class Member(users.User):
         """
         return await self.app.rest.fetch_member(self.guild_id, self.id)
 
+    async def fetch_dm_channel(self) -> channels_.DMChannel:
+        return await self.user.fetch_dm_channel()
+
     async def ban(
         self,
         *,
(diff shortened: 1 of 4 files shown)


hikari/guilds.py
```diff
@@ -299,6 +299,14 @@ class Member(users.User):
     user: users.User = attr.ib(repr=True)
     """This member's corresponding user object."""
 
+    @property
+    def _dm_channel(self) -> typing.Optional[channels_.DMChannel]:
```
FasterSpeeding, line 303: Should we be accessing private attributes like this? Would it not make more sense for member’s important of send and fetch DM channel To just call self.user.send and self.user.fetch_dm_channel

hikari/users.py
```diff
@@ -120,6 +125,15 @@ class PartialUser(snowflakes.Unique, abc.ABC):
     API call.
     """
 
+    @property
+    @abc.abstractmethod
+    def _dm_channel(self) -> typing.Optional[channels.DMChannel]:
+        ...
+
+    @_dm_channel.setter
```
FasterSpeeding, line 133: Shouldn’t this be implementation detail of the users implementing this rather than a part of the interface?

FasterSpeeding (approved):

PR #369, opened 2020-11-22
Switch to using `x-ratelimit-reset-after` as reset_after

FasterSpeeding:
### Summary
Switch to using `x-ratelimit-reset-after` as reset_after
* This is generally speaking more accurate than calculating this value from the `Date` header
  and `X-RateLimit-Reset` headers as these seem to drift away and on add reaction endpoints
  were seen to by almost a second larger than `x-ratelimit-reset-after` in some cases
  (e.g 0.25 vs 0.403, 0.25 vs 0.425, 0.25 vs 1.098, 0.25 vs 0.361 and 0.25 vs 0.634)
* Make owner mandatory on application

### Checklist

### DIFF
--- a/hikari/impl/buckets.py
+++ b/hikari/impl/buckets.py
@@ -119,9 +119,6 @@ Once you have received your response, you are expected to extract the values of
 the vital rate limit headers manually and parse them to the correct data types.
 These headers are:
 
-* `Date`:
-    the response date on the server. This should be parsed to a
-    `datetime.datetime` using `email.utils.parsedate_to_datetime`.
 * `X-RateLimit-Limit`:
     an `builtins.int` describing the max requests in the bucket from empty to
     being rate limited.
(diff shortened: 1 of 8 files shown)


hikari/impl/buckets.py
```diff
@@ -578,16 +576,13 @@ def update_rate_limits(
             The `X-RateLimit-Remaining` header cast to an `builtins.int`.
         limit_header : builtins.int
             The `X-RateLimit-Limit`header cast to an `builtins.int`.
-        date_header : datetime.datetime
-            The `Date` header value as a `datetime.datetime`.
-        reset_at_header : datetime.datetime
-            The `X-RateLimit-Reset` header value as a `datetime.datetime`.
+        reset_after : builtins.float
+            The `x-ratelimit-reset-after` cast to a `builtins.float`.
```
davfsa, line 580: The huge docstring at the top of this file will also need updating

hikari/impl/rest.py
```diff
@@ -654,18 +654,14 @@ async def _parse_ratelimits(self, compiled_route: routes.CompiledRoute, response
         limit = int(resp_headers.get(_X_RATELIMIT_LIMIT_HEADER, "1"))
         remaining = int(resp_headers.get(_X_RATELIMIT_REMAINING_HEADER, "1"))
         bucket = resp_headers.get(_X_RATELIMIT_BUCKET_HEADER, "None")
-        reset_at = float(resp_headers.get(_X_RATELIMIT_RESET_HEADER, "0"))
         reset_after = float(resp_headers.get(_X_RATELIMIT_RESET_AFTER_HEADER, "0"))
-        reset_date = datetime.datetime.fromtimestamp(reset_at, tz=datetime.timezone.utc)
-        now_date = time.rfc7231_datetime_string_to_datetime(resp_headers[_DATE_HEADER])
```
davfsa, line 660: Removing this makes `rfc7231_datetime_string_to_datetime` not used anywhere else. Should we remove it?

davfsa (approved):

Forbidden-A:
I'd say either rename the reply function or make it reply by default..

PR #374, opened 2020-11-24
Fix bug on wait_for so that timeouts remove it from the waiters_set

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/event_manager_base.py
+++ b/hikari/impl/event_manager_base.py
@@ -223,7 +223,8 @@ class EventManagerBase(event_dispatcher.EventDispatcher):
                     tasks.append(self._invoke_callback(callback, event))
 
             if cls in self._waiters:
-                for predicate, future in tuple(self._waiters[cls]):
+                waiter_set = self._waiters[cls]
+                for predicate, future in tuple(waiter_set):
                     try:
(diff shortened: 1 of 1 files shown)


hikari/impl/event_manager_base.py
```diff
@@ -272,7 +272,11 @@ async def wait_for(
         pair = (predicate, future)
 
         waiter_set.add(pair)  # type: ignore[arg-type]
-        return await asyncio.wait_for(future, timeout=timeout)
+        try:
+            return await asyncio.wait_for(future, timeout=timeout)
+        except Exception:
+            waiter_set.remove(pair)  # type: ignore[arg-type]
```
FasterSpeeding, line 278: i assume this won't be removed anywhere else right?

davfsa: It does, so changed it so that its only removed on Timeout instead of all exceptions.

FasterSpeeding (approved):
I assume that on a timeout error specifically it won't have been removed anywhere else

davfsa:
> 
> 
> I assume that on a timeout error specifically it won't have been removed anywhere else

Yep, that is the issue this pr fixes 😄

PR #382, opened 2020-12-01
Several bugfixes

davfsa:
- New error `ComponentNotRunningError` that will be raised when trying to interact with a component that is not running
  - Actually closing when one of the shards has been manually shut down
  - `sent_close` not being set when calling `send_close`
  - Unused ratelimiter now hoocked up to prevent getting disconnected on too many requests
  - `_shards` is now populated as soon as the shard object is created to be able to be used through `bot.x` on shard events

### Checklist

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -234,6 +234,7 @@ class BotApp(traits.BotAware, event_dispatcher.EventDispatcher):
         "_executor",
         "_http_settings",
         "_intents",
+        "_is_alive",
         "_proxy_settings",
         "_raw_event_consumer",
         "_rest",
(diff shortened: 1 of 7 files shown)


hikari/impl/bot.py
```diff
@@ -861,20 +862,22 @@ async def start(
...
-            for started_shard in started_shards:
-                self._shards[started_shard.id] = started_shard
-
         await self.dispatch(lifetime_events.StartedEvent(app=self))
 
         _LOGGER.info("application started successfully in approx %.2f seconds", time.monotonic() - start_time)
 
+    def _check_if_alive(self) -> None:
+        if self._closing_event.is_set():
+            raise errors.ComponentNotRunningError("bot is not running so it cannot be interacted with")
```
FasterSpeeding, line 871: how would someone looking in from out be able to programmatically tell that a shard or client is closing or closed?

FasterSpeeding (dismissed):
lgtm asides the one general inquiry question

hikari/impl/bot.py
davfsa: Added an `is_alive` property to the bot. Thx for the spot

FasterSpeeding (approved):

PR #384, opened 2020-12-05
Cache configuration

### LINKED ISSUE #368: Cache configuration
Make some cache functionalities allowed to be enabled/disabled by using `CacheSettings`

davfsa:
### Checklist

### Related issues
Closes #368


hikari/impl/cache.py
```diff
@@ -114,16 +113,11 @@ class StatefulCacheImpl(cache.MutableCache):
...
-    def __init__(
-        self, app: traits.RESTAware, intents: intents_.Intents, settings: typing.Optional[config.CacheSettings]
-    ) -> None:
-        if settings is None:
-            settings = config.CacheSettings()
+    _message_entries: collections.ExtendedMutableMapping[snowflakes.Snowflake, messages.Message]
 
+    def __init__(self, app: traits.RESTAware, settings: config.CacheSettings) -> None:
         self._app = app
+        self._settings = settings
```
FasterSpeeding, line 120: might be nice to have some way to introspectively check the settings on the cache if we know which type it is. (so just impl detail where you'd have to instance check it first rather than as a part of the abc or protocol)

PR #385, opened 2020-12-06
Implement role tags

### LINKED ISSUE #345: Role tags
https://github.com/discord/discord-api-docs/issues/1537#issuecomment-656432643

davfsa:
### Checklist

### Related issues
Closes #345


hikari/guilds.py
```diff
@@ -551,6 +551,21 @@ class Role(PartialRole):
...
+    If `builtins.None`, this is not a bot role.
+    """
+
+    integration_id: typing.Optional[snowflakes.Snowflake] = attr.ib(eq=False, hash=False, repr=True)
+    """The ID of the integration this role belongs to.
+
+    If `builtins.None`, this is not a integration role.
+    """
+
+    is_premium_subscriber_role: bool = attr.ib(eq=False, hash=False, repr=True)
```
FasterSpeeding, line 566: Might be worth mentioning that this is "boosts" somewhere here as that's what it's more likely going to be known as

FasterSpeeding (dismissed):

hikari/guilds.py
davfsa: Also thought about that one, but then scrapped it cause discord specially specified it as "premium subscriber" so they might make it do something else with it in the future. 

Will add it to the docstring tho

PR #411, opened 2020-12-23
Make applications.TeamMember extend users.User plus chunk changes

FasterSpeeding:
### Summary
* request presences when chunking on guild create if possible
* Make applications.TeamMember extend users.User

### Checklist

### DIFF
--- a/hikari/applications.py
+++ b/hikari/applications.py
@@ -42,14 +42,15 @@ from hikari import files
 from hikari import guilds
 from hikari import snowflakes
 from hikari import urls
+from hikari import users
 from hikari.internal import attr_extensions
 from hikari.internal import enums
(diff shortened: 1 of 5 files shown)


davfsa (changes requested):
Could you please add integration tests for this too? I believe they will fail as you need to implement __eq__ (check guilds.Member for more info :+1:)

FasterSpeeding:
> Could you please add integration tests for this too? I believe they will fail as you need to implement **eq** (check guilds.Member for more info 👍)

The equality override and integration tests is already there, this pr just adds in delegation.

davfsa:
Oh, didnt see that 😅

davfsa (approved):

PR #388, opened 2020-12-14
Cache refactors

FasterSpeeding:
### Summary
Public interface refactors
* have OwnUser.send match the inherited signature so the type error can actually be reached
* Type UndefinedType.__bool__ as always returning False
* Remove public mutable user methods from the cache abc and impl
* Allow snowflakes.Snowflakeish to be used for IDs in calls to cache

Internal refactors:
* Remove user methods from the MutableCache interface as this should be managed internally
* Switch to tracking referenced entries through RefCells as opposed to by ID to lower look up time and building complexity
* Remove modifying user cache calls from the stateful event manager
* Switch message cache store to a relational style plus add in Message + relevant message cache data object impls.
* Also add in a weak referenced messages store
* Add "clear" method overrides to ExtendedMutableMapping impls to allow these to take advantage of the underlying dict's optimisations
* Add on_expire fields to ExtendedMutableMapping implementations to allow for triggering referential garbage collection within self managing cache stores.
* Refactor cache's yet to be hooked in intent error raising methods to warn instead
* Extract out the logic for internally called cache "set" methods to private attributes to allow for future configuration to not effect internal calls.
* Add WeakMap collection and Cell for possible use with user cache + test fixes
 (plus some internal consistency fixes)
* Switch from GuildRecord.__bool__() to .empty() for checking if a record is "empty"
* Switch away from using **kwargs in build_entity to specifying only a required "app" arg

### Checklist

- [x] I have run `nox` and all the pipelines have passed.
- [x] I have made unittests according to the code I have added/modified/deleted.

PR #412, opened 2020-12-27
Remove guild chunker and add cache configuration

### LINKED ISSUE #291: Stateless bot should be its own class.
I am currently refactoring message mention introspection; part of this is to update a part of the cache-aware.

As part of this, I want to split the standard bot definition into a stateless and stateful copy. This can then ensure the traits are clearer underneath in determining if the app has a proper cache or not.

This should also allow the complete destruction of the stateless cache, which is a large amount of useless wet code right now.

davfsa:
- Fix traits
   - Fix updating voice state payload
   - Remove stateless cache and stateless guild chunker
   - Clear out cache on bot close
   - Added all cache configuration
       - Lay bases for configurable cache
   - Temporarily removed guild_chunker
   - General cleanup

### Checklist

### Related issues
Closes #291

# Note
This is not finished and will probably be merged in an unfinished state as it overlaps with #388.


hikari/internal/cache.py
```diff
@@ -70,7 +70,7 @@
 """Type-hint for mapping values."""
 
 
-class StatefulCacheMappingView(cache.CacheView[KeyT, ValueT], typing.Generic[KeyT, ValueT]):
+class CacheMappingView(cache.CacheView[KeyT, ValueT], typing.Generic[KeyT, ValueT]):
```
FasterSpeeding, line 73: did we both coincidentally make the exact same name change in two separate draft prs? 👀

hikari/impl/cache.py
```diff
@@ -134,15 +133,16 @@ def __init__(
...
+    def _is_cache_enabled_for(self, setting: str) -> bool:
+        return self._settings.enable and getattr(self._settings, setting)
 
-    def _is_intent_enabled(self, intents: intents_.Intents, /) -> bool:
-        return (self._intents & intents) == intents
+    def clear(self) -> None:
+        if not self._is_cache_enabled_for("enable"):
+            return None
+
+        self._create_cache()
```
FasterSpeeding, line 145: what are there benefits like between this and calling "clear" on each resource to stuff like how quickly memory is deallocated/how long the call takes, etc etc

hikari/impl/cache.py
```diff
@@ -278,19 +290,28 @@ def _get_emojis_view(  # TODO: split out the two cases (specific guild vs global
...
     def get_emojis_view_for_guild(
         self, guild_id: snowflakes.Snowflake, /
     ) -> cache.CacheView[snowflakes.Snowflake, emojis.KnownCustomEmoji]:
+        if not self._is_cache_enabled_for("emojis"):
+            return cache_utility.EmptyCacheView()
+
         return self._get_emojis_view(guild_id=guild_id)
 
     def set_emoji(self, emoji: emojis.KnownCustomEmoji, /) -> None:
+        if not self._is_cache_enabled_for("emojis"):
```
FasterSpeeding, line 312: might as well make constants for these names

hikari/impl/stateless_event_manager.py
```diff
@@ -25,7 +25,6 @@
 
 __all__: typing.List[str] = ["StatelessEventManagerImpl"]
 
-import logging
```
FasterSpeeding, line 28: I assume you just haven't gotten around to this yet but do we still need the stateless vs stateful event manager?

hikari/impl/cache.py
davfsa: Have not tried it, will see the differences

hikari/impl/cache.py
davfsa: Was thinking about that, will do 👍

hikari/internal/cache.py
davfsa: lol

hikari/impl/stateless_event_manager.py
davfsa: For now I have made it so it uses stateless event manager if the whole cache is disabled. Before merging into master I will see what I do. The thing that is worrying me rn and kept me from merging them is iterating throught stuff and making a call for that. If its disabled, they wont really make sense. As I type this, I realize I can check if the cache is enabled for what I want to do before doing it. But will think about it. For now, this pr will have them separate.

hikari/impl/cache.py
davfsa:
```py
    def clear(self) -> None:
        if not self._is_cache_enabled_for(ENABLE):
            return None

        current, peak = tracemalloc.get_traced_memory()
        print(f"Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
        print("Collecting")
        gc.set_debug(gc.DEBUG_STATS)
        self._create_cache()
        gc.collect()
        gc.set_debug(0)
        print("Done collecting")
        current, peak = tracemalloc.get_traced_memory()
        print(f"Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
```
image

hikari/impl/guild_chunker.py
```diff
@@ -200,17 +200,8 @@ def update(self, event: shard_events.MemberChunkEvent) -> None:
         self.last_received = time.utc_datetime()
 
 
-class StatefulGuildChunkerImpl(chunker.GuildChunker):
-    """Guild chunker implementation.
-
-    Parameters
-    ----------
-    app : hikari.traits.BotAware
```
FasterSpeeding, line 208: not gonna document the parameters anymore?

hikari/impl/bot.py
```diff
@@ -332,16 +307,16 @@ def __init__(
...
 
     @property
     def cache(self) -> cache_.Cache:
         return self._cache
 
-    @property
-    def chunker(self) -> chunker_.GuildChunker:
-        return self._chunker
+    # @property
+    # def chunker(self) -> guild_chunker.GuildChunker:
```
FasterSpeeding, line 318: what're we doing with the chunker?

hikari/impl/cache.py
```diff
@@ -295,24 +357,39 @@ def _get_guilds_view(self, *, availability: bool) -> cache.CacheView[snowflakes.
...
+        if not self._is_cache_enabled_for(GUILDS):
+            return None
+
         guild_record = self._get_or_create_guild_record(guild.id)
         guild_record.guild = copy.copy(guild)
         guild_record.is_available = True
 
     def set_guild_availability(self, guild_id: snowflakes.Snowflakeish, is_available: bool, /) -> None:
+        if not self._is_cache_enabled_for(GUILDS):
+            return None
```
FasterSpeeding, line 381: could we not methods which perform/rely on lookups implicitly pass/return None after they find nothing? Would also mean changing clear methods to do `if not resource: return empty cache view`

hikari/config.py
```diff
@@ -397,8 +397,72 @@ class that will **NOT** enforce SSL verification. This is then stored
...
+    Defaults to `builtins.True`.
+    """
+
+    members: bool = attr.ib(default=True)
+    """Whether to enable the members cache.
+
+    Defaults to `builtins.True`.
+    """
+
+    channels: bool = attr.ib(default=True)
```
FasterSpeeding, line 420: might want to call this guild_channels and specify that it's only guild channels to be consistent with the cache

hikari/config.py
```diff
@@ -397,8 +397,72 @@ class that will **NOT** enforce SSL verification. This is then stored
 class CacheSettings:
     """Settings to control the cache."""
```
FasterSpeeding, line 399: also if this is a lib-wise config then it might be worth it having a property on the cache which exposes it's settings

hikari/impl/bot.py
davfsa: Since the chunker relies on the cache, but currently disabling the cache will not trigger the cache anywhere, im going to be rewriting it to work with both. Since this pr is already really big, it will be done before the feature branch is merged into master

hikari/impl/guild_chunker.py
davfsa: This was too keep consistency with other components. Thinking about it again, I'm going to actually document it on the ones that are missing

hikari/config.py
davfsa: Good idea, will add

hikari/impl/cache.py
davfsa: Hmm, yea, could change it to that

davfsa: After thinking about this, I decided not to implement it as I prefer (and its faster) to return the default immediately if we already know its not going to be there.

davfsa:
@FasterSpeeding Implemented the stuff you pointed out. Also rewrote checking for the settings being enabled, which I would like your opinion on

hikari/impl/cache.py
```diff
@@ -72,11 +95,11 @@ class StatefulCacheImpl(cache.MutableCache):
         "_unknown_custom_emoji_entries",
         "_user_entries",
         "_message_entries",
+        "_referenced_messages",
```
FasterSpeeding, line 98: you need to add the missing slots to MutableCache

hikari/impl/cache.py
```diff
@@ -128,13 +155,29 @@ def _assert_has_intent(self, intents: intents_.Intents, /) -> None:
                 stacklevel=1,
             )
 
+    def _is_cache_enabled_for(self, setting: str) -> bool:
+        return bool(self._settings_map["enable"] and self._settings_map[setting])
```
FasterSpeeding, line 159: would be more consistent to use ENABLE here

davfsa: My bad :sweat_smile:

hikari/impl/cache.py
davfsa: MutableCache doesnt have any slots. Is that what you mean?

FasterSpeeding: Slots aren't defined on mutable cache

FasterSpeeding (approved):

PR #416, opened 2021-01-05
Fix usage/setting of bot._is_alive and invalid channel type assertions

FasterSpeeding:
### Summary
Fix usage/setting of bot._is_alive and invalid channel type assertions

### Checklist

### DIFF
--- a/hikari/events/typing_events.py
+++ b/hikari/events/typing_events.py
@@ -171,7 +171,7 @@ class GuildTypingEvent(TypingEvent):
     """
 
     @property
-    def channel(self) -> typing.Union[channels.GuildTextChannel, channels.GuildNewsChannel]:
+    def channel(self) -> typing.Union[channels.GuildTextChannel, channels.GuildNewsChannel, None]:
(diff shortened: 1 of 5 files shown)


hikari/events/typing_events.py
```diff
@@ -171,7 +171,7 @@ class GuildTypingEvent(TypingEvent):
     """
 
     @property
-    def channel(self) -> typing.Union[channels.GuildTextChannel, channels.GuildNewsChannel]:
+    def channel(self) -> typing.Union[channels.GuildTextChannel, channels.GuildNewsChannel, None]:
```
davfsa, line 174: These are getting fixed in #412

FasterSpeeding: Don't see that file in its diff

davfsa: Nvm, it was implemented in #404 (which is missing the assert to allow `None`, got to fix that)

davfsa (approved):

PR #419, opened 2021-01-08
Rename EventDispatcher to EventManager and add a public raw event consumer system

FasterSpeeding:
### Summary
Proof of concept around allowing for raw listeners

### Checklist

### DIFF
--- a/hikari/api/event_dispatcher.py
+++ b/hikari/api/event_dispatcher.py
@@ -31,12 +31,17 @@ import typing
 
 if typing.TYPE_CHECKING:
     from hikari import event_stream
+    from hikari.api import shard as gateway_shard
     from hikari.events import base_events
+    from hikari.internal import data_binding
 
     EventT_co = typing.TypeVar("EventT_co", bound=base_events.Event, covariant=True)
     EventT_inv = typing.TypeVar("EventT_inv", bound=base_events.Event)
     PredicateT = typing.Callable[[EventT_co], bool]
     CallbackT = typing.Callable[[EventT_inv], typing.Coroutine[typing.Any, typing.Any, None]]
+    ConsumerT = typing.Callable[
+        [gateway_shard.GatewayShard, data_binding.JSONObject], typing.Coroutine[typing.Any, typing.Any, None]
+    ]
 
 
 class EventDispatcher(abc.ABC):
@@ -48,6 +53,76 @@ class EventDispatcher(abc.ABC):
     """
 
     __slots__: typing.Sequence[str] = ()
+
+    @abc.abstractmethod
+    def add_raw_consumer(self, name: str, consumer: ConsumerT, /) -> None:
+        """Register a given async callback to a raw event name.
+
+        Parameters
+        ----------
+        name : str
+            The case-insensitive name this event should be triggered based on.
+        consumer : typing.Callable[[gateway_shard.GatewayShard, hikari.internal.data_binding.JSONObject], typing.Coroutine[typing.Any, typing.Any, None]]
+            The async function to invoke on each raw event. This should take
+            two positional arguments, the shard that this event is being
+            triggered by and this raw event's payload.
+        """  # noqa: E501 - Line too long
+
+    @abc.abstractmethod
+    def get_raw_consumers(self, name: str, /) -> typing.Sequence[ConsumerT]:
+        """Get the async callbacks registered for a raw event.
+
+        Parameters
+        ----------
+        name : str
+            The case-insensitive name of the event to remove a raw consumer for.
+
+        Returns
+        -------
+        typing.Sequence[typing.Callable[[gateway_shard.GatewayShard, hikari.internal.data_binding.JSONObject], typing.Coroutine[typing.Any, typing.Any, None]]]
+            A sequence of the found async functions registered for the provided
+            event name, this will be an empty sequence if no consumers were found.
+        """  # noqa: E501 - Line too long
+
+    @abc.abstractmethod
+    def remove_raw_consumer(self, name: str, consumer: ConsumerT, /) -> None:
+        """Remove a registered async raw event consumer callback.
+
+        Parameters
+        ----------
+        name : str
+            The case-insensitive name of the event to remove a raw consumer for.
+        consumer : typing.Callable[[gateway_shard.GatewayShard, hikari.internal.data_binding.JSONObject], typing.Coroutine[typing.Any, typing.Any, None]]
+            The async function to remove from the registered raw consumers.
+
+        Raises
+        ------
+        LookupError
+            If the either `name` or `consumer` couldn't be found in the
+            registered raw consumers.
+        """  # noqa: E501 - Line too long
+
+    @abc.abstractmethod
+    def consume_raw_event(
+        self, shard: gateway_shard.GatewayShard, event_name: str, payload: data_binding.JSONObject
+    ) -> None:
+        """Consume a raw event.
+
+        Parameters
+        ----------
+        shard : hikari.api.gateway_shard.GatewayShard
+            Object of the shard that received this event.
+        event_name : str
+            The case-insensitive name of the event being triggered.
+        payload : hikari.internal.data_binding.JSONObject
+            Payload of the event being triggered.
+
+        Raises
+        ------
+        LookupError
+            If no registered raw consumers were found for the provided event
+            name.
+        """
 
     @abc.abstractmethod
     def dispatch(self, event: EventT_inv) -> asyncio.Future[typing.Any]:
@@ -388,4 +463,4 @@ class EventDispatcher(abc.ABC):
         Stream: `hikari.api.event_dispatcher.EventDispatcher.stream`
         Subscribe: `hikari.api.event_dispatcher.EventDispatcher.subscribe`
         Dispatch: `hikari.api.event_dispatcher.EventDispatcher.dispatch`
-        """  # noqa: E501 - Line too long
+        """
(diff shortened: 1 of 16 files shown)


FasterSpeeding:
This is just a random idea I had while checking out how much of a different calling `getattr` for every dispatch vs pre-loading them into a dict (spoiler pre-loading is marginally faster even with the other stuff i added over it), no idea if it'd make sense to have the raw consumer methods on the public interface for the event manager or not rn but just wanted to see what @davfsa has to say about this idea

hikari/impl/event_manager_base.py
```diff
@@ -71,23 +74,50 @@ class EventManagerBase(event_dispatcher.EventDispatcher):
...
+
+        return []
+
+    def remove_raw_consumer(self, name: str, consumer: ConsumerT, /) -> None:
+        name = name.casefold()
+        if name not in self._consumers or consumer not in self._consumers[name]:
+            raise LookupError(name, consumer)
+
+        self._consumers[name].remove(consumer)
+
```
davfsa, line 109: What could these two be used for? With the implementation you have, we could just make everything private and not provide an interface for it

hikari/impl/event_manager_base.py
```diff
@@ -71,23 +74,50 @@ class EventManagerBase(event_dispatcher.EventDispatcher):
...
-            callback = getattr(self, "on_" + event_name.lower())
-        except AttributeError:
+            callbacks = self._consumers[event_name.casefold()]
+        except KeyError:
             _LOGGER.debug("ignoring unknown event %s:\n    %r", event_name, payload)
         else:
-            asyncio.create_task(self._handle_dispatch(callback, shard, payload), name=f"dispatch {event_name}")
+            asyncio.gather(
+                *(self._handle_dispatch(callback, shard, payload) for callback in callbacks), return_exceptions=True
+            )
```
davfsa, line 120: Does using asyncio.gather work if you don't await it? Also, in what case would this be more than one? We could just limit it to 1

davfsa (commented):
LGTM, just some questions

hikari/impl/event_manager_base.py
FasterSpeeding: asyncio.gather doesn't need to be awaited

hikari/impl/event_manager_base.py
FasterSpeeding: well remove_raw_consumer is a consistency thing, since this thing isn't designed to be randomly replaced it'd make more sense for them to be able to remove consumers as well as add them.
get_raw_consumers exists for the sake of not leaving this as a black box which you work out what it's doing seeing as the fact that it loads raw consumers from attributes in itself is impl detail rather than a part of the interface. I doubt it'd be overly helpful but it's nicer than a black box

hikari/impl/event_manager_base.py
FasterSpeeding: More than one would allow ppl to register raw consumers along side the builtin ones (the use cases for this are likely a bit niche but they do exist and this would make the lib generally more flexible).
as a note gathering like this won't hurt performance unless they start registering a ton of raw consumers in which case that's on them and the performance effect probably won't be too noticeable anyway (we're talking microseconds difference in execution time here)

FasterSpeeding:
Ig the goal off this ended up unintentionally being turning the event manager from a black box where any inspection or modification relies on implementation detail to something that provides an interface for runtime inspection and modification

hikari/impl/event_manager_base.py
davfsa: Ok, sounds and looks good

hikari/impl/event_manager_base.py
```diff
@@ -71,23 +74,51 @@ class EventManagerBase(event_dispatcher.EventDispatcher):
...
+        return []
+
+    def remove_raw_consumer(self, name: str, consumer: ConsumerT, /) -> None:
+        name = name.casefold()
+
+        try:
+            self._consumers[name].remove(consumer)
+
+        except ValueError:
+            raise LookupError(name, consumer) from None
```
FasterSpeeding, line 110: ValueError?

FasterSpeeding:
Just a few concepts i'm not too sure on here
* While the event dispatcher abc is still an event dispatcher, it kind of feels like based on it's state in this PR it might make more sense to rename it to a event manager or something along those lines
* Should the event manager impl disallow removing builtin raw consumers
* And i'm not too sure if the raw consumers would make more sense being called "raw listeners" or something along those lines for the sake of consistency or not

hikari/api/event_dispatcher.py
```diff
@@ -49,6 +54,75 @@ class EventDispatcher(abc.ABC):
...
+        Raises
+        ------
+        LookupError
+            If the either `name` or `consumer` couldn't be found in the
+            registered raw consumers.
+        """  # noqa: E501 - Line too long
+
+    def consume_raw_event(
+        self, shard: gateway_shard.GatewayShard, event_name: str, payload: data_binding.JSONObject
+    ) -> None:
```
davfsa, line 107:
```suggestion

   @abc.abstractmethod
    def consume_raw_event(
        self, shard: gateway_shard.GatewayShard, event_name: str, payload: data_binding.JSONObject
    ) -> None:
```

hikari/impl/shard.py
```diff
@@ -666,7 +672,11 @@ def _dispatch(self, name: str, seq: int, data: data_binding.JSONObject) -> None:
             self._logger.info("shard has resumed [session:%s, seq:%s]", self._session_id, self._seq)
             self._handshake_completed.set()
 
-        self._event_consumer(self, name, data)
+        try:
+            self._dispatcher.consume_raw_event(self, name, data)
+
+        except LookupError:
+            self._logger.debug("ignoring unknown event %s:\n    %r", name, data)
```
davfsa, line 679: Unknown events should be handled in `consume_raw_event`

hikari/impl/shard.py
```diff
@@ -851,7 +861,10 @@ async def _run_once(self) -> bool:
 
         try:
             # Dispatch CONNECTED synthetic event.
-            self._event_consumer(self, "CONNECTED", {})
+            asyncio.create_task(
+                self._dispatcher.dispatch(self._event_factory.deserialize_connected_event(self)),
+                name=f"shard {self._shard_id} CONNECTED event dispatch",
+            )
```
davfsa, line 867: The the task needed here?

hikari/impl/shard.py
```diff
@@ -921,7 +934,10 @@ async def _run_once(self) -> bool:
             if dispatch_disconnect:
                 # If we managed to connect, we must always send the DISCONNECT event
                 # afterwards.
-                self._event_consumer(self, "DISCONNECTED", {})
+                asyncio.create_task(
+                    self._dispatcher.dispatch(self._event_factory.deserialize_disconnected_event(self)),
+                    name=f"shard {self._shard_id} DISCONNECTED event dispatch",
+                )
```
davfsa, line 940: Same here

FasterSpeeding:
> Just a few concepts i'm not too sure on here
> 
> * While the event dispatcher abc is still an event dispatcher, it kind of feels like based on it's state in this PR it might make more sense to rename it to a event manager or something along those lines
> * Should the event manager impl disallow removing builtin raw consumers
> * And i'm not too sure if the raw consumers would make more sense being called "raw listeners" or something along those lines for the sake of consistency or not

@davfsa any comment for any of these?

hikari/impl/shard.py
FasterSpeeding: Would that mean adding an unknown event handler or just having consume_raw_event be a black box where you can't tell if it actually dispatched an event or just ignore it at runtime after calling it without also calling get_raw_consumers afterwards?

davfsa: Hmm, true, good point

davfsa:
> While the event dispatcher abc is still an event dispatcher, it kind of feels like based on it's state in this PR it might make more sense to rename it to a event manager or something along those lines

Yea, event manager makes more sense

> Should the event manager impl disallow removing builtin raw consumers

If the user does it, they have their reasons

> And i'm not too sure if the raw consumers would make more sense being called "raw listeners" or something along those lines for the sake of consistency or not

Yea, maybe just rename them to `consumers`.

Technical specifications:
- `consumer`: takes in json and shard object and return an event
- `listener`: takes in event and does not return anything

hikari/impl/shard.py
FasterSpeeding: Also random point but logging it here has the added bonus of telling you which shard the payload was received on

FasterSpeeding:
Renamed the event dispatcher interface to event manager and changed other relevant names 👍

davfsa:
While you are at it, could you also do a part of #431

FasterSpeeding:
> While you are at it, could you also do a part of #431

i'll keep that to another PR

hikari/traits.py
```diff
@@ -124,23 +124,22 @@ def proxy_settings(self) -> config.ProxySettings:
 
 
 @typing.runtime_checkable
-class DispatcherAware(typing.Protocol):
-    """Structural supertype for a dispatcher-aware object.
+class EventAware(typing.Protocol):
```
davfsa, line 127: `EventManagerAware`?

FasterSpeeding: seems a bit redundant but if you prefer that name then sure

FasterSpeeding: On the other hand it may also make sense to add methods like "subscribe" here like we do on ShardAware but that wouldn't be necessary if we're renaming it to EventManagerAware

FasterSpeeding:
Bump

davfsa:
> Bump

Thought I approved it :sweat_smile:

davfsa (approved):

PR #425, opened 2021-01-12
Feature/traits fix

### LINKED ISSUE #291: Stateless bot should be its own class.

### LINKED ISSUE #368: Cache configuration

davfsa:
### Checklist

### Related issues
Closes #368 and closes #291


hikari/_about.py
```diff
@@ -36,5 +36,5 @@
 __issue_tracker__ = "https://github.com/hikari-py/hikari/issues"
 __license__ = "MIT"
 __url__ = "https://github.com/hikari-py/hikari"
-__version__ = "2.0.0"
+__version__ = "0.0.0"
```
FasterSpeeding, line 39: what's with this ver change?

hikari/events/guild_events.py
```diff
@@ -632,17 +635,6 @@ def guild_id(self) -> snowflakes.Snowflake:
         """
         return self.presence.guild_id
 
-    # TODO: make this nicer, as it is inconsistent with stuff elsewhere I guess.
-    def get_cached_user(self) -> typing.Optional[users.User]:
```
FasterSpeeding, line 636: If we're removing this shouldn't `fetch_user` also be removed? (plus wouldn't a rename to `get_user` also work?

hikari/events/member_events.py
```diff
@@ -98,6 +98,9 @@ def guild(self) -> typing.Optional[guilds.GatewayGuild]:
             The guild that this event occurred in, if known, else
             `builtins.None`.
         """
+        if not isinstance(self.app, traits.CacheAware):
+            return None
```
FasterSpeeding, line 102: why not just 
```py
if isinstance(self.app, traits.CacheAware):
    return elf.app.cache.get_available_guild(self.guild_id) or self.app.cache.get_unavailable_guild(self.guild_id)
```
and let it implicitly return None otherwise for these 1 liners?

hikari/events/message_events.py
```diff
@@ -598,10 +610,11 @@ def channel(self) -> typing.Union[None, channels.GuildTextChannel, channels.Guil
             if it is a normal message, or `hikari.channels.GuildNewsChannel` if
             sent in an announcement channel.
         """
+        if not isinstance(self.app, traits.CacheAware):
+            return None
+
         channel = self.app.cache.get_guild_channel(self.channel_id)
-        assert channel is None or isinstance(
```
FasterSpeeding, line 602: what if the channel isn't cached?

hikari/events/guild_events.py
davfsa: Yea, could rename it

hikari/_about.py
davfsa: I forgot to open a pr that fixes this. I was playing around with a version that we dont need to change and can stay constant between all versions. This would also help us separate between official versions and non-official ones.

hikari/events/member_events.py
davfsa: Mypy doesnt allow functions that say that return a union of something and `None` to return `None` in an indirect way.

hikari/_about.py
davfsa: Opened #426 for this

FasterSpeeding (approved):

PR #430, opened 2021-01-16
Allow the event manager to run stateless-ly

FasterSpeeding:
### Checklist

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -262,7 +262,7 @@ class BotApp(traits.BotAware, event_dispatcher.EventDispatcher):
         self._cache = cache_impl.CacheImpl(self, cache_settings)
 
         # Event handling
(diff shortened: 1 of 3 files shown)


davfsa:
Not really sure about this one, as the cache will never be `None`.

FasterSpeeding:
> Not really sure about this one, as the cache will never be `None`.

Which assumes it will always be a state aware bot which a isn't how our interface is set and b wouldn't apply to something like a rest only impl which might still want to use dispatch

davfsa:
> Which assumes it will always be a state aware bot which a isn't how our interface is set and b wouldn't apply to something like a rest only impl which might still want to use dispatch

True, yea, then fine by me

davfsa (approved):

davfsa (approved):

PR #436, opened 2021-01-17
Slash commands

### LINKED ISSUE #413: Implement slash commands
https://discord.com/developers/docs/interactions/slash-commands

https://github.com/discord/discord-api-docs/commit/762da2910e678bffc4d2c2fe13499bbb9d9927f5

- [x] Events
- [x] Endpoints
- [x] Objects

FasterSpeeding:
### Checklist

### Related issues

Close #413


hikari/api/interaction_server.py
```diff
@@ -0,0 +1,186 @@
+# -*- coding: utf-8 -*-
+# cython: language_level=3
+# Copyright (c) 2020 Nekokatt
```
davfsa, line 3: Copyright is wrong

hikari/api/interaction_server.py
```diff
@@ -0,0 +1,186 @@
...
+
+!!! note
+    For the standard implementation of
+    `hikari.api.special_endpoints.InteractionResponseBuilder` see
+    `hikari.impl.special_endpoints.InteractionResponseBuilder`
+"""
+
+
+class Response(typing.Protocol):
+    """Protocol of the data returned by `InteractionServer.on_interaction`.
```
davfsa, line 55: Should we rename this and make it an abc? You seem to be using abc.abstractmethod, so im going to guess its meant to be that

hikari/api/special_endpoints.py
```diff
@@ -417,3 +422,142 @@ def add_voice_channel(
...
+        Parameters
+        ----------
+        entity_factory : hikari.api.entity_factory.EntityFactory
+            The entity factory to use to serialize entities within this builder.
+
+        Returns
+        -------
+        hikari.internal.data_binding.JSONObject
+            The built json object representation of this builder.
+        """
```
davfsa, line 563: This doesnt need to be a special endpoint, it can just be a normal one which takes in `embed` and `embeds`

hikari/events/interaction_events.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+# cython: language_level=3
+# Copyright (c) 2020 Nekokatt
```
davfsa, line 3: Copyright

hikari/impl/entity_factory.py
```diff
@@ -232,6 +234,14 @@ def __init__(self, app: traits.RESTAware) -> None:
             channel_models.ChannelType.GUILD_STORE: self.deserialize_guild_store_channel,
             channel_models.ChannelType.GUILD_VOICE: self.deserialize_guild_voice_channel,
         }
+        self._interaction_type_mapping: typing.Mapping[
+            int,
+            typing.Callable[[data_binding.JSONObject], interaction_models.PartialInteraction],
+        ] = {
+            # PING
+            1: self._deserialize_partial_interaction,
```
davfsa, line 242: This should be in the InteractionType enum

hikari/impl/entity_factory.py
```diff
@@ -1729,18 +1880,24 @@ def deserialize_message(  # noqa CFQ001 - Function too long
...
             channels = {c.id: c for c in map(self.deserialize_partial_channel, raw_channels)}
 
-        users: typing.Mapping[snowflakes.Snowflake, user_models.User] = {}
+        else:
+            channels = {}
+
         if raw_users := payload.get("mentions"):
             users = {u.id: u for u in map(self.deserialize_user, raw_users)}
 
-        role_ids: typing.Sequence[snowflakes.Snowflake] = []
```
davfsa, line 1740: Why remove the typehints? They allow us to make sure that what is getting added to is what we expect

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,450 @@
+# -*- coding: utf-8 -*-
+# cython: language_level=3
+# Copyright (c) 2020 Nekokatt
```
davfsa, line 3: Copyright

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,450 @@
...
+            _LOGGER.info("Starting site on %s", site.name)
+            await site.start()
+
+    def set_listener(
+        self, listener: typing.Optional[interaction_server.MainListenerT], /, *, replace: bool = False
+    ) -> None:
+        if not replace and self._listener:
+            raise TypeError("Main listener already set")
+
+        self._listener = listener
```
davfsa, line 450: Im not sure this is the correct way to do this, and there are some stuff in here that I don't really like. I'm going to have a think what the best way to implement this is and get back to you

hikari/impl/rest.py
```diff
@@ -717,39 +718,6 @@ async def _parse_ratelimits(self, compiled_route: routes.CompiledRoute, response
...
-
-        if role_mentions is True:
-            parsed_mentions.append("roles")
-        elif isinstance(role_mentions, typing.Collection):
-            # Duplicates will cause discord to error.
-            ids = {str(int(r)) for r in role_mentions}
-            allowed_mentions["roles"] = list(ids)
-
-        return allowed_mentions
-
```
davfsa, line 752: This makes no sense to be moved to `special_endpoints`, but rather to a `internal` class

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,364 @@
+# -*- coding: utf-8 -*-
+# cython: language_level=3
+# Copyright (c) 2020 Nekokatt
```
davfsa, line 3: Copyright

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,364 @@
...
+            ssl_context=ssl_context,
+            backlog=backlog,
+            reuse_address=reuse_address,
+            reuse_port=reuse_port,
+        )
+
+    def set_listener(
+        self, listener: typing.Optional[interaction_server_.MainListenerT], /, *, replace: bool = False
+    ) -> None:
+        self._server.set_listener(listener, replace=replace)
```
davfsa, line 364: Also having a think about this one

hikari/interactions.py
```diff
@@ -0,0 +1,190 @@
+# -*- coding: utf-8 -*-
+# cython: language_level=3
+# Copyright (c) 2020 Nekokatt
```
davfsa, line 3: Copyright

hikari/interactions.py
```diff
@@ -0,0 +1,190 @@
...
+    BOOL = BOOLEAN
+    USER = 6
+    CHANNEL = 7
+    ROLE = 8
+
+
+@typing.final
+class InteractionType(int, enums.Enum):
+    # PING isn't here as it should be handled as internal detail of the REST
+    # server rather than as a part of the public interface.
```
davfsa, line 75: Document it as being a internal, but have it in this enum

hikari/interactions.py
```diff
@@ -0,0 +1,190 @@
+# -*- coding: utf-8 -*-
```
davfsa, line 1: A lot of documentation is missing in this file

hikari/internal/ed25519.py
```diff
@@ -0,0 +1,116 @@
+# -*- coding: utf-8 -*-
+# cython: language_level=3
+# Copyright (c) 2020 Nekokatt
```
davfsa, line 3: Copyright

hikari/internal/ed25519.py
```diff
@@ -0,0 +1,116 @@
...
+                self._public_key.verify(signature, timestamp + body)
+                return True
+
+            except _ed25519.BadSignatureError:
+                return False
+
+    FastED25519Key = _FastED25519Key
+
+except ImportError:
+    FastED25519Key = None
```
davfsa, line 112: This is not the way to do this. There should be a `ED25519Key` class. In the top of the file, you try to import `ed25519`, if it fails, fallback to `pure25519` in the `ED25519Key` class

davfsa (changes requested):

hikari/internal/ed25519.py
FasterSpeeding: And if we're using that implicit approach how would we test that we're using both implementations properly

davfsa: For what I can see, they seem to do the same thing:
```py
try:
    import ed25519 as _ed25519
    LIB = _ed25519
except ImportError:
    from pure25519 import ed25519_oop as _pure_ed25519
    LIB = _pure_ed25519
    
class ED25519Key:
    """ED25519 key class used for verifying payloads.

    Parameters
    ----------
    public_key : builtins.bytes
        The byte public key this is being initiated for.
    """

    __slots__: typing.Sequence[str] = ("_public_key",)

    def __init__(self, public_key: bytes, /) -> None:
        self._public_key = LIB.VerifyingKey(public_key)

    def verify(self, body: bytes, timestamp: bytes, signature: bytes) -> bool:
        """Verify a interaction payload received from Discord against this public key.

        Parameters
        ----------
        body : bytes
            The interaction payload.
        signature : bytes
            Value of the `"X-Signature-Ed25519"` header.
        timestamp : bytes
            Value of the `"X-Signature-Timestamp"` header.

        Returns
        -------
        bool
            Whether the provided arguments match this public key.
        """
            try:
                self._public_key.verify(signature, timestamp + body)
                return True

            except lib.BadSignatureError:
                return False

Comment (FasterSpeeding) [reply hikari/internal/ed25519.py:112]: Still how would we cover both cases with tests, the case of lib a and the case of lib b with a conditional if available check before running

File: hikari/interactions.py (hunk 13 above)
Comment (FasterSpeeding) [reply hikari/interactions.py:75]: And then presumably also make two union types that look like
```py
InteractionTypeT = typing.Literal[InteractionType[APPLICATION_COMMAND]]
```
and
```py
InteractionResponseTypeT = typing.Union[
    typing.Literal[InteractionResponseType.ACKNOWLEDGE],
    typing.Literal[InteractionResponseType.CHANNEL_MESSAGE],
    typing.Literal[InteractionResponseType.CHANNEL_MESSAGE_WITH_SOURCE],
    typing.Literal[InteractionResponseType.ACKNOWLEDGE_WITH_SOURCE],
]
```
to match that behaviour?

hikari/internal/ed25519.py
davfsa: You can assume `pure25519` will always be there. Then, in the tests, you can unimport them, mock the imports and see what `LIB` (temp var, dont name it that, pls) is set to

FasterSpeeding: > You can assume `pure25519` will always be there. Then, in the tests, you can unimport them, mock the imports and see what `LIB` (temp var, dont name it that, pls) is set to

That's not an integration test that makes sure we're using these different libs properly though, that just tests the behaviour around conditional imports

davfsa: Both libs use a use `VerifyingKey`, `VerifyingKey.verify` and `BadSignatureError`. There is no difference between them. You can also write down some integration tests if you want that mock the imports for both cases. Then you can write tests per lib

FasterSpeeding: Is there even a way to mock imports and force them to fail though

davfsa:
```py
import mock
import builtins

with mock.patch.object(builtins, "__import__", new=mock.Mock(side_effect=RuntimeError)):
    import test
```

```
  File "/home/davfsa/coding/test-bots/test5.py", line 5, in <module>
    import test
  File "/home/davfsa/coding/hikari/.venv/lib/python3.8/site-packages/mock/mock.py", line 1100, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/home/davfsa/coding/hikari/.venv/lib/python3.8/site-packages/mock/mock.py", line 1104, in _mock_call
    return _mock_self._execute_mock_call(*args, **kwargs)
  File "/home/davfsa/coding/hikari/.venv/lib/python3.8/site-packages/mock/mock.py", line 1161, in _execute_mock_call
    raise effect
RuntimeError
```

FasterSpeeding: except you can't target specific imports that way so you're going to have to rely on order and import every other import the module uses which might fuck with future imports before you eh so that you can provide every other import in the mock patch's side effects which sounds excessive, jank and fragile.

hikari/impl/entity_factory.py
FasterSpeeding: was removing the building a list/dict twice also wouldn't passing them to the class's init cover what you described since these are only touched in one place

hikari/internal/ed25519.py
davfsa:
```py
import mock
import builtins
import contextlib


@contextlib.contextmanager
def mock_import(name, error):
    original = builtins.__import__

    def side_effect(self, *args, **kwargs):
        module = mock.call_args[0][0]  # get the call
        if module == name:
            raise error
        else:
            side_effect.self = self
            return original(self, *args, **kwargs)

    patcher = mock.patch.object(builtins, "__import__", side_effect=side_effect)
    yield patcher.start()
    patcher.stop()


with mock_import("test", RuntimeError) as mock:
    import hikari

    print("hikari imported")
    print(hikari.BotApp)

    import test

    print("this will never get printed")
```
```
hikari imported
<class 'hikari.impl.bot.BotApp'>
  File "/home/davfsa/coding/test-bots/test5.py", line 29, in <module>
    import test
  File "/home/davfsa/coding/hikari/.venv/lib/python3.8/site-packages/mock/mock.py", line 1100, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/home/davfsa/coding/hikari/.venv/lib/python3.8/site-packages/mock/mock.py", line 1104, in _mock_call
    return _mock_self._execute_mock_call(*args, **kwargs)
  File "/home/davfsa/coding/hikari/.venv/lib/python3.8/site-packages/mock/mock.py", line 1167, in _execute_mock_call
    result = effect(*args, **kwargs)
  File "/home/davfsa/coding/test-bots/test5.py", line 13, in side_effect
    raise error
RuntimeError
```

FasterSpeeding: I still don't exactly see what the issue is with the original approach though, it's just the class equivalent of what we do in time where we define a standard implementation then try to overwrite that except without redefining a class which would be a mypy and flake8 violation and with an approach that's easier to test but if you don't want the individual classes to be accessible by themselves at runtime as internal detail (tbf this whole thing is internal) then we could just delete them or keep them private or just put a flake8 and mypy ignore on the redefine errors ig

hikari/api/special_endpoints.py
FasterSpeeding: except this isn't an endpoint you call, this is the result that's used as a response within a rest server (you are the endpoint)

davfsa: Well, for the methods that are here, this doesnt need to be a builder, it can be a normal call

hikari/internal/ed25519.py
davfsa: Its nicer and removes a lot of duplicate code

hikari/api/special_endpoints.py
FasterSpeeding: A call to/on what?

hikari/internal/ed25519.py
FasterSpeeding: How is it nicer if it forces you to do a ton of hacky code to test it properly?

hikari/impl/rest.py
FasterSpeeding: I mean i just moved it from one impl module to another but if that's what you say then whatever

hikari/impl/rest_bot.py
FasterSpeeding: correct way to do what specifically

hikari/impl/interaction_server.py
FasterSpeeding: correct way to do what specifically

hikari/api/rest.py
```diff
@@ -3857,7 +3865,7 @@ async def edit_member(
             snowflakes.SnowflakeishOr[channels.GuildVoiceChannel]
         ] = undefined.UNDEFINED,
         reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
-    ) -> None:
+    ) -> guilds.Member:
```
FasterSpeeding, line 3868: Remove from this pr

hikari/impl/entity_factory.py
```diff
@@ -1698,7 +1843,7 @@ def deserialize_message(  # noqa CFQ001 - Function too long
                 description=application_payload["description"],
                 icon_hash=application_payload["icon"],
                 summary=application_payload["summary"],
-                cover_image_hash=application_payload["cover_image"],
+                cover_image_hash=application_payload.get("cover_image"),
```
FasterSpeeding, line 1846: Remove from this pr

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,543 @@
...
+    )
+
+    def __init__(
+        self,
+        *,
+        entity_factory: entity_factory_.EntityFactory,
+        event_factory: event_factory_.EventFactory,
+        event_manager: typing.Optional[event_manager_.EventManager] = None,
+        public_key: str,
+        dumps: aiohttp.typedefs.JSONEncoder = json.dumps,
```
FasterSpeeding, line 162: data_binding json methods? (also might want to use these in shards)

FasterSpeeding:
TODO unknown (1)

FasterSpeeding:
`https://github.com/discord/discord-api-docs/commit/762da2910e678bffc4d2c2fe13499bbb9d9927f5`

FasterSpeeding:
@davfsa btw i've made quite a few changes to this and another quick review would be welcomed (although I do have 1 possible change planned which is letting them register interaction listeners with the interaction server specifically (not gateway) for a specific interaction type as to make the design more forwards compatible but to uphold the fact that this will decide the result only let them register 1 per type with partial interaction obviously not being allowed to be directly registered (if that makes any sense)

hikari/api/rest.py
```diff
@@ -1910,8 +1913,10 @@ async def delete_webhook(
     @abc.abstractmethod
     async def execute_webhook(
         self,
-        webhook: snowflakes.SnowflakeishOr[webhooks.Webhook],
+        # TODO: update docs
+        webhook: typing.Union[snowflakes.Snowflakeish, webhooks.Webhook, guilds.PartialApplication],
```
FasterSpeeding, line 1917: Executable webhook instead?

davfsa: Nah, that can stay as `webhook`. All webhooks are executable :)

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,597 @@
...
+
+        result: typing.Optional[special_endpoints.InteractionResponseBuilder] = None
+        if self._listener:
+            try:
+                result = await self._listener(event)
+
+            except Exception as exc:
+                asyncio.get_running_loop().call_exception_handler(
+                    {"message": "Exception occurred during interaction dispatch", "exception": exc}
+                )
```
davfsa, line 368: Cant we just log this normally? And if you do keep it like this, please use the loop first and then call `call_exception_handler` for redability

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,597 @@
...
+                return _Response(_INTERNAL_SERVER_ERROR_STATUS, "Exception occurred during interaction dispatch")
+
+        payload = self._dumps(result.build(self._entity_factory)) if result else _ACK_PAYLOAD
```
davfsa, line 371: Why assume that if its not a result its an ack?

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,597 @@
...
+            loop.set_debug(True)
+
+        if coroutine_tracking_depth is not None:
+            try:
+                # Provisionally defined in CPython, may be removed without notice.
+                sys.set_coroutine_origin_tracking_depth(coroutine_tracking_depth)  # type: ignore[attr-defined]
+            except AttributeError:
+                _LOGGER.log(
+                    ux.TRACE, "cannot set coroutine tracking depth for %s, no functionality exists for this", loop
+                )
```
davfsa, line 459:
```suggestion
                _LOGGER.log(ux.TRACE, "cannot set coroutine tracking depth for sys, no functionality exists for this")
```

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,597 @@
...
+        try:
+            event = self._event_factory.deserialize_interaction_create_event(
+                None, payload, application_id=self._application_id
+            )
+
+        except Exception as exc:
+            asyncio.get_running_loop().call_exception_handler(
+                {"message": "Exception occurred during interaction deserialization", "exception": exc}
+            )
+            return _Response(_INTERNAL_SERVER_ERROR_STATUS, "Exception occurred during interaction deserialization")
```
davfsa, line 351: Why not just a default logging?

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,597 @@
...
+        try:
+            payload = self._loads(body.decode("utf-8"))
+            interaction_type = int(payload["type"])
+
+        except (data_binding.JSONDecodeError, ValueError) as exc:
+            _LOGGER.info("Received a request with an invalid JSON body", exc_info=exc)
+            return _Response(_BAD_REQUEST_STATUS, "Invalid JSON body")
+
+        except (KeyError, TypeError) as exc:
+            _LOGGER.info("Invalid or missing 'type' field in received JSON payload", exc_info=exc)
```
davfsa, line 335: `_LOGGER.error` for these?

hikari/api/rest.py
FasterSpeeding: Executable Webhook is mixin for any entity that can be executed like a webhook (webhook and command interactions)

hikari/api/rest.py
```diff
@@ -5551,3 +5559,402 @@ async def sync_guild_template(
...
+            guild if `guild` is provided or the global commands if not.
+        """
+
+    @abc.abstractmethod
+    async def set_application_commands(
+        self,
+        commands: typing.Sequence[special_endpoints.CommandBuilder],
+        guild: undefined.UndefinedOr[snowflakes.SnowflakeishOr[guilds.PartialGuild]] = undefined.UNDEFINED,
+    ) -> typing.Sequence[interactions.Command]:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
FasterSpeeding: The docs aren't finished anyway

hikari/api/rest.py
davfsa: A command interaction is basically a webhook. Tbh we can treat them as the same object

hikari/impl/special_endpoints.py
```diff
@@ -26,7 +26,11 @@
 """
 from __future__ import annotations
 
-__all__: typing.List[str] = ["TypingIndicator", "GuildBuilder"]
+__all__: typing.List[str] = [
+    "TypingIndicator",
+    "GuildBuilder",
+    "InteractionResponseBuilder",
```
FasterSpeeding, line 32: missing all entry

hikari/applications.py
```diff
@@ -58,6 +62,17 @@
     from hikari import traits
 
 
+@typing.final
+class ApplicationFlags(enums.Flag):
```
FasterSpeeding, line 66: Remove from this pr

hikari/messages.py
```diff
@@ -107,9 +107,18 @@ class MessageType(int, enums.Enum):
     CHANNEL_FOLLOW_ADD = 12
     """Channel follow add."""
 
+    GUILD_DISCOVERY_DISQUALIFIED = 14
+    """A message to indicate that a guild has been disqualified from discovery."""
+
```
FasterSpeeding, line 112: Remove from this pr

hikari/applications.py
```diff
@@ -98,6 +113,9 @@ class OAuth2Scope(str, enums.Enum):
         You must be whitelisted to use this scope.
     """
 
+    APPLICATIONS_COMMANDS_UPDATE = "applications.commands.update"
+    """Allows your app to update slash commands via this bearer token."""
```
FasterSpeeding, line 117: remove from this pr

hikari/webhooks.py
```diff
@@ -485,6 +429,96 @@ async def delete_message(self, message: snowflakes.SnowflakeishOr[messages_.Mess
 
         await self.app.rest.delete_webhook_message(self.id, token=self.token, message=message)
 
+
+@attr_extensions.with_copy
+@attr.s(eq=True, hash=True, init=True, kw_only=True, slots=True, weakref_slot=False)
+class Webhook(ExecutableWebhook):
```
FasterSpeeding, line 435: TODO: check nothing was lost here during the rebase

hikari/internal/mentions.py
```diff
@@ -0,0 +1,92 @@
...
+    """Generate an allowed mentions JSON object.
+
+    Parameters
+    ----------
+    mentions_everyone : hikari.undefined.UndefinedOr[builtins.bool]
+        Whether @everyone and @here mentions are enabled. If
+        `hikari.undefined.UNDEFINED` or `False` then this will be disabled.
+    mentions_reply : hikari.undefined.UndefinedOr[builtins.bool]
+        Whether the reply mention should be enabled. If `hikari.undefined.UNDEFINED`
+        or `False` then this will be disabled.
```
FasterSpeeding, line 54: TODO: check nothing was lost here during the rebase

hikari/impl/entity_factory.py
```diff
@@ -1558,6 +1577,159 @@ def deserialize_invite_with_metadata(self, payload: data_binding.JSONObject) ->
...
+    def deserialize_interaction(
+        self, payload: data_binding.JSONObject, *, application_id: snowflakes.Snowflake
+    ) -> interaction_models.PartialInteraction:
+        interaction_type = interaction_models.InteractionType(payload["type"])
+
+        try:
+            deserialize = self._interaction_type_mapping[interaction_type]
+
+        except KeyError:
+            return interaction_models.PartialInteraction(
```
FasterSpeeding, line 1706: TODO: update behaviour here

hikari/api/interaction_server.py
```diff
@@ -0,0 +1,222 @@
...
+        """Start the REST server then return."""
+
+    # @typing.overload
+    # def get_listener(
+    #     self, interaction_type: typing.Type[interactions.CommandInteraction]
+    # ) -> typing.Optional[ListenerT[interactions.CommandInteraction, special_endpoints.CommandResponseBuilder]]:
+    #     raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_listener(
```
FasterSpeeding, line 174: /

FasterSpeeding:
Ability to go from command object to a populated builder or just accept pure command objects as well on bulk edit commands?

PR #439, opened 2021-01-19
Implement Sticker model

### LINKED ISSUE #351: Discord Stickers
https://github.com/discord/discord-api-docs/commit/b9ffd52b09f5fdd41751947fb6b5cd0d61f1fd76

FasterSpeeding:
### Summary
Implement Sticker model

### Checklist

### Related issues

Close #351

### DIFF
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -1451,6 +1451,70 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
     # MESSAGE MODELS #
     ##################
 
+    def _deserialize_message_activity(self, payload: data_binding.JSONObject) -> message_models.MessageActivity:
+        return message_models.MessageActivity(
+            type=message_models.MessageActivityType(payload["type"]), party_id=payload.get("party_id")
+        )
+
+    def _deserialize_message_application(self, payload: data_binding.JSONObject) -> message_models.MessageApplication:
+        primary_sku_id: typing.Optional[snowflakes.Snowflake] = None
+        if (raw_primary_sku_id := payload.get("primary_sku_id")) is not None:
+            primary_sku_id = snowflakes.Snowflake(raw_primary_sku_id)
+
+        return message_models.MessageApplication(
+            id=snowflakes.Snowflake(payload["id"]),
+            name=payload["name"],
+            description=payload["description"],
+            icon_hash=payload["icon"],
+            summary=payload["summary"],
+            cover_image_hash=payload["cover_image"],
+            primary_sku_id=primary_sku_id,
+        )
+
+    def _deserialize_message_attachment(self, payload: data_binding.JSONObject) -> message_models.Attachment:
+        return message_models.Attachment(
+            id=snowflakes.Snowflake(payload["id"]),
+            filename=payload["filename"],
+            size=int(payload["size"]),
+            url=payload["url"],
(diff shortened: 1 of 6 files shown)


hikari/impl/entity_factory.py
```diff
@@ -1451,6 +1451,70 @@ def deserialize_invite_with_metadata(self, payload: data_binding.JSONObject) ->
...
+            id=snowflakes.Snowflake(payload["id"]),
+            pack_id=snowflakes.Snowflake(payload["pack_id"]),
+            name=payload["name"],
+            description=payload["description"],
+            tags=[tag.strip() for tag in payload["tags"].split(",")] if "tags" in payload else [],
+            asset_hash=payload["asset"],
+            preview_asset_hash=payload["preview_asset"],
+            format_type=message_models.StickerFormatType(payload["format_type"]),
+        )
+
```
davfsa, line 1517: These should be public for consistency sake

hikari/impl/entity_factory.py
```diff
@@ -1580,6 +1598,7 @@ def deserialize_partial_message(  # noqa CFQ001 - Function too long
             message_reference=message_reference,
             referenced_message=referenced_message,
             flags=message_models.MessageFlag(payload["flags"]) if "flags" in payload else undefined.UNDEFINED,
+            stickers=stickers,
```
davfsa, line 1601: Stickers are not in the object

hikari/internal/cache.py
```diff
@@ -619,7 +619,7 @@ class MentionsData(BaseData[messages.Mentions]):
     """A model for storing message mentions data in an in-memory cache."""
 
     users: undefined.UndefinedOr[typing.Mapping[snowflakes.Snowflake, RefCell[users_.User]]] = attr.ib()
-    role_ids: undefined.UndefinedOr[typing.Sequence[snowflakes.Snowflake]] = attr.ib()
+    role_ids: undefined.UndefinedOr[typing.Tuple[snowflakes.Snowflake, ...]] = attr.ib()
```
davfsa, line 622: Why all these changes to Tuples? They should be Sequences

davfsa (changes requested):

hikari/internal/cache.py
FasterSpeeding: Because they are specifically tuples, putting sequences for these few was a mistake since that allows for a mutable list to be used but in the cache but we want them to be specifically immutable

hikari/impl/entity_factory.py
FasterSpeeding: as in the json object or the marshalled object?

davfsa: `PartialMessage` in this case

davfsa: nvm, for some reason github didnt show me that change

hikari/impl/entity_factory.py
FasterSpeeding: The approach this marshaller was built on was to only make deserthey ialize methods public if are specifically called from somewhere rather than only internally within the factory rather than keeping it 1-to-1 relationship for models to deserialize/serialize methods but this didn't stop the use of private methods for avoiding duplication.
And the only place stickers are documented is on message objects right now

davfsa (approved):

PR #442, opened 2021-01-20
Revert raw event consumer system and traits optimization

davfsa:
- Remove unused args from slots
- Cleanup code
- Remove unused variables
- Fix documentation links

```
isinstance(long, BasicPyProtocol) 10.502764225999272 µs
isinstance(long, BasicHikariProtocol) 0.8962424719993578 µs
isinstance(failfast, BasicPyProtocol) 19.83291287300017 µs
isinstance(failfast, BasicHikariProtocol) 0.3696172330000991 µs
issubclass(long, BasicPyProtocol) 0.3248153929998807 µs
issubclass(long, BasicHikariProtocol) 0.30592563200116274 µs
issubclass(failfast, BasicPyProtocol) 0.3142782159993658 µs
issubclass(failfast, BasicHikariProtocol) 0.29908469400106696 µs
```

### Checklist

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -45,7 +45,6 @@ from hikari import intents as intents_
 from hikari import presences
 from hikari import traits
 from hikari import undefined
-from hikari.api import event_manager as event_manager_
 from hikari.impl import cache as cache_impl
 from hikari.impl import entity_factory as entity_factory_impl
 from hikari.impl import event_factory as event_factory_impl
@@ -65,15 +64,15 @@ if typing.TYPE_CHECKING:
     from hikari.api import cache as cache_
     from hikari.api import entity_factory as entity_factory_
     from hikari.api import event_factory as event_factory_
+    from hikari.api import event_manager as event_manager_
     from hikari.api import rest as rest_
     from hikari.api import shard as gateway_shard
     from hikari.api import voice as voice_
-    from hikari.internal import data_binding
 
 _LOGGER: typing.Final[logging.Logger] = logging.getLogger("hikari")
 
 
-class BotApp(traits.BotAware, event_manager_.EventManager):
+class BotApp(traits.BotAware):
     """Basic auto-sharding bot implementation.
 
     This is the class you will want to use to start, control, and build a bot
@@ -213,7 +212,7 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
         "_closing_event",
         "_closed",
         "_entity_factory",
-        "_events",
+        "_event_manager",
         "_event_factory",
         "_executor",
         "_http_settings",
@@ -262,7 +261,7 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
         self._cache = cache_impl.CacheImpl(self, cache_settings)
 
         # Event handling
-        self._events = event_manager_impl.EventManagerImpl(self, cache=self._cache)
+        self._event_manager = event_manager_impl.EventManagerImpl(self, cache=self._cache)
 
         # Entity creation
         self._entity_factory = entity_factory_impl.EntityFactoryImpl(self)
@@ -297,7 +296,7 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
 
     @property
     def event_manager(self) -> event_manager_.EventManager:
-        return self._events
+        return self._event_manager
 
     @property
     def entity_factory(self) -> entity_factory_.EntityFactory:
@@ -354,19 +353,9 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
     def is_alive(self) -> bool:
         return self._is_alive
 
-    def add_raw_consumer(self, name: str, consumer: event_manager_.ConsumerT, /) -> None:
-        self._events.add_raw_consumer(name, consumer)
-
-    def get_raw_consumers(self, name: str, /) -> typing.Sequence[event_manager_.ConsumerT]:
-        return self._events.get_raw_consumers(name)
-
-    def remove_raw_consumer(self, name: str, consumer: event_manager_.ConsumerT, /) -> None:
-        self._events.remove_raw_consumer(name, consumer)
-
-    def consume_raw_event(
-        self, shard: gateway_shard.GatewayShard, event_name: str, payload: data_binding.JSONObject
-    ) -> None:
-        self._events.consume_raw_event(shard, event_name, payload)
+    def _check_if_alive(self) -> None:
+        if not self._is_alive:
+            raise errors.ComponentNotRunningError("bot is not running so it cannot be interacted with")
 
     async def close(self, force: bool = True) -> None:
         """Kill the application by shutting all components down."""
@@ -396,7 +385,7 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
                     }
                 )
 
-        await self.dispatch(self.event_factory.deserialize_stopping_event())
+        await self._event_manager.dispatch(self._event_factory.deserialize_stopping_event())
 
         _LOGGER.log(ux.TRACE, "StoppingEvent dispatch completed, now beginning termination")
 
@@ -417,17 +406,19 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
         self._shards.clear()
         self._is_alive = False
 
-        await self.dispatch(self.event_factory.deserialize_stopped_event())
+        await self._event_manager.dispatch(self._event_factory.deserialize_stopped_event())
 
     def dispatch(self, event: event_manager_.EventT_inv) -> asyncio.Future[typing.Any]:
-        return self._events.dispatch(event)
+        return self._event_manager.dispatch(event)
 
     def get_listeners(
         self, event_type: typing.Type[event_manager_.EventT_co], *, polymorphic: bool = True
     ) -> typing.Collection[event_manager_.CallbackT[event_manager_.EventT_co]]:
-        return self._events.get_listeners(event_type, polymorphic=polymorphic)
+        return self._event_manager.get_listeners(event_type, polymorphic=polymorphic)
 
     async def join(self, until_close: bool = True) -> None:
+        self._check_if_alive()
+
         awaitables: typing.List[typing.Awaitable[typing.Any]] = [s.join() for s in self._shards.values()]
         if until_close:
             awaitables.append(self._closing_event.wait())
@@ -437,9 +428,10 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
     def listen(
         self, event_type: typing.Optional[typing.Type[event_manager_.EventT_co]] = None
     ) -> typing.Callable[
-        [event_manager_.CallbackT[event_manager_.EventT_co]], event_manager_.CallbackT[event_manager_.EventT_co]
+        [event_manager_.CallbackT[event_manager_.EventT_co]],
+        event_manager_.CallbackT[event_manager_.EventT_co],
     ]:
-        return self._events.listen(event_type)
+        return self._event_manager.listen(event_type)
 
     @staticmethod
     def print_banner(banner: typing.Optional[str], allow_color: bool, force_color: bool) -> None:
@@ -570,7 +562,15 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
         status : hikari.presences.Status
             The initial status to show for the user presence on startup.
             Defaults to `hikari.presences.Status.ONLINE`.
+
+        Raises
+        ------
+        builtins.RuntimeError
+            If bot is already running.
         """
+        if self._is_alive:
+            raise RuntimeError("bot is already running")
+
         if shard_ids is not None and shard_count is None:
             raise TypeError("'shard_ids' must be passed with 'shard_count'")
 
@@ -747,7 +747,7 @@ class BotApp(traits.BotAware, event_manager_.EventManager):
             )
 
         requirements_task = asyncio.create_task(self._rest.fetch_gateway_bot(), name="fetch gateway sharding settings")
-        await self.dispatch(self.event_factory.deserialize_starting_event())
+        await self._event_manager.dispatch(self._event_factory.deserialize_starting_event())
         requirements = await requirements_task
 
         if shard_count is None:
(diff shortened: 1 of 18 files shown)


FasterSpeeding:
If the solution to avoiding instance checks against traits and try except duck checking in model methods then shouldn't shard aware and voice aware also be added to rest aware?

FasterSpeeding:
Also doesn't this make the CacheAware trait useless as it can no longer be used to type something as needing a cache implementation aware client

davfsa:
> If this is the final and best solution for avoiding instance checks against traits and try except duck checking in model methods then shouldn't shard aware and voice aware also be added to rest aware?

I don't see the reason to add them.

FasterSpeeding:
> > If this is the final and best solution for avoiding instance checks against traits and try except duck checking in model methods then shouldn't shard aware and voice aware also be added to rest aware?
> 
> I don't see the reason to add them.

are you going to get rid of methods on entities that use shard aware if it's available and avoid adding voice methods to entities?

hikari/impl/event_manager_base.py
```diff
@@ -71,51 +71,22 @@ class EventManagerBase(event_manager.EventManager):
...
+    __slots__: typing.Sequence[str] = ("_app", "_listeners", "_waiters")
 
     def __init__(self, app: traits.BotAware) -> None:
         self._app = app
-        self._consumers: typing.Dict[str, typing.List[ConsumerT]] = {}
-        self._intents = app.intents
         self._listeners: ListenerMapT[base_events.Event] = {}
         self._waiters: WaiterMapT[base_events.Event] = {}
 
-        for name, member in inspect.getmembers(self):
```
FasterSpeeding, line 83: probably want to rename the abc for this back to event dispatcher

hikari/undefined.pyi
```diff
@@ -29,9 +29,7 @@ from typing import TypeVar as __TypeVar
 from typing import Union as __Union
 
 class UndefinedType(__enum.Enum):
-    def __bool__(self) -> __Literal[False]:
-        ...
-
+    def __bool__(self) -> __Literal[False]: ...
```
FasterSpeeding, line 32: should we not also be reformatting pyi files with black?

hikari/traits.py
```diff
@@ -60,42 +57,64 @@
...
-"""
+_Protocol = NotImplemented
 
-EventT_inv = typing.TypeVar("EventT_inv", bound="base_events.Event")
-"""Type-hint for an invariant event implementation instance.
 
-This will bind to the bound event type only. Subclasses and superclasses will
-not be matched.
-"""
+class _FastProtocol(abc.ABCMeta):
```
FasterSpeeding, line 66: Wouldn't it make sense to house this in a "internal" module?

hikari/impl/bot.py
```diff
@@ -396,7 +385,7 @@ async def handle(name: str, awaitable: typing.Awaitable[typing.Any]) -> None:
                     }
                 )
 
-        await self.dispatch(self.event_factory.deserialize_stopping_event())
+        await self._event_manager.dispatch(self._event_factory.deserialize_stopping_event())
```
FasterSpeeding, line 388: Isn't this inconsistent compared with how artificial events are dispatched in bot.py

FasterSpeeding: (also the event manager doesn't have managers for any artificial events anymore)

hikari/impl/shard.py
```diff
@@ -861,7 +856,7 @@ async def _run_once(self) -> bool:
 
         try:
             # Dispatch CONNECTED synthetic event.
-            self._event_manager.dispatch(self._event_factory.deserialize_connected_event(self)),
```
FasterSpeeding, line 864: Isn't this inconsistent compared with how artificial events are dispatched in bot.py
(also the event manager doesn't have managers for any artificial events anymore)

hikari/impl/event_manager_base.py
```diff
@@ -71,51 +71,22 @@ class EventManagerBase(event_manager.EventManager):
...
     def consume_raw_event(
-        self, shard: gateway_shard.GatewayShard, event_name: str, payload: data_binding.JSONObject
+        self, event_name: str, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject
     ) -> None:
-        callbacks = self._consumers[event_name.casefold()]
-        asyncio.gather(
-            *(self._handle_dispatch(callback, shard, payload) for callback in callbacks), return_exceptions=True
-        )
+        try:
+            callback = getattr(self, "on_" + event_name.lower())
```
FasterSpeeding, line 85: random point but loading these into a dict instead of calling getattr every dispatch was a small optimisation, not just a part of the multi-consumer

davfsa: Ik, I was thinking that it might be better this way as it allows to add more programmatically. But after thinking about it, im going to add the consumers dict back

hikari/traits.py
```diff
@@ -60,42 +57,64 @@
...
 
-This is not expected to return anything.
-"""
+            for base in bases:
+                if not issubclass(base, _Protocol):
+                    raise TypeError(f"Protocols can only inherit from other protocols, got {base!r}")
+                attributes += base._attributes_
 
+            namespace["_is_protocol"] = True
+            namespace["_attributes_"] = attributes
```
FasterSpeeding, line 96: What about using a set here to avoid duplicated entries?

hikari/traits.py
```diff
@@ -60,42 +57,64 @@
...
-@typing.runtime_checkable
-class NetworkSettingsAware(typing.Protocol):
+        return super().__new__(mcs, cls_name, bases, namespace)
+
+    def __instancecheck__(cls, obj: typing.Any) -> bool:
+        if not cls._is_protocol:
+            return super().__instancecheck__(obj)
+
+        for i in cls._attributes_:
+            if not hasattr(obj, i):
```
FasterSpeeding, line 105: how about a `return all(...)` statement here?

hikari/traits.py
```diff
@@ -60,42 +57,64 @@
...
 
-# Fixme: shouldn't this be using covariance instead of invariance?
-AsyncCallbackT = typing.Callable[[EventT_inv], typing.Coroutine[typing.Any, typing.Any, None]]
-"""Type-hint for an asynchronous coroutine function event callback.
+        if _Protocol not in bases:
+            namespace["_is_protocol"] = False
 
-This should consume a single argument: the event that was dispatched.
+        else:
+            attributes = tuple(name for name in namespace if not name.startswith("_"))
```
FasterSpeeding, line 88: While this approach is definitely a lot more efficient, this does seem to slightly violate how are checked against objects and types (since the reference impl does count private and protected attributes, would it be worth it making this better match the behaviour seen in the typing module by using typing.EXCLUDED_ATTRIBUTES to explicitly blacklist attributes (and also maybe checking against type hints as well)?

hikari/traits.py
FasterSpeeding, line 66: Would it be worth it also defining subclass check behaviour here?

hikari/traits.py
davfsa: `typing.Protocol` doesnt check for typehints anyways, so including it might break something downstream. And will change the attribute check :+1:

hikari/traits.py
davfsa: There is a reason I didn't go for that, and that is because this way we can quickly fail as soon as we see that one of the attributes is not there.

hikari/traits.py
FasterSpeeding: > typing.Protocol doesnt check for typehints anyways

it does check against a class's annotations for instance variables

```python
def _get_protocol_attrs(cls):
    """Collect protocol members from a protocol class objects.

    This includes names actually defined in the class dictionary, as well
    as names that appear in annotations. Special names (above) are skipped.
    """
    attrs = set()
    for base in cls.__mro__[:-1]:  # without object
        if base.__name__ in ('Protocol', 'Generic'):
            continue
        annotations = getattr(base, '__annotations__', {})
        for attr in list(base.__dict__.keys()) + list(annotations.keys()):
            if not attr.startswith('_abc_') and attr not in EXCLUDED_ATTRIBUTES:
                attrs.add(attr)
    return attrs
    ```

File: hikari/traits.py (hunk 8 above)
Comment (FasterSpeeding) [reply hikari/traits.py:105]: `all` returns on the first item in the iterator/iterable which isn't truthful (and `any` returns on the first which is truthful)
```py
>>> def divisible_by_4(i) -> bool:
...     print(i)
...     return i % 4 == 0
... print(all(divisible_by_4(i) for i in (4,8,12, 16, 17, 20, 24, 28)))
4
8
12
16
17
False
```

hikari/undefined.pyi
davfsa: It should already be targetting them

hikari/traits.py
davfsa: Seems to be slower, so will keep what I got rn :+1:

hikari/traits.py
davfsa: I would rather keep that to the default

hikari/impl/shard.py
davfsa: `bot.py` is able to dispatch artificial events due to it being the "owner". As shard is not the "owner", it is not able to do this and has to rely on the implementation of `bot.py`

hikari/traits.py
FasterSpeeding: Except wouldn't the default behaviour not match up with protocol subclass behaviour?

hikari/impl/shard.py
FasterSpeeding: So the shards are meant to just blindly dispatch an fake raw event which may or may not be dispatched based on implementation detail separate from them?

hikari/traits.py
davfsa: I have implemented all it seems to do into the creation of the class. If I'm missing something, let me know

FasterSpeeding: Except this won't have the subclass behaviour
```python
>>> import typing
>>> @typing.runtime_checkable
class FooProto(typing.Protocol):
	def foo():
		...

>>> class Foo:
	def foo():
		...

>>> issubclass(Foo, FooProto)
True
>>> @typing.runtime_checkable
class FooProto(typing.Protocol):
	bar: int
	def foo():
		    ...

		    
>>> class Foo:
	def bar():
		...
	def foo():
		...

		
>>> issubclass(Foo, FooProto)
  File "<pyshell#27>", line 1, in <module>
    issubclass(Foo, FooProto)
  File "C:\Python38\lib\abc.py", line 102, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
  File "C:\Python38\lib\typing.py", line 1081, in _proto_hook
    raise TypeError("Protocols with non-method members"
TypeError: Protocols with non-method members don't support issubclass()
```

hikari/impl/shard.py
FasterSpeeding: This feels less like a case of "ownership" and more like a case of ensuring the events are actually dispatched which you cannot do with the consume raw event method

hikari/internal/protocol.py
```diff
@@ -0,0 +1,118 @@
...
+
+        if not cls._is_protocol:
+            return
+
+        def _subclass_hook(other: type) -> bool:
+            # Since we are overwriting, we need to check this
+            if not isinstance(other, type):
+                raise TypeError("issubclass() arg 1 must be a class")
+
+            return any(hasattr(other, a) for a in cls._attributes_) or not cls._attributes_
```
FasterSpeeding, line 116: wouldn't this be all rather than any?

davfsa: No, cause i want it to return as soon as it find one that is the same

FasterSpeeding: That doesn't exactly sound like how protocol instance or subclass checking works, wouldn't they usually only pass if all the attributes are present?

davfsa: No, `issubclass` checks if one of the attrs or annotations is there, if so, its a subclass, if not, its not

As I wrote this i realized that if they are all not present, it checks the actual inheritance, so will add that

FasterSpeeding: Does it? Doesn't look like the behaviour i'm seeing and if that was the case it'd break a ton of stuff that was using it since having one attribute incommon doesn't mean it's compatible
```py
>>> import typing
>>> @typing.runtime_checkable
class FooProto(typing.Protocol):
	def foo():
		...
	def bar():
		...

		
>>> class Bar():
	def foo():
		...

		
>>> class Foo():
	def foo():
		...
	def bar():
		...

		
>>> issubclass(Foo, FooProto)
True
>>> issubclass(Bar, FooProto)
False
```
is the behaviour i saw

hikari/impl/event_dispatcher_base.py
```diff
@@ -51,79 +51,57 @@
...
+        self, event_name: str, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject
     ) -> None:
-        callbacks = self._consumers[event_name.casefold()]
-        asyncio.gather(
-            *(self._handle_dispatch(callback, shard, payload) for callback in callbacks), return_exceptions=True
-        )
+        try:
+            callback = self._consumers[event_name.casefold()]
+        except KeyError:
+            raise LookupError(event_name) from None
```
FasterSpeeding, line 94: A KeyError is a LookupError

hikari/api/event_dispatcher.py
```diff
@@ -54,74 +54,25 @@ class EventManager(abc.ABC):
...
-
-        Raises
-        ------
-        LookupError
-            If the either `name` or `consumer` couldn't be found in the
-            registered raw consumers.
-        """  # noqa: E501 - Line too long
-
     @abc.abstractmethod
     def consume_raw_event(
```
FasterSpeeding, line 58: Also sorry if this is annoying but if we're keeping the consume method on the interface would it be worth it keeping the interface called event manager?

FasterSpeeding:
Other than those 2 comments, lgtm

FasterSpeeding (approved):

hikari/impl/event_manager_base.py
```diff
@@ -71,51 +71,23 @@ class EventManagerBase(event_manager.EventManager):
...
-
-        try:
-            self._consumers[name].remove(consumer)
-
-        except ValueError:
-            raise LookupError(name, consumer) from None
-
-        if not self._consumers[name]:
-            del self._consumers[name]
+                self._consumers[name[3:]] = member
```
FasterSpeeding, line 84: oh yeah, you did make sure that this behaviour is working right with that casefold down in consume_raw_event right?

davfsa: Yep :+1:

PR #446, opened 2021-01-21
Enable python 3.10 ci

davfsa:
### Checklist


davfsa:
Macos runners will be stuck using 10.15 instead of 11.0 due to the pools being private: https://github.com/actions/virtual-environments/issues/2486

PR #447, opened 2021-01-21
Change black's include pattern to include .pyi files

FasterSpeeding:
### Checklist


davfsa (approved):
Didn't realize we overwrote it there. Thought we already covered it :+1:

PR #449, opened 2021-01-25
Fix how we're provisionally setting the coroutine tracking depth

FasterSpeeding:
### Summary
Use sys.set_coroutine_origin_tracking_depth instead of non-existent loop.set_coroutine_tracking_depth

### Checklist


davfsa (approved):

davfsa (changes requested):
Do we want to set it in everything or only the loop?

FasterSpeeding:
> Do we want to set it in everything or only the loop?

I don't see any way to set this for a specific loop and this is thread specific anyway

davfsa (approved):

PR #452, opened 2021-01-25
Implement the fetch authorization information rest endpoint and model

FasterSpeeding:
### Summary
Implement the fetch authorization information rest endpoint and model

### Checklist

### DIFF
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -307,13 +307,36 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
             owner=self.deserialize_user(payload["owner"]),
             rpc_origins=payload["rpc_origins"] if "rpc_origins" in payload else None,
             summary=payload["summary"],
-            verify_key=bytes(payload["verify_key"], "utf-8") if "verify_key" in payload else None,
+            verify_key=bytes.fromhex(payload["verify_key"]) if "verify_key" in payload else None,
             icon_hash=payload.get("icon"),
             team=team,
             guild_id=snowflakes.Snowflake(payload["guild_id"]) if "guild_id" in payload else None,
             primary_sku_id=primary_sku_id,
             slug=payload.get("slug"),
             cover_image_hash=payload.get("cover_image"),
+        )
+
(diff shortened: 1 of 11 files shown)


hikari/impl/entity_factory.py
```diff
@@ -299,7 +299,7 @@ def deserialize_application(self, payload: data_binding.JSONObject) -> applicati
             owner=self.deserialize_user(payload["owner"]),
             rpc_origins=payload["rpc_origins"] if "rpc_origins" in payload else None,
             summary=payload["summary"],
-            verify_key=bytes(payload["verify_key"], "utf-8") if "verify_key" in payload else None,
+            verify_key=bytes.fromhex(payload["verify_key"]) if "verify_key" in payload else None,
```
davfsa, line 302: Don't really understand this change

FasterSpeeding: why not? It wasn't handling it properly before

davfsa: Actually, checking in the docs, this should just be a string :thinking:

FasterSpeeding: Except the docs state in the section for this object that this is base64 (which is wrong) encoded bytes and in other places state that it should be handled as hexed bytes

davfsa (dismissed):

hikari/impl/entity_factory.py
FasterSpeeding: btw this field is also known as the application's public key and it might be worth renaming it since erm it's only ever called verify key here

davfsa: Hmmm, docs are weird for this, so ima just trust you on this one. As per the rename, sure, go for it.

davfsa (approved):

hikari/impl/entity_factory.py
FasterSpeeding: Just renamed it to "public_key" and fixed our docs ~~which were also wrong cause copied from discord's docs~~

davfsa (approved):

davfsa (approved):

PR #455, opened 2021-01-26
Add entities and rest endpoints for guild member screening

### LINKED ISSUE #409: Implement new member verification gate endpoints
https://github.com/discord/discord-api-docs/pull/2396

FasterSpeeding:
### Summary
Add entities and rest endpoints for guild member screening

### Checklist

### Related issues

Closes #409

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -5048,6 +5048,107 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         """
 
     @abc.abstractmethod
+    async def fetch_guild_membership_gate(
+        self, guild: snowflakes.SnowflakeishOr[guilds.PartialGuild]
+    ) -> guilds.MembershipGate:
+        """Fetch a guild's membership verification gate screen.
+
+        Parameters
+        ----------
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            Object or ID of the guild to fetch the member verification screen for
+
+        Returns
+        -------
+        hikari.guilds.MembershipGate
+            The requested membership verification gate screen.
+
+        Raises
+        ------
+        hikari.errors.ForbiddenError
+            If you are not part of the guild or if you are missing the
+            "MANAGE_GUILD" permission or if the guild does not have access to the
+            feature (is not a community guild).
+        hikari.errors.NotFoundError
+            If the guild is not found.
+        hikari.errors.UnauthorizedError
+            If you are unauthorized to make the request (invalid/missing token).
+        hikari.errors.RateLimitTooLongError
+            Raised in the event that a rate limit occurs that is
+            longer than `max_rate_limit` when making a request.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
+            rate-limits automatically. This includes most bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
+        hikari.errors.InternalServerError
+            If an internal error occurs on Discord while handling the request.
+        """
+
+    @abc.abstractmethod
+    async def edit_guild_membership_gate(
+        self,
+        guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
+        *,
+        description: undefined.UndefinedNoneOr[str] = undefined.UNDEFINED,
+        enabled: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
+        fields: undefined.UndefinedOr[typing.Sequence[guilds.MembershipGateField]] = undefined.UNDEFINED,
+    ) -> guilds.MembershipGate:
+        """Edit the verification gate screen of a community guild.
+
+        Parameters
+        ----------
+        guild : hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]
+            ID or object of the guild to edit the welcome screen for.
+
+        Other Parameters
+        ----------------
+        description : hikari.undefined.UndefinedNoneOr[builtins.str]
+            The new description to set. This may be passed as `builtins.None` to
+            remove the set description or left as `hikari.undefined.UNDEFINED`
+            to leave the description unchanged.
+        enabled : hikari.undefined.UndefinedOr[builtins.bool]
+            Whether the member screening feature should be enabled or not.
+            Leave as `hikari.undefined.UNDEFINED` to leave unchanged.
+        fields : hikari.undefined.UndefinedOr[typing.Sequence[hikari.guilds.MembershipGateField]]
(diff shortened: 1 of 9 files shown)


hikari/api/rest.py
```diff
@@ -5047,6 +5047,107 @@ async def edit_welcome_screen(
...
+            feature (is not a community guild).
+        hikari.errors.NotFoundError
+            If the guild is not found.
+        hikari.errors.UnauthorizedError
+            If you are unauthorized to make the request (invalid/missing token).
+        hikari.errors.RateLimitTooLongError
+            Raised in the event that a rate limit occurs that is
+            longer than `max_rate_limit` when making a request.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
```
A5rocks, line 5080: is hikari canonically lower or uppercase? (the readme has lowercase hikari)

hikari/api/rest.py
```diff
@@ -5047,6 +5047,107 @@ async def edit_welcome_screen(
...
+
+        Other Parameters
+        ----------------
+        description : hikari.undefined.UndefinedNoneOr[builtins.str]
+            The new description to set. This may be passed as `builtins.None` to
+            remove the set description or left as `hikari.undefined.UNDEFINED`
+            to leave the description unchanged.
+        enabled : hikari.undefined.UndefinedOr[builtins.bool]
+            Whether the member screening feature should be enabled or not.
+            Leave as `hikari.undefined.UNDEFINED` to leave unchanged.
```
A5rocks, line 5115: Should this "leave as default to leave it unchanged" be moved to the description? (also, 2 different styles here... gah... "leave as ..." and "left as ...")

hikari/api/rest.py
```diff
@@ -5047,6 +5047,107 @@ async def edit_welcome_screen(
...
+            screening feature.
+        hikari.errors.NotFoundError
+            If the guild is not found.
+        hikari.errors.UnauthorizedError
+            If you are unauthorized to make the request (invalid/missing token).
+        hikari.errors.RateLimitTooLongError
+            Raised in the event that a rate limit occurs that is
+            longer than `max_rate_limit` when making a request.
+        hikari.errors.RateLimitedError
+            Usually, Hikari will handle and retry on hitting
```
A5rocks, line 5140: again... lowercase or uppercase hikari (I sure hope it isn't alternating everywhere -- if it is feel free to leave this as is and I'll just make an issue).

hikari/impl/entity_factory.py
```diff
@@ -1055,6 +1055,27 @@ def deserialize_member(
...
+        fields: typing.List[guild_models.MembershipGateField] = []
+        for field_payload in payload["form_fields"]:
+            fields.append(
+                guild_models.MembershipGateField(
+                    type=guild_models.VerificationGateFieldType(field_payload["field_type"]),
+                    label=field_payload["label"],
+                    values=field_payload["values"] if "values" in field_payload else [],
+                    is_required=field_payload["required"],
+                )
+            )
```
A5rocks, line 1068: shouldn't this be a list comprehension?

hikari/impl/rest.py
```diff
@@ -2425,6 +2425,42 @@ async def edit_welcome_screen(
...
+    async def fetch_guild_membership_gate(
+        self, guild: snowflakes.SnowflakeishOr[guilds.PartialGuild]
+    ) -> guilds.MembershipGate:
+        route = routes.GET_GUILD_MEMBER_VERIFICATION.compile(guild=guild)
+        response = await self._request(route)
+
+        # If the member screen of a guild has never been set
+        # then Discord will respond with a 204 (no content).
+        if not response:
+            return guilds.MembershipGate(description=None, fields=[], updated_at=None)
```
A5rocks, line 2437: Should this maybe return a `None` since there's no member screen? ... Unless I'm misinterpreting the comment, that is.

A5rocks (commented):
Dunno, looks good to me for the most part but IDK hikari...

hikari/api/rest.py
FasterSpeeding: It should always be upper case in docs and comments since it's a name (same would apply to stuff like Discord but we're not as consistent with that one lol)

hikari/impl/rest.py
FasterSpeeding: We'd never return a None in REST for a GET request cause that's what NotFoundError but even then for this endpoint that only applies to if the guild existed or not. Regardless of whether this is enabled or not as long as the guild exists this entity should also exist just Discord rather than having a special case which handles when it's never been set and makes sure the returned data fits their defined model for the endpoint (as opposed to idk just returning the empty string they fetched from a database) they instead seem to just return nothing which is undocumented behaviour and inconsistent.

hikari/impl/entity_factory.py
FasterSpeeding: This is a bit of an unspoken style thing we've been semi-following here but we've been avoiding list comprehensions that cover more than a couple of lines at mostly cause that can lead to messy looking code

hikari/impl/rest.py
```diff
@@ -2425,6 +2425,42 @@ async def edit_welcome_screen(
...
         return self._entity_factory.deserialize_welcome_screen(response)
 
+    async def fetch_guild_membership_gate(
+        self, guild: snowflakes.SnowflakeishOr[guilds.PartialGuild]
+    ) -> guilds.MembershipGate:
+        route = routes.GET_GUILD_MEMBER_VERIFICATION.compile(guild=guild)
+        response = await self._request(route)
+
+        # If the member screen of a guild has never been set
+        # then Discord will respond with a 204 (no content).
```
davfsa, line 2435: Hopefully we can remove this in the future. https://github.com/discord/discord-api-docs/issues/2530

tests/hikari/impl/test_entity_factory.py
```diff
@@ -1635,6 +1635,52 @@ def test_deserialize_member_with_passed_through_user_object_and_guild_id(self, e
...
+
+        assert len(screen.fields) == 2
+        # MemberScreenField
+        field = screen.fields[0]
+        assert field.type is guild_models.VerificationGateFieldType.SERVER_RULES
+        assert field.label == "Read and agree to the server rules"
+        assert field.values == ["Be treated.", "Go Spam DM this guild's users."]
+        assert field.is_required is True
+        assert isinstance(field, guild_models.MembershipGateField)
+        assert screen.fields[1].values == []
```
davfsa, line 1667: Should also be checking that the second field is what you expect

FasterSpeeding: the 2nd field only exists to cover the case of fields being null

davfsa (approved):

tests/hikari/impl/test_entity_factory.py
davfsa: We should still be testing all of the values

FasterSpeeding: Even though those cases are already covered?

davfsa: Just to make sure it acts the same way as we expect it to

PR #456, opened 2021-01-26
Fix permissions being mishandled in audit logs and incorrect change keys

FasterSpeeding:
### Summary
Fix permissions being mishandled in audit logs and incorrect change keys
* Added some debug logging to the entity factory
* Fix the value type on enums.Flag to be specifically int

### Checklist


hikari/impl/entity_factory.py
```diff
@@ -402,15 +410,19 @@ def deserialize_audit_log(self, payload: data_binding.JSONObject) -> audit_log_m
...
                     new_value: typing.Any = change_payload.get("new_value")
                     old_value: typing.Any = change_payload.get("old_value")
                     if value_converter := self._audit_log_entry_converters.get(key):
                         new_value = value_converter(new_value) if new_value is not None else None
                         old_value = value_converter(old_value) if old_value is not None else None
 
+                    elif __debug__ and not isinstance(key, audit_log_models.AuditLogChangeKey):
+                        _LOGGER.debug("Unknown audit log change key found %r", key)
+
                     changes.append(audit_log_models.AuditLogChange(key=key, new_value=new_value, old_value=old_value))
```
A5rocks, line 426: uh... doesn't this break type safety? `key` is not guaranteed to be a `AuditLogChangeKey`... (Or at least, you say it might be by putting that `if` statement above. Not sure of the context here).

hikari/impl/entity_factory.py
```diff
@@ -402,15 +410,19 @@ def deserialize_audit_log(self, payload: data_binding.JSONObject) -> audit_log_m
...
+                    )
 
                     new_value: typing.Any = change_payload.get("new_value")
                     old_value: typing.Any = change_payload.get("old_value")
                     if value_converter := self._audit_log_entry_converters.get(key):
                         new_value = value_converter(new_value) if new_value is not None else None
                         old_value = value_converter(old_value) if old_value is not None else None
 
+                    elif __debug__ and not isinstance(key, audit_log_models.AuditLogChangeKey):
+                        _LOGGER.debug("Unknown audit log change key found %r", key)
```
A5rocks, line 424: Also, I think discord sending invalid data should probably be an `info` or so, so that you don't have to be debugging something to complain to discord and get it fixed for everyone... Dunno logging conventions though.

A5rocks (commented):
Also, the `str` on line 65 of `audit_logs.py` is extra (though it's probably there to be extra explicit).

Otherwise, two questions that are probably just me missing context for this bit of code.

hikari/impl/entity_factory.py
FasterSpeeding: key is typed as being a Union of AuditLogChangeKey or str on AuditLogChange inline with how we type other attributes which are typed as enum values

hikari/impl/entity_factory.py
FasterSpeeding: This is less invalid data and more just data we haven't implemented the handling for yet which in other places is kept at debug log level right now

davfsa (approved):

PR #461, opened 2021-01-28
Remove partial implementation from GuildBuilder ABC

FasterSpeeding:
### Summary
Remove partial implementation from GuildBuilder ABC
* replace attrs instance variables with settable abstract proprties

### Checklist

### DIFF
--- a/hikari/impl/special_endpoints.py
+++ b/hikari/impl/special_endpoints.py
@@ -58,6 +58,7 @@ if typing.TYPE_CHECKING:
     from hikari import messages
     from hikari import permissions as permissions_
     from hikari import users
+    from hikari import voices
     from hikari.api import entity_factory as entity_factory_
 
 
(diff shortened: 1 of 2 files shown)


hikari/impl/special_endpoints.py
```diff
@@ -218,6 +220,19 @@ class GuildBuilder(special_endpoints.GuildBuilder):
...
+    )
+    explicit_content_filter_level: undefined.UndefinedOr[guilds.GuildExplicitContentFilterLevel] = attr.ib(
+        default=undefined.UNDEFINED
+    )
+    icon: undefined.UndefinedOr[files.Resourceish] = attr.ib(default=undefined.UNDEFINED)
+    region: undefined.UndefinedOr[voices.VoiceRegionish] = attr.ib(default=undefined.UNDEFINED)
+    verification_level: undefined.UndefinedOr[typing.Union[guilds.GuildVerificationLevel, int]] = attr.ib(
+        default=undefined.UNDEFINED
+    )
+
```
davfsa, line 235: These should go first

davfsa (changes requested):

hikari/impl/special_endpoints.py
FasterSpeeding: Shouldn't they go after the required arguments rather than before?

davfsa: You fixed it already. I meant that they should go before the non-arguments

davfsa (approved):

PR #469, opened 2021-02-03
Rewrite `protocol.Protocol` to a typing extension

### LINKED ISSUE #467: protocol.Protocol isn't runtime generic
### Steps to reproduce
run this code on while targetting master
```py
import typing

from hikari.internal import protocol

TypeT = typing.TypeVar("TypeT")

protocol.Protocol[TypeT]
```

### Expected Result
It should pass like the following equivalent code would

```py
import typing

TypeT = typing.TypeVar("TypeT")

typing.Protocol[TypeT]
```

### Actual Result
It errors with the following traceback
```py
  File "C:\Users\...\PycharmProjects\hikari\bored.py", line 7, in <module>
    protocol.Protocol[TypeT]
  File "C:\Python39\lib\typing.py", line 262, in inner
    return func(*args, **kwds)
  File "C:\Python39\lib\typing.py", line 980, in __class_getitem__
    _check_generic(cls, params, len(cls.__parameters__))
  File "C:\Python39\lib\typing.py", line 196, in _check_generic
    raise TypeError(f"{cls} is not a generic class")
TypeError: <class 'hikari.internal.protocol.Protocol'> is not a generic class
```

### System info
```
hikari (SNAPSHOT) [HEAD]
CPython 3.9.1 MSC v.1927 64 bit (AMD64)
Windows YangWen-li 10 10.0.21277 AMD64 AMD64 Family 23 Model 113 Stepping 0, AuthenticAMD
```

### Further info
targeting the master branch

### Checklist

davfsa:
### Checklist

### Related issues
Fixes #467

### DIFF
--- a/hikari/internal/attr_extensions.py
+++ b/hikari/internal/attr_extensions.py
@@ -20,7 +20,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-"""Utility for extending and optimising the usage of attr models."""
+"""Utility for extending and optimising the usage of `attr` models."""
 from __future__ import annotations
 
 __all__: typing.List[str] = [
--- a/hikari/internal/enums.py
+++ b/hikari/internal/enums.py
@@ -116,7 +116,6 @@ class _EnumMeta(type):
     def __iter__(cls) -> typing.Iterator[typing.Any]:
         yield from cls._name_to_member_map_.values()
 
-    @staticmethod
     def __new__(
         mcs: typing.Type[_T],
         name: str,
--- a/hikari/internal/enums.pyi
+++ b/hikari/internal/enums.pyi
@@ -30,6 +30,8 @@
 # ship my own MyPy plugin for this, so just make MyPy think that the types
 # we are using are just aliases from the enum types in the standard library.
 
+__all__ = ["Enum", "Flag"]
+
 import enum as __enum
 from typing import Iterator as __Iterator
 from typing import Sequence as __Sequence
@@ -69,5 +71,3 @@ class Flag(__enum.IntFlag):
     __rsub__ = __sub__ = difference
     __rxor__ = __xor__ = symmetric_difference
     __invert__ = invert
-
-__all__ = ["Enum", "Flag"]
--- a/hikari/internal/fast_protocol.py
+++ b/hikari/internal/fast_protocol.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+# cython: language_level=3
 # Copyright (c) 2020 Nekokatt
 # Copyright (c) 2021 davfsa
 #
@@ -19,33 +20,28 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-"""Implementation of parts of Python's `typing.Protocol` to be more performant."""
+"""A utility for faster `typing.Protocol` instance checks."""
 
 from __future__ import annotations
 
-__all__: typing.List[str] = []
+__all__: typing.List[str] = ["FastProtocolChecking"]
 
-import abc
 import typing
 
 if typing.TYPE_CHECKING:
     _T = typing.TypeVar("_T")
 
 _Protocol = NotImplemented
-IGNORED_ATTRS = typing.EXCLUDED_ATTRIBUTES + ["__qualname__", "__slots__"]
+_IGNORED_ATTRS = typing.EXCLUDED_ATTRIBUTES + ["__qualname__", "__slots__"]
 
 
 def _check_if_ignored(name: str) -> bool:
-    return name.startswith("_abc_") or name in IGNORED_ATTRS
+    return name.startswith("_abc_") or name in _IGNORED_ATTRS
 
 
-def _no_init(cls) -> None:
-    if type(cls)._is_protocol:
-        raise TypeError("Protocols cannot be instantiated")
-
-
-class _FastProtocol(abc.ABCMeta):
-    @staticmethod
+# This metaclass needs to subclass the same type as `typing.Protocol` to be
+# able to overwrite it
+class _FastProtocolChecking(type(typing.Protocol)):
     def __new__(
         cls: typing.Type[_T],
         cls_name: str,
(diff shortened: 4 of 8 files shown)


hikari/internal/protocol.py
```diff
@@ -102,6 +105,29 @@ class Protocol(typing.Generic, metaclass=_FastProtocol):
...
+            # Generic and Protocol can only be subscripted with unique type variables.
+            if not all(isinstance(p, typing.TypeVar) for p in params):
+                raise TypeError("Parameters to Protocol[...] must all be type variables")
+            if len(set(params)) != len(params):
+                raise TypeError("Parameters to Protocol[...] must all be unique")
+        else:
+            # Subscripting a regular Generic subclass.
+            typing._check_generic(cls, params)
+
+        return typing._GenericAlias(cls, params)
```
FasterSpeeding, line 129: Are you actually ok with this fix and how it's become a pretty complex, edge casey, possibly difficult to maintain and possibly still broken in ways which we can't forsee "fix" which seems to just bring in even more issues and ends up having to rely the protected impl detail of stdlib stuffe for someting that could've otherwise been fixed by adding a few more attributes to places and also makes our "protocols" incompatible with standard protocols?

davfsa: As I said in the issue, i dont really like this, but it is the only way to fix it. I have been trying to figure out a better way, but none for the moment.

FasterSpeeding: Well as it stands this is probably one of the more breaking and complex approaches to fixing the issue of slow instance checks being used internally that could've been picked outside of just removing stuff as it involves changing pre-established parts of the public interface as opposed to just using a helper function which handles it using a different approach (maybe duck typing) or adding attributes which is an addition not a change and works better with static typing

davfsa:
Changed the scope of this pr to just leaving clear that it is just a subclass of `typing.Protocol` that we are implementing. This will also help as it doesn't shadow everything that `typing` implement, just what we want.

FasterSpeeding (approved):
I assume you've checked the behaviour for this

davfsa:
> I assume you've checked the behaviour for this

I have, but would rather you have another look.

FasterSpeeding:
> > I assume you've checked the behaviour for this
> 
> I have, but would rather you have another look.

The scope of this class is honestly too wide for me to check since I don't know what the typing module covers within the stdlib and standard methods as far as nuanced and special case behaviour goes and don't really think it's worth the time looking into all of that for something unrelated in a lib thats also underelated but the scope of the issue this pr is for is only to make sure that getitem works recursively on protocol.Protocol which should be a simple yes or no unless I'm missing something.
 
And seeing as there's no consistent library or language standard that's been being enforced recently I've got nothing to comment on there either

FasterSpeeding:
Did you actually get the generic behaviourto work on this at run time, i still seem to be getting this traceback when running the following code while targetting your branch 
```py
<class 'typing._ProtocolMeta'>
  File "C:\Users\blah\PycharmProjects\hikari\bored.py", line 15, in <module>
    class FooProto(protocol.Protocol[TypeT]):
  File "C:\Python39\lib\typing.py", line 262, in inner
    return func(*args, **kwds)
  File "C:\Python39\lib\typing.py", line 980, in __class_getitem__
    _check_generic(cls, params, len(cls.__parameters__))
  File "C:\Python39\lib\typing.py", line 196, in _check_generic
    raise TypeError(f"{cls} is not a generic class")
TypeError: <class 'hikari.internal.protocol.Protocol'> is not a generic class
```
```py
import typing

from hikari.internal import protocol

TypeT = typing.TypeVar("TypeT")


print(protocol.TYPING_PROTOCOL_METACLASS )
@typing.runtime_checkable
class FooProtoVanilla(typing.Protocol[TypeT]):
    def foo(self) -> TypeT:
        raise NotImplementedError


class FooProto(protocol.Protocol[TypeT]):
    def foo(self) -> TypeT:
        raise NotImplementedError


Foo = type("Foo", (), {"foo": lambda _: 42})

foo = Foo()
print(issubclass(Foo, FooProto))
print(isinstance(foo, FooProto))

print(issubclass(Foo, FooProtoVanilla))
print(isinstance(foo, FooProtoVanilla))

```

davfsa:
> 
> 
> Did you actually get the generic behaviour to work on this at run time, i still seem to be getting this traceback when running the following code while targetting your branch
> 
> ```python
> <class 'typing._ProtocolMeta'>
> Traceback (most recent call last):
>   File "C:\Users\blah\PycharmProjects\hikari\bored.py", line 15, in <module>
>     class FooProto(protocol.Protocol[TypeT]):
>   File "C:\Python39\lib\typing.py", line 262, in inner
>     return func(*args, **kwds)
>   File "C:\Python39\lib\typing.py", line 980, in __class_getitem__
>     _check_generic(cls, params, len(cls.__parameters__))
>   File "C:\Python39\lib\typing.py", line 196, in _check_generic
>     raise TypeError(f"{cls} is not a generic class")
> TypeError: <class 'hikari.internal.protocol.Protocol'> is not a generic class
> ```
> 
> ```python
> import typing
> 
> from hikari.internal import protocol
> 
> TypeT = typing.TypeVar("TypeT")
> 
> 
> print(protocol.TYPING_PROTOCOL_METACLASS )
> @typing.runtime_checkable
> class FooProtoVanilla(typing.Protocol[TypeT]):
>     def foo(self) -> TypeT:
>         raise NotImplementedError
> 
> 
> class FooProto(protocol.Protocol[TypeT]):
>     def foo(self) -> TypeT:
>         raise NotImplementedError
> 
> 
> Foo = type("Foo", (), {"foo": lambda _: 42})
> 
> foo = Foo()
> print(issubclass(Foo, FooProto))
> print(isinstance(foo, FooProto))
> 
> print(issubclass(Foo, FooProtoVanilla))
> print(isinstance(foo, FooProtoVanilla))
> ```

I am aware of this. I have switched `protocol.Protocol` from being a replacement to `typing.Protocol` as just being a subclass of the Protocol (making it another protocol, just like the traits) which has a faster implementation of `isinstance` and `issubclass`

FasterSpeeding:
> I am aware of this. I have switched protocol.Protocol from being a replacement to typing.Protocol as just being a subclass of the Protocol (making it another protocol, just like the traits) which has a faster implementation of isinstance and issubclass

In which case wouldn't typing.Protocol also have to be inherited for each of those trait classes along side the direct or indirect inheritance of protocol.Protocol rather than protocol.Protocol always being directly inherited since otherwise they'd be considered not protocol. Also doesn't that mean that all these protocols still need to be marked as runtime checkable

hikari/internal/typing_extensions.pyi
```diff
@@ -20,9 +20,13 @@
...
 # SOFTWARE.
-"""Typehints for `hikari.internal.protocol`."""
+"""Typehints for `hikari.internal.typing_extensions`."""
 from typing import Protocol as __Protocol
+from typing import runtime_checkable as __runtime_checkable
 
-class Protocol(__Protocol): ...
+# fmt: off - Stops isort and black from fighting. Also, black makes this look weird
+@__runtime_checkable
+class FastProtocolChecking(__Protocol): ...
```
FasterSpeeding, line 29: isn't this whole .pyi module unnecessary

hikari/internal/typing_extensions.pyi
```diff
@@ -20,9 +20,13 @@
...
 
-__all__ = ["Protocol"]
+# fmt: on
+__all__ = ["FastProtocolChecking"]
```
FasterSpeeding, line 32: shouldn't this be before the imports?

hikari/internal/typing_extensions.pyi
```diff
@@ -20,9 +20,13 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-"""Typehints for `hikari.internal.protocol`."""
+"""Typehints for `hikari.internal.typing_extensions`."""
 from typing import Protocol as __Protocol
+from typing import runtime_checkable as __runtime_checkable
 
-class Protocol(__Protocol): ...
+# fmt: off - Stops isort and black from fighting. Also, black makes this look weird
```
FasterSpeeding, line 27: How do they "fight" and what's wrong with "weird" if it follows the repo's formatting convention (unless your inference here is that it's a bug in which case it should probably be labled as that with a link to an issue for reference as to when the comments can just be removed)

hikari/internal/typing_extensions.py
```diff
@@ -55,62 +51,79 @@ def __new__(
         global _Protocol
 
         if _Protocol is NotImplemented:
-            if cls_name != "Protocol":
-                raise TypeError("First instance of _FastProtocol must be Protocol")
+            if cls_name != "FastProtocolChecking":
```
FasterSpeeding, line 54: can't you just check against the class itself so it doesn't have collision risks?

FasterSpeeding (approved):

PR #471, opened 2021-02-11
Even more testing

davfsa:
### Checklist


tests/hikari/test_audit_logs.py
```diff
@@ -57,6 +102,42 @@ def test_repr(self):
...
+            id=123,
+            target_id=None,
+            changes=[],
+            user_id=None,
+            action_type=0,
+            options=None,
+            reason=None,
+        )
+
+        assert await model.fetch_user() is None
```
FasterSpeeding, line 119: btw is this the decided way for handling this scenario? Kind of feels similar to the webhook methods where if the token isn't present we raise as we can't finish the operation without it and not sure if there's any semantic difference that i'm missing

davfsa: Will have a think

davfsa: Yea, this is the way we are going to do to it

PR #473, opened 2021-02-12
Enums consistency changes/fixes

FasterSpeeding:
### Summary
* Fixup enums.Enum and enums.IntFlag to allow methods to be overwritten
* Add in a special case to let string enums return their value when cast to string rather than name
* Remove unnecessary/out-dated str operator tests that targeted enums which now inherit the behaviour

### Checklist

### DIFF
--- a/hikari/internal/enums.py
+++ b/hikari/internal/enums.py
@@ -140,13 +140,19 @@ class _EnumMeta(type):
             "_member_names_": (member_names := []),
             # Required to be immutable by enum API itself.
             "__members__": types.MappingProxyType(namespace.names_to_values),
-            **namespace,
             **{
                 name: value
                 for name, value in Enum.__dict__.items()
                 if name not in ("__class__", "__module__", "__doc__")
             },
         }
+
+        # We don't want to override the __str__ behaviour inherited from builtins.str for string based enums.
+        if issubclass(base, str):
+            new_namespace.pop("__str__", None)
+
+        # We update the name space to ensure new fields override inherited attributes and methods.
+        new_namespace.update(namespace)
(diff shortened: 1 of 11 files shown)


hikari/internal/enums.py
```diff
@@ -395,7 +402,6 @@ def __new__(
             "_member_names_": (member_names := []),
             # Required to be immutable by enum API itself.
             "__members__": types.MappingProxyType(namespace.names_to_values),
-            **namespace,
```
davfsa, line 398: Moving this to the bottom will achieve the same effect as `new_namespace.update`

hikari/internal/enums.py
```diff
@@ -140,14 +140,20 @@ def __new__(
...
                 name: value
                 for name, value in Enum.__dict__.items()
                 if name not in ("__class__", "__module__", "__doc__")
             },
         }
 
+        # We don't want to override the __str__ behaviour inherited from builtins.str for string based enums.
+        if issubclass(base, str):
+            new_namespace.pop("__str__", None)
+
```
davfsa, line 153: Why not? What difference does it make?

hikari/internal/enums.py
FasterSpeeding: While relying on implementation detail on the order in which those are handled and being less explicit that it's meant to be handled after the fields are brought in from Flag

hikari/internal/enums.py
FasterSpeeding: This is the difference between a string enum value acting as its value rather than its name

davfsa (approved):

PR #475, opened 2021-02-23
Fix mis-usage of PAYLOAD_ZLIB_STREAM where TRANSPORT_ZLIB_STREAM is being referred to

FasterSpeeding:
### Summary
Fix mis-usage of PAYLOAD_ZLIB_STREAM where TRANSPORT_ZLIB_STREAM is being referred to

### Checklist

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -342,7 +342,7 @@ class GatewayShardImpl(shard.GatewayShard):
     ----------------
     compression : typing.Optional[buitlins.str]
         Compression format to use for the shard. Only supported values are
-        `"payload_zlib_stream"` or `builtins.None` to disable it.
+        `"transport_zlib_stream"` or `builtins.None` to disable it.
     initial_activity : typing.Optional[hikari.presences.Activity]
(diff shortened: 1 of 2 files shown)


hikari/impl/shard.py
```diff
@@ -342,7 +342,7 @@ class GatewayShardImpl(shard.GatewayShard):
     ----------------
     compression : typing.Optional[buitlins.str]
         Compression format to use for the shard. Only supported values are
-        `"payload_zlib_stream"` or `builtins.None` to disable it.
+        `"transport_zlib_stream"` or `builtins.None` to disable it.
```
davfsa, line 345: Actually, this is wrong, we only allow `"zlib-stream"`

davfsa (approved):

hikari/impl/shard.py
FasterSpeeding: That's the raw value we send to discord, not the value of TRANSPORT_ZLIB_STREAM

davfsa: True

davfsa (approved):

PR #476, opened 2021-02-23
Send a full heartbeat on a heartbeat request, not just an ACK

FasterSpeeding:
### Summary
According to the [API's docs](https://discord.com/developers/docs/topics/gateway#heartbeating) ` The gateway may request a heartbeat from you in some situations, and you should send a heartbeat back to the gateway as you normally would` with the bit "you should send a heartbeat back to the gateway as you normally would" infers that we should be responding with a full heartbeat which is inline with how [JDA](https://github.com/DV8FromTheWorld/JDA/blob/3a9fcd0ecc88f988ac81eb2c18527283a83f12d1/src/main/java/net/dv8tion/jda/internal/requests/WebSocketClient.java#L843) and [Detritus](https://github.com/detritusjs/client-socket/blob/4361739b1c824790e269db4f3abbde63fe5a5d24/src/gateway.ts#L492) handle received heartbeats and matches up with how `HEARTBEAT_ACK` is marked as Received only in the [API's docs](https://discord.com/developers/docs/topics/opcodes-and-status-codes#gateway-gateway-opcodes)

### Checklist

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -744,7 +744,7 @@ class GatewayShardImpl(shard.GatewayShard):
             self._logger.log(ux.TRACE, "dispatching %s with seq %s", t, s)
             self._dispatch(t, s, d)
         elif op == _HEARTBEAT:
-            await self._send_heartbeat_ack()
+            await self._send_heartbeat()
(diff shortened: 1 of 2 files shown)


davfsa (approved):
Nice spot

PR #478, opened 2021-02-23
Improve the UX around RESTClient.delete_messages

FasterSpeeding:
### Summary
* Enforce that at least one message is passed in the interface by adding a required positional message argument before the other_messages variable length positional argument.
* Allow the first argument to either be SnowflakeishOr[PartialMessage] or an Iterable[SnowflakeishOr[PartialMessage]] to support directly passing an iterable as this could otherwise be a pretty common mistake.

### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1137,13 +1137,26 @@ class RESTClientImpl(rest_api.RESTClient):
     async def delete_messages(
         self,
         channel: snowflakes.SnowflakeishOr[channels_.TextChannel],
+        messages: typing.Union[
+            snowflakes.SnowflakeishOr[messages_.PartialMessage],
+            snowflakes.SnowflakeishIterable[messages_.PartialMessage],
+        ],
         /,
-        *messages: snowflakes.SnowflakeishOr[messages_.PartialMessage],
+        *other_messages: snowflakes.SnowflakeishOr[messages_.PartialMessage],
     ) -> None:
         route = routes.POST_DELETE_CHANNEL_MESSAGES_BULK.compile(channel=channel)
 
-        pending: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque(messages)
+        pending: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque()
         deleted: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque()
+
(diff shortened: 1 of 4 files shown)


hikari/impl/rest.py
```diff
@@ -1137,14 +1137,27 @@ async def delete_message(
...
 
-        pending: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque(messages)
+        pending: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque()
         deleted: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque()
 
+        if isinstance(messages, typing.Iterable):
+            pending.extend(messages)
+
+        else:
+            pending.append(messages)
```
davfsa, line 1156: Mind adding a comment? `# Syntactic sugar. Allows to use iterables`

hikari/impl/rest.py
```diff
@@ -1137,14 +1137,27 @@ async def delete_message(
...
-        *messages: snowflakes.SnowflakeishOr[messages_.PartialMessage],
+        *other_messages: snowflakes.SnowflakeishOr[messages_.PartialMessage],
     ) -> None:
         route = routes.POST_DELETE_CHANNEL_MESSAGES_BULK.compile(channel=channel)
 
-        pending: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque(messages)
+        pending: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque()
         deleted: typing.Deque[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = collections.deque()
 
+        if isinstance(messages, typing.Iterable):  # Syntactic sugar. Allows to use iterables
```
davfsa, line 1152: Just for consistency, could you move the comment above the line?

Then feel free to merge :)

FasterSpeeding: Does this mean that other places where short comments aren't put on their own line also be refactored for consistency at one point?

davfsa: This isnt a comment on that specific line, this is a comment about the block that is under it, which is why it should be above

FasterSpeeding: hmm why above? isn't it generally kept within the if statement in this module since that's where the block starts

davfsa (approved):

PR #479, opened 2021-02-23
Add the use application commands permission

FasterSpeeding:
### Summary
Add the use application commands permission

### Checklist

### DIFF
--- a/hikari/permissions.py
+++ b/hikari/permissions.py
@@ -186,3 +186,6 @@ class Permissions(enums.Flag):
 
     MANAGE_EMOJIS = 1 << 30
     """Allows management and editing of emojis."""
+
+    USE_APPLICATION_COMMANDS = 1 << 31
+    """Allows for using the application commands of guild integrations within a text channel."""


hikari/permissions.py
```diff
@@ -186,3 +186,6 @@ class Permissions(enums.Flag):
 
     MANAGE_EMOJIS = 1 << 30
     """Allows management and editing of emojis."""
+
+    USE_APPLICATION_COMMANDS = 1 << 31
```
davfsa, line 190: Its currently called `Use Slash Commands` in the app. Should we use this terminology or "Application commands"?

FasterSpeeding: Slash commands is a purely front end name and when they're named on the backend they're called application commands; I've been trying to consistently refer to them as application commands within the lib for similar reasons to why we refer to guilds as guilds rather than servers (in both cases the API docs inconsistently use both in places although it's more inconsistent for slash commands)

davfsa: Ok, thx, noted 👍

davfsa (approved):

PR #480, opened 2021-02-23
Add application flags and ID to the ready event.

FasterSpeeding:
### Summary
Add application flags and ID to the ready event.

### Checklist

### DIFF
--- a/hikari/applications.py
+++ b/hikari/applications.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 
 __all__: typing.List[str] = [
     "Application",
+    "ApplicationFlags",
     "AuthorizationApplication",
     "AuthorizationInformation",
     "ConnectionVisibility",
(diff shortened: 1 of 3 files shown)


hikari/applications.py
```diff
@@ -58,6 +59,43 @@
...
+    """Denotes that a verified application can use the GUILD_MEMBERS intent."""
+
+    GUILD_MEMBERS_INTENT = 1 << 15
+    """Denotes that the application has the GUILD_MEMBERS intent enabled in it's dashboard."""
+
+    VERIFICATION_PENDING_GUILD_LIMIT = 1 << 16
+    """Denotes that the application's verification is pending."""
+
+    EMBEDDED = 1 << 17
+    # TODO: what does this mean
```
davfsa, line 96: Before merging this I would like to find out what this actually means

FasterSpeeding: The feature hasn't been released yet so there's no reference to document it against

davfsa: Then I think its best to wait and see what gets documented

FasterSpeeding: Since they don't merge most prs anymore that could mean waiting for years

davfsa: Could you add this?

```suggestion
    """Denotes that the application is embedded."""
```

or maybe


```suggestion
    """Denotes that the application is embedded into Discord."""
```
as this is probably what it means

davfsa (approved):

PR #501, opened 2021-03-15
Emoji validation

davfsa:
### Checklist

### DIFF
--- a/hikari/emojis.py
+++ b/hikari/emojis.py
@@ -29,9 +29,9 @@ __all__: typing.List[str] = ["Emoji", "UnicodeEmoji", "CustomEmoji", "KnownCusto
 import abc
 import re
 import typing
-import unicodedata
 
 import attr
+import discord_emojis
 
 from hikari import files
 from hikari import snowflakes
@@ -159,20 +159,7 @@ class UnicodeEmoji(Emoji):
     @property
     def filename(self) -> str:
         """Filename to use if re-uploading this emoji's PNG."""
-        codepoints = self.codepoints
-        # It looks like the rule is to delete character #2 if the value
-        # of this is 0xfe0f and the character is 4 characters long.
-        # Other than that, the mapping is 1-to-1. I'll set up a CI task to
-        # double check this each day so we know when Discord breaks it again.
-        # The gay-pride flag is an outlier, for god knows what reason. I don't
-        # care that much but if Discord start breaking this regularly I might
-        # need to ask for a more permanent solution.
-        # This is provisionally provided. If we find it breaks in other ways, I
-        # will just revoke this functionality in a future update.
-        if codepoints[1:2] == [0xFE0F] and len(codepoints) <= 4 and codepoints != [0x1F3F3, 0xFE0F, 0x200D, 0x1F308]:
-            codepoints = [codepoints[0], *codepoints[2:]]
-
-        return "-".join(hex(c)[2:] for c in codepoints) + ".png"
+        return "-".join(hex(c)[2:] for c in self.codepoints) + ".png"
 
     @property
     def url(self) -> str:
@@ -188,19 +175,9 @@ class UnicodeEmoji(Emoji):
 
         Example
         -------
-            https://github.com/twitter/twemoji/raw/master/assets/72x72/1f004.png
+            https://raw.githubusercontent.com/twitter/twemoji/master/assets/72x72/1f004.png
         """
         return _TWEMOJI_PNG_BASE_URL + self.filename
-
-    @property
-    @typing.final
-    def unicode_names(self) -> typing.Sequence[str]:
-        """Get the unicode name of the emoji as a sequence.
-
-        This returns the name of each codepoint. If only one codepoint exists,
-        then this will only have one item in the resulting sequence.
-        """
-        return [unicodedata.name(c) for c in self.name]
 
     @property
     @typing.final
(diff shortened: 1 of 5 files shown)


hikari/emojis.py
```diff
@@ -235,10 +241,6 @@ def parse(cls, string: str, /) -> UnicodeEmoji:
         UnicodeEmoji
             The parsed UnicodeEmoji object.
         """
-        # Ensure validity.
-        for i, codepoint in enumerate(string, start=1):
```
FasterSpeeding, line 239: So which is it, does this use unicodedata.name or does it not?

hikari/emojis.py
```diff
@@ -235,10 +241,6 @@ def parse(cls, string: str, /) -> UnicodeEmoji:
         UnicodeEmoji
             The parsed UnicodeEmoji object.
         """
-        # Ensure validity.
```
FasterSpeeding, line 238: Do we even need to keep the parse method if we're not keeping this behaviour around?

scripts/test_twemoji_mapping.py
```diff
@@ -54,16 +54,22 @@
...
+            emoji = emojis.UnicodeEmoji.parse(emoji_surrogates)
+
+            if emoji.filename in known_files:
+                valid_emojis.append((emoji_surrogates, name))
+                print("[  OK  ]", f"{i}/{n}", name, *map(hex, map(ord, emoji_surrogates)), emoji.url)
+            else:
+                invalid_emojis.append((emoji_surrogates, name))
+                print("[ FAIL ]", f"{i}/{n}", name, *map(hex, map(ord, emoji_surrogates)), emoji.url)
+
+        except Exception as ex:
```
FasterSpeeding, line 69: surely we don't need this behaviour and any exception should just raise since with the current changes it won't be expected to raise anywhere

scripts/test_twemoji_mapping.py
```diff
@@ -54,16 +54,22 @@
...
+        try:
+            emoji_surrogates = emoji["surrogates"]
+            name = emoji["primaryName"]
+            emoji = emojis.UnicodeEmoji.parse(emoji_surrogates)
+
+            if emoji.filename in known_files:
+                valid_emojis.append((emoji_surrogates, name))
+                print("[  OK  ]", f"{i}/{n}", name, *map(hex, map(ord, emoji_surrogates)), emoji.url)
+            else:
+                invalid_emojis.append((emoji_surrogates, name))
```
FasterSpeeding, line 66: Also if we kept the whole parse validates stuff behaviour around couldn't this pipeline in theory just directly call UnicodeEmoji's init instead of the parse method to get the same result?

scripts/test_twemoji_mapping.py
```diff
@@ -54,16 +54,22 @@
...
+                print("[ FAIL ]", f"{i}/{n}", name, *map(hex, map(ord, emoji_surrogates)), emoji.url)
+
+        except Exception as ex:
             invalid_emojis.append((emoji_surrogates, name))
```
FasterSpeeding, line 70: Something being broken doesn't necessarily mean the emojis are invalid

hikari/emojis.py
```diff
@@ -199,6 +199,12 @@ def unicode_names(self) -> typing.Sequence[str]:
 
         This returns the name of each codepoint. If only one codepoint exists,
         then this will only have one item in the resulting sequence.
+
+        !!! warning
+            This uses `unicodedata.name` to fetch the names. This list
+            only gets updated between python versions. This means
+            that this might raise `builtins.ValueError` if there is
```
FasterSpeeding, line 206: I feel like this could be better worded, something like "this method relies on the standard unicodedata module and thus is limited by the unicode version python knows" (also might make sense to call it a newer unicode version rather than a "new version of emojis")

FasterSpeeding:
also another thought, could a backport of the latest unicodedata work for this (e.g. https://pypi.org/project/unicodedata2/)?

davfsa:
> also another thought, could a backport of the latest unicodedata work for this (e.g. https://pypi.org/project/unicodedata2/)?

It would, but this way is something that I have been planning for a while as we now actually verify that the emojis are correct, instead of just assuming people pass the correct values. This can also help people have some sort of emoji check for commands, as it will fail to parse it if its an invalid emoji.

FasterSpeeding:
> > also another thought, could a backport of the latest unicodedata work for this (e.g. https://pypi.org/project/unicodedata2/)?
> 
> It would, but this way is something that I have been planning for a while as we now actually verify that the emojis are correct, instead of just assuming people pass the correct values. This can also help people have some sort of emoji check for commands, as it will fail to parse it if its an invalid emoji.

wasn't that why UnicodeEmoji.parse had validation in the first place?

davfsa:
> wasn't that why UnicodeEmoji.parse had validation in the first place?

That validation didn't check if it was a valid unicode emoji, let alone a valid unicode emoji. It just checked if all the characters were valid unicode

FasterSpeeding:
```py
import asyncio

import unicodedata
from hikari import emojis
from hikari.impl import rest as rest_impl


CHANNEL_ID: int =
MESSAGE_ID: int = 
FOO = "\N{WAVING WHITE FLAG}"
TOKEN: str = 


async def main():
    for i, codepoint in enumerate(FOO, start=1):
        print(unicodedata.name(codepoint))
        print(emojis.is_valid_discord_emoji(FOO))

    print(emojis.is_valid_discord_emoji(FOO))
    async with rest_impl.RESTApp().acquire(TOKEN, token_type="Bot") as cli:
        await cli.add_reaction(CHANNEL_ID, MESSAGE_ID, FOO)

asyncio.run(main())
```
Output: ```py
WAVING WHITE FLAG
False
False
```
To be honest this kind of feels like overcomplication that goes against the whole "light" aspect of this lib. In the few cases I can think of where people would want to make sure they're using/getting "valid discord emojis" (right now just unicode/Emoji 13.0 emojis) they'll end getting an error down the line from discord anyway and I don't really see any good way to keep up to date with the unicode version that discord is on as is (they barely publicise or announce when they update, reverse engineering the client to check which emojis it has declared in a way that's forwards compatible would be more effort than it's worth and just manually following and checking each client update seems like too much effort with using the raw client funnily enough likely not directly matching up with what the REST API allows and disses out). Plus accounting for any special casing Discord may introduce or remove for emojis in the future also seems like a difficult task if it's even possible (right now this likely isn't compatible with how discord kept backwards compatibility for their weird use of variation selector 16 in places as a random off the top of my head example). I've not actually seen any lists that keep up to date with the unicode version discord's doing properly with even the list we currently rely on being manually updated and lagging behind by something between 3 months to half a year so...
Anyway, the example I sent above uses an emoji that's supposedly "invalid" but is accepted (and returned) by the REST API.

FasterSpeeding:
Anyway, why does unicode_names need to be removed when it works perfectly fine on python 3.9 and there are backports which can make it work fine with 3.8

davfsa:
> Anyway, the example I sent above uses an emoji that's supposedly "invalid" but is accepted (and returned) by the REST API.

This is actually cause the client follows a separate emoji set from the server, which seems to normalize them using https://github.com/davfsa/discord_emojis/blob/master/discord_emojis/utils.py#L24 (will push an new commit here in a bit)

I don't see how having this as an external package defeats the point of "light" to the lib. This is the same thing as having a backport of unicodedata, although it actually verifies for the user that the unicode emoji they are using is valid and not just a normal string.

As per the update problem you mentioned, unicode emojis don't update regularly, nor does Discord. We can always keep an eye out and ask emzi to update it if he missed an update.

And just wanted to add, the emoji you sent in the example is valid by REST, but it will never return it. I believe this update should cover all edge cases

FasterSpeeding:
> I don't see how having this as an external package defeats the point of "light" to the lib. This is the same thing as having a backport of unicodedata, although it actually verifies for the user that the unicode emoji they are using is valid and not just a normal string.

why does this emoji validation have to be a part of hikari's core that brings in another dependency rather than a separate utility lib for the niche amount of ppl who can't rely on Discord to tell them which emojis are valid for some unexplained reason

FasterSpeeding:
> This is the same thing as having a backport of unicodedata

Last time i checked version specific dependencies are a thing in setuptools though

davfsa:
> why does this emoji validation have to be a part of hikari's core that brings in another dependency rather than a separate utility lib for the niche amount of ppl who can't rely on Discord to tell them which emojis are valid for some unexplained reason

If we are going to have an UnicodeEmoji object with validation, I would rather have proper validation. The reason for this is not only for Discord, it is also for `.url` and supporting re-uploading it. This way we can ensure that all emojis passed into that object will be able to be re-uploaded as an image, and not result in a nasty error.

> also last time I checked python version specific dependencies are a thing in setuptools though

That is not what I meant, what I mean it is that it is another dependency that will not actually provide what we want.

davfsa:
I'm going to change this pr into a draft for emoji validation. I would like to merge the other open prs and continue development without blocking everything for this pr.

PR #509, opened 2021-03-28
OAuth2 feature: endpoint logic and REST cli internal token strategy

FasterSpeeding:
### Summary
OAuth2 feature: endpoint logic and REST cli internal token strategy

### Checklist


davfsa:
Looks really nice. I will have a look later 👍

README.md
```diff
@@ -121,10 +121,6 @@ async def print_my_user(token):
 asyncio.run(print_my_user("user token here"))
```
davfsa, line 121: Should we update this codeblock to use the new `ClientCredentialsStrategy`?

hikari/applications.py
```diff
@@ -560,8 +567,104 @@ class AuthorizationInformation:
...
+    """Token types used within Hikari clients."""
+
+    BOT = "Bot"
+    """Bot token type."""
+
+    BASIC = "Basic"
+    """OAuth2 basic token type."""
+
+    BEARER = "Bearer"
+    """OAuth2 bearer token type."""
```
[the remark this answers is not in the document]
davfsa: Consistency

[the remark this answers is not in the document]
FasterSpeeding: With what also what're you referring to here?

hikari/impl/rest.py
```diff
@@ -280,8 +365,8 @@ def proxy_settings(self) -> config.ProxySettings:
 
     def acquire(
         self,
-        token: typing.Optional[str] = None,
-        token_type: str = _BEARER_TOKEN_PREFIX,
+        token: typing.Union[str, rest_api.TokenStrategy, None] = None,
+        token_type: typing.Union[str, applications.TokenType] = applications.TokenType.BEARER,
     ) -> RESTClientImpl:
         loop = asyncio.get_running_loop()
```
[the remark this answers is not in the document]
FasterSpeeding: And nope since we'd need more info (i.e client id + secret and the refresh token) to handle most tokens as a strategy

README.md
FasterSpeeding: Don't think it'd quite fit there

hikari/applications.py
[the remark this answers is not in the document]
davfsa: I'm sorry, was on very low hours of sleep and didn't know what I was talking about. Ignore this

hikari/applications.py
```diff
@@ -560,8 +567,104 @@ class AuthorizationInformation:
...
+    """
+
+
+@attr_extensions.with_copy
+@attr.s(eq=True, hash=True, init=True, kw_only=True, slots=True, weakref_slot=False)
+class OAuth2ImplicitToken(PartialOAuth2Token):
+    """Model for the OAuth2 token data returned by the implicit grant flow."""
+
+    state: typing.Optional[str] = attr.ib(eq=False, hash=False, repr=False)
+    """State parameter that was present in the authorization request if provided."""
```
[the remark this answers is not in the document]
davfsa: The problem now is that applications are now a different thing (eg, bots). I think it would be best to separate them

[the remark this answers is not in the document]
FasterSpeeding: Well most of the objects and enums in that module are only returned by oauth2 endpoints (pretty sure ApplicationFlags is the only exception since you get that on identify).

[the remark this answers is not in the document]
FasterSpeeding: Also fwiw an application in the context of Discord is an oauth2 application which can optionally have features like a bot user or "embedded" functionality etc etc attached to it.

README.md
davfsa: Why not?

hikari/impl/rest.py
```diff
@@ -133,6 +134,115 @@ def acquire(self) -> aiohttp.BaseConnector:
...
+            self._token = f"{response.token_type} {response.access_token}"
+            return self._token
+
+    async def close(self) -> None:
+        return None
+
+    def invalidate(self, token: typing.Optional[str]) -> None:
+        if token == self._token:
+            self._expire_at = 0.0
+            self._token = None
```
davfsa, line 243: The docstring for this function says that if `None` is passed, it will be invalidated regardless, which is not being done here

davfsa, line 243: Should this also actually call the endpoint to invalidate the token?

hikari/impl/rest.py
```diff
@@ -133,6 +134,115 @@ def acquire(self) -> aiohttp.BaseConnector:
         return self.connector
 
 
+class ClientCredentialsStrategy(rest_api.TokenStrategy):
```
davfsa, line 137: I initially thought that this would handle authorizing and refreshing user tokens through OAuth. I think I'm confused cause this seem to do something that I have not seen before nor can find documented on the API docs. Would you mind explaining it to me?

Also, if this indeed doesn't handle user OAuth, would a class for that be worth implementing or we re-add the comment to use `aiohttp-oauth`?

hikari/impl/rest.py
```diff
@@ -379,7 +489,7 @@ class RESTClientImpl(rest_api.RESTClient):
     token : hikari.undefined.UndefinedOr[builtins.str]
         The bot or bearer token. If no token is to be used,
         this can be undefined.
-    token_type : hikari.undefined.UndefinedOr[builtins.str]
+    token_type : typing.Union[builtins.str, hikari.applications.TokenType]
```
davfsa, line 492: Missing `builtins.None` in this type

hikari/impl/rest.py
```diff
@@ -379,7 +489,7 @@ class RESTClientImpl(rest_api.RESTClient):
     token : hikari.undefined.UndefinedOr[builtins.str]
```
davfsa, line 489:
```suggestion
    token : typing.Union[builtins.str, builtins.None, hikari.api.rest.TokenStrategy]
```

hikari/impl/rest.py
```diff
@@ -441,15 +551,15 @@ def __init__(
...
         self._lock = asyncio.Lock()
 
-        if token is None:
-            full_token = None
-        else:
-            if token_type is None:
-                token_type = _BOT_TOKEN_PREFIX
+        self._token: typing.Union[str, rest_api.TokenStrategy, None]
+        if isinstance(token, str):
+            self._token = f"{token_type.title()} {token}" if token_type else token
```
davfsa, line 556: I don't this we should allow passing the token type into `token`

FasterSpeeding: token type can't be None anymore

hikari/impl/rest.py
FasterSpeeding: nope

hikari/impl/rest.py
```diff
@@ -133,6 +134,115 @@ def acquire(self) -> aiohttp.BaseConnector:
...
+
+                raise
+
+            # Expires in is lowered a bit in-order to lower the chance of a dead token being used.
+            self._expire_at = time.monotonic() + math.floor(response.expires_in.total_seconds() * 0.99)
+            self._token = f"{response.token_type} {response.access_token}"
+            return self._token
+
+    async def close(self) -> None:
+        return None
```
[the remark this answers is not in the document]
FasterSpeeding: I don't see any benefit to doing that in the case of a client credentials token specifically

hikari/impl/rest.py
davfsa: You are still allowing passing `token_type` as `None` and checking if its `None` here :thinking:

FasterSpeeding: > I don't this we should allow passing the token type into token
BTW what do you mean by this

davfsa: You are allowing `token_type` being passed as `None` and, if its not there and the token is a string, taking the token as it is. This means that the `token_type` is already in the `token`, which I don't think is something we should allow

hikari/impl/rest.py
FasterSpeeding: That's the client credentials code grant and its very well documented. This pr introduces all the logic needed for authenticating and refreshing oauth2 access grant, extended bot auth, client credentials, and access tokens and won't handle stuff like storing the token in a database and providing an auth page for initiating the oauth2 grant since those are out of scope and highly dependent on the env of the project being developed. The token strategy is barely an oauth2 feature, it's initial intended use is to allow a rest client to make requests as an application while authorising using the client credentials grant while regenerating said auth whenever negating the need to rebuild the whole client and was designed to be flexible enough to let users slot in their own logic for generating the token as they please; this is a feature that's primarily intended for bot-less slash commands.

davfsa: Oh, nice. First time I see that page in the documentation :sweat_smile:

hikari/impl/rest.py
FasterSpeeding: well we don't get any extra functionality from enforcing that token type is passed separately and the old behaviour would've confused ppl who were passing the type as a part of the token when building a rest client since it'd also append a bot token to it making it invalid

README.md
FasterSpeeding: since it's for an awfully specific oauth2 flow

hikari/impl/rest.py
```diff
@@ -96,8 +99,6 @@
 _APPLICATION_JSON: typing.Final[str] = "application/json"
 _APPLICATION_OCTET_STREAM: typing.Final[str] = "application/octet-stream"
```
davfsa, line 100: While you are here, a bit of an unrelated change. Could you make these `sys.intern()`?

davfsa (approved):

hikari/impl/rest.py
FasterSpeeding: Do we need to intern constant values?

davfsa: If we use them a lot, it is probably best (in this case these are used quite a lot). It gives it a faster lookup time and a bit more efficient

FasterSpeeding: Seems unnecessary to tag that into random prs though and cpython should be interning strings implicitly anyway

PR #516, opened 2021-04-04
Switch python3.10 to dev version

davfsa:
### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, macos-10.15, windows-2019]
-        python-version: [3.8, 3.9, 3.10.0-alpha.4]
+        python-version: [3.8, 3.9, 3.10-dev]
         image: [null]
         include:
           - os: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


FasterSpeeding:
The tests just aren't gonna run until https://github.com/python/typed_ast/pull/155 if we bump to 3.10.0-apha6 or above, is that really what we want?

davfsa:
> The tests just aren't gonna run until https://github.com/python/typed_ast/pull/155 if we bump to 3.10.0-apha6 or above, is that really what we want?

That bug got fixed on master and could be fixed at any time, which is why I would have this one merged so we can just have pipelines running to try and pass on 3.10

FasterSpeeding:
> > The tests just aren't gonna run until [python/typed_ast#155](https://github.com/python/typed_ast/pull/155) if we bump to 3.10.0-apha6 or above, is that really what we want?
> 
> That bug got fixed on master and could be fixed at any time, which is why I would have this one merged so we can just have pipelines running to try and pass on 3.10

Could we just temporarily target a specific commit on their master until they release the change?

davfsa:
https://github.com/aio-libs/aiohttp/issues/5604

I'm thinking on just not installing `aiodns` on Python3.10. @FasterSpeeding opinions?

FasterSpeeding:
Would it be worth it opening an issue here to track aiodns on 3.10 even if it ends up being a long hanging one just so we dont forget about it?

PR #519, opened 2021-04-06
Add intent requirement to only be enabled in dashboard

davfsa:
### Checklist

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -3865,8 +3865,12 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             itself will not raise anything.
 
         !!! warning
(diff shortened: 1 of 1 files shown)


hikari/api/rest.py
```diff
@@ -3865,7 +3865,8 @@ def fetch_members(
             itself will not raise anything.
 
         !!! warning
-            This endpoint requires the `GUILD_MEMBERS` intent. Alternatively,
+            This endpoint requires the `GUILD_MEMBERS` intent to be enabled in
```
FasterSpeeding, line 3868: What does "authenticated with it" mean in this context when authenticating with an intent just doesn't exist on rnst?

davfsa: Updated

FasterSpeeding (approved):

PR #526, opened 2021-04-07
Implement invite "target_application" and rename "target_user_type" to "target_type"

davfsa:
### Checklist

### DIFF
--- a/hikari/applications.py
+++ b/hikari/applications.py
@@ -476,8 +476,11 @@ class Application(guilds.PartialApplication):
     Will be `builtins.None` if this application doesn't have a bot.
     """
 
-    owner: users.User = attr.ib(eq=False, hash=False, repr=True)
-    """The application's owner."""
+    owner: typing.Optional[users.User] = attr.ib(eq=False, hash=False, repr=True)
+    """The application's owner.
+
+    Will be `builtins.None` if the application is owned by Discord.
+    """
 
     rpc_origins: typing.Optional[typing.Sequence[str]] = attr.ib(eq=False, hash=False, repr=False)
     """A collection of this application's RPC origin URLs, if RPC is enabled."""
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -147,7 +147,8 @@ class _InviteFields:
     channel_id: snowflakes.Snowflake = attr.ib()
     inviter: typing.Optional[user_models.User] = attr.ib()
     target_user: typing.Optional[user_models.User] = attr.ib()
-    target_user_type: typing.Union[invite_models.TargetUserType, int, None] = attr.ib()
+    target_application: typing.Optional[application_models.Application] = attr.ib()
+    target_type: typing.Union[invite_models.TargetType, int, None] = attr.ib()
     approximate_active_member_count: typing.Optional[int] = attr.ib()
     approximate_member_count: typing.Optional[int] = attr.ib()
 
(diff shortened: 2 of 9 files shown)


hikari/impl/entity_factory.py
```diff
@@ -1469,8 +1470,9 @@ def _set_invite_attributes(self, payload: data_binding.JSONObject) -> _InviteFie
         else:
             channel_id = snowflakes.Snowflake(payload["channel_id"])
 
-        target_user_type = (
-            invite_models.TargetUserType(payload["target_user_type"]) if "target_user_type" in payload else None
+        target_type = invite_models.TargetType(payload["target_type"]) if "target_type" in payload else None
+        target_application = (
+            self.deserialize_application(payload["target_application"]) if "target_application" in payload else None
```
FasterSpeeding, line 1475: Isn't this a partial application object? Also isn't this feature unreleased rn?

davfsa: It has been documented already: https://github.com/discord/discord-api-docs/commit/1b4e363e324eb1f49a47e32cb0108fbe276c8e0e

And we currently don't have a proper PartialApplication object, so idk what they send and they don't

FasterSpeeding: > It has been documented already: discord/discord-api-docs@1b4e363

it's still unreleased though, wouldn't be the first time discord's (let someone) document stuff too early then had to remove it or made breaking changes and we do have a PartialApplication just it only exists for other application objects to subclass off of and isn't actually returned anywhere itself so it wouldn't be appropriate to just slap that here

davfsa: > it's still unreleased though, wouldn't be the first time discord's (let someone) document stuff too early then had to remove it or made breaking changes and we do have a PartialApplication just it only exists for other application objects to subclass off of and isn't actually returned anywhere itself so it probably wouldn't be appropriate to just slap that here

I meant not proper due to it not being exposed in the `__all__` of guilds.py. We can keep this pr on hold until Discord implements it

hikari/applications.py
```diff
@@ -476,8 +476,11 @@ class Application(guilds.PartialApplication):
     Will be `builtins.None` if this application doesn't have a bot.
     """
 
-    owner: users.User = attr.ib(eq=False, hash=False, repr=True)
-    """The application's owner."""
+    owner: typing.Optional[users.User] = attr.ib(eq=False, hash=False, repr=True)
+    """The application's owner.
+
+    Will be `builtins.None` if the application is owned by Discord.
```
FasterSpeeding, line 482: Is this ever actually the case or is this field just not included in this place for privacy reasons. (Do we really want to muddle up the interface of main objects to cover cases where a partial representation could also work better with a lot less fields)

davfsa: This field seems to be included in every single application except for the embedded ones. The only one I have tested this against is YouTube Together, which seems to be implemented by Discord.

I wasn't sure how to implement this without having to create a separate object that might break in the future.

FasterSpeeding: This field is only included in two of the application models and is rarely included in partial representations though

FasterSpeeding: Also making this object cover more cases than just full application representations will be just as likely to be broken down the line

davfsa:
Superseded by #631 to fix conflicting files

FasterSpeeding:
F

PR #532, opened 2021-04-12
Handle unknown channels

davfsa:
### Checklist

### DIFF
--- a/hikari/audit_logs.py
+++ b/hikari/audit_logs.py
@@ -435,7 +435,7 @@ class UnrecognisedAuditLogEntryInfo(BaseAuditLogEntryInfo):
     def __str__(self) -> str:
         attributes = ", ".join((f"{name}={value!r}" for name, value in self.__dict__.items()))
 
-        return f"UnrecognisedAuditLogEntryInfo({attributes})"
+        return f"{self.__class__.__name__}({attributes})"
 
     __repr__ = __str__
 
--- a/hikari/channels.py
+++ b/hikari/channels.py
@@ -30,6 +30,7 @@ __all__: typing.List[str] = [
     "PermissionOverwrite",
     "PermissionOverwriteType",
     "PartialChannel",
+    "UnrecognisedChannel",
     "TextChannel",
     "PrivateChannel",
     "DMChannel",
@@ -812,3 +813,36 @@ class GuildVoiceChannel(GuildChannel):
 
     If this is `0`, then assume no limit.
     """
+
+
+class UnrecognisedChannel(PartialChannel):
+    """Channels that haven't been implemented in the library.
+
+    The attributes on this object are undocumented and dynamic.
+
+    Example
+    -------
+        >>> entry_info.foobar.baz
+
+    !!! note
+        This model has no slots and will have arbitrary undocumented attributes
+        (in it's `__dict__` based on the received payload).
+        The only attributes that are always garantied to be there are `app`,
+        `id`, `name` and `type`.
+    """
+
+    def __init__(
+        self,
+        /,
+        *,
+        app: traits.RESTAware,
+        id: snowflakes.Snowflake,
+        name: typing.Optional[str],
+        type: typing.Union[ChannelType, int],
+        payload: typing.Mapping[str, str],
+    ) -> None:
+        self.__dict__.update(payload)
+        self.app = app
+        self.id = id
+        self.name = name
+        self.type = type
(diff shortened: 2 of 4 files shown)


hikari/audit_logs.py
```diff
@@ -406,6 +406,7 @@ async def fetch_channel(self) -> channels.GuildVoiceChannel:
         return channel
 
 
+@attr.s(eq=False, hash=False, init=False, slots=False, str=False, repr=False)
```
FasterSpeeding, line 409: What's the point of this?

hikari/channels.py
```diff
@@ -812,3 +813,37 @@ class GuildVoiceChannel(GuildChannel):
...
+
+    Example
+    -------
+        >>> entry_info.foobar.baz
+
+    !!! note
+        This model has no slots and will have arbitrary undocumented attributes
+        (in it's `__dict__` based on the received payload).
+        The only attributes that are always garantied to be there are `app`,
+        `id`, `name` and `type`.
```
FasterSpeeding, line 832: What about a nullable guild id attribute

hikari/channels.py
```diff
@@ -812,3 +813,37 @@ class GuildVoiceChannel(GuildChannel):
...
+    def __init__(
+        self,
+        /,
+        *,
+        app: traits.RESTAware,
+        id: snowflakes.Snowflake,
+        name: typing.Optional[str],
+        type: typing.Union[ChannelType, int],
+        payload: typing.Mapping[str, str],
+    ) -> None:
```
FasterSpeeding, line 844: Do we really want to use this style here? Wouldn't this have memory implications for the cache

hikari/impl/entity_factory.py
```diff
@@ -728,10 +728,15 @@ def deserialize_channel(
...
-        if channel_model := self._guild_channel_type_mapping.get(channel_type):
-            return channel_model(payload, guild_id=guild_id)
+        if guild_channel_model := self._guild_channel_type_mapping.get(channel_type):
+            return guild_channel_model(payload, guild_id=guild_id)
 
-        return self._dm_channel_type_mapping[channel_type](payload)
+        if dm_channel_model := self._dm_channel_type_mapping.get(channel_type):
+            return dm_channel_model(payload)
+
+        return channel_models.UnrecognisedChannel(
```
FasterSpeeding, line 737: Would it be easier to just fall back to a partial channel and log that it's an unexpected channel type was received?

davfsa: This is exactly what this is doing. Forgot the logging part tho, so will add

hikari/channels.py
davfsa: In what way?

hikari/channels.py
davfsa: I rather keep it with only the same attributes as Partial Channel

hikari/audit_logs.py
davfsa: Was testing something out, forgot to remove it

hikari/channels.py
FasterSpeeding: Ignore the 2nd part but whole this style is used in the audit log entry as its a compromise between handling unimplemented cases while not throwing away all useful data as there isn't any base partial data so an unknown type would have no attributes otherwise do we really want to introduce that behaviour in channels? Channels are a more core part of the lib and applying this approach here seems kinda questionable as its no longer a case of otherwise there'd just be no data and it would technically make all new channel type impls breaking since previous behaviour for that type would've likely had different attribute types and names.

davfsa: I would like to provide this as a nicer alternative than having `KeyError`s being raised everywhere. This at least provides a semi-useful object you can use to make some generic checks. An example of this would be so you can detect the channel create on recently implemented channels without having to update the library

FasterSpeeding: Could just return a partial channel or a partial channel that's been extended to include an optional guild id attribute

davfsa: This is what it is doing. The reason why I don't want to implement an optional guild_id is because I would like to keep the attributes provided by PartialChannel

FasterSpeeding: Not quite, what you're doing is returning an object with an arbitrary amount of attributes, if you really want to return a the data it'd be better to just have a other data dict attribute but I don't really see the point of providing all the data when this only exists to handle the case of an unexpected channel type softly and is intended to be ignored not implemented around

FasterSpeeding:
Also this would still break all the cases of `assert isinstance(channel, channel_models.GuildChannel)` we have lying around in the entity factory deserialize

hikari/channels.py
FasterSpeeding: Also when i say a partial channel i mean literal PartialChannel plus also GuildChannel since need to handle that case as well

davfsa:
> Also this would still break all the cases of `assert isinstance(channel, channel_models.GuildChannel)` we have lying around in the entity factory deserialize if it was a new guild channel type

I was running my bot with `-O` which is why I wasn't getting the assertion errors. Furthermore, this seems to also break with the cache, so will have to rethink a better way of doing this

PR #545, opened 2021-04-16
Handle unknown entities

davfsa:
### Checklist

### DIFF
--- a/hikari/api/entity_factory.py
+++ b/hikari/api/entity_factory.py
@@ -580,10 +580,12 @@ class EntityFactory(abc.ABC):
 
         Raises
         ------
-        KeyError
+        builtins.KeyError
             If `guild_id` is left as `hikari.undefined.UNDEFINED` when
             `"guild_id"` is not present in the passed payload of a guild
             channel.
+        hikari.errors.UnrecognisedEntityError
+            If the channel type is unknown.
         """
 
     ################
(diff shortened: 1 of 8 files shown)


hikari/api/entity_factory.py
```diff
@@ -103,7 +103,12 @@ class GatewayGuildDefinition:
 
 
 class EntityFactory(abc.ABC):
-    """Interface for components that serialize and deserialize JSON payloads."""
+    """Interface for components that serialize and deserialize JSON payloads.
+
+    !!! note
+        All functions in this class may raise `hikari.errors.UnrecognisedEntityError`
```
FasterSpeeding, line 109: Would they though? doesn't this only really apply to a few methods where this behaviour could be explicitly documented for them

davfsa: Since this is the API, we don't know how each implementation will want to raise it or not.

FasterSpeeding: depending on how "unrecognised entity" is interpreted this could be suggests that implementations should suppress all encountered errors and re-raise a UnrecognisedEntityError for all deserialization

FasterSpeeding: if it isn't meant to just cover cases of methods which dynamically map payloads to different types then the docs should show this but if it is then there isn't any reason why it can't just be documented on those methods

FasterSpeeding (approved):

PR #556, opened 2021-04-21
Move away from distutils as it has been deprecated

davfsa:
https://www.python.org/dev/peps/pep-0632/

### Checklist

### DIFF
--- a/hikari/internal/ux.py
+++ b/hikari/internal/ux.py
@@ -25,7 +25,6 @@ from __future__ import annotations
 
 __all__: typing.List[str] = ["init_logging", "print_banner", "supports_color", "HikariVersion", "check_for_updates"]
 
-import distutils.version
 import importlib.resources
 import logging.config
 import os
@@ -44,6 +43,8 @@ from hikari.internal import net
 
 if typing.TYPE_CHECKING:
     from hikari import config
+
+    CmpTuple = typing.Tuple[int, int, int, typing.Union[int, float]]
 
 # While this is discouraged for most purposes in libraries, this enables us to
 # filter out the vast majority of clutter that most network logger calls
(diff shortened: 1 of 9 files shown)


hikari/internal/ux.py
```diff
@@ -253,45 +254,62 @@ def supports_color(allow_color: bool, force_color: bool) -> bool:
...
+
+    def __gt__(self, other: typing.Any) -> bool:
+        return self._compare(other, lambda s, o: s > o)
+
+    def __ge__(self, other: typing.Any) -> bool:
+        return self._compare(other, lambda s, o: s >= o)
+
+    def _compare(self, other: typing.Any, method: typing.Callable[[CmpTuple, CmpTuple], bool]) -> bool:
+        if not isinstance(other, HikariVersion):
+            return NotImplemented
```
FasterSpeeding, line 310: Could this not just be False?

davfsa: In the context this is being used in, returning NotImplemented makes it fallback to the overwritten functionality. In other words, it is the same as doing `super().__whatever__(other)`

hikari/internal/ux.py
```diff
@@ -253,45 +254,62 @@ def supports_color(allow_color: bool, force_color: bool) -> bool:
     return color_support
 
 
-class HikariVersion(distutils.version.StrictVersion):
-    """Hikari-compatible strict version."""
+_VERSION_REGEX: typing.Final[re.Pattern[str]] = re.compile(r"^(\d+)\.(\d+)\.(\d+)(\.[a-z]+)?(\d+)?$", re.I)
```
FasterSpeeding, line 257: isn't re.Pattern not subscriptable on 3.8 (maybe typing.Pattern might be better for backwards compat here)?

davfsa: `typing.Pattern` has been deprecated as of Python 3.9 in favour of `re.Pattern`. I have tested it and 3.8 seems to support `re.Pattern`

FasterSpeeding: did you test that on 3.8

davfsa: Yep, my dev enviroment runs python 3.8.

FasterSpeeding:
```py
Python 3.8.7 (tags/v3.8.7:6503f05, Dec 21 2020, 17:59:51) [MSC v.1928 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license()" for more information.
>>> import re
>>> re.Pattern[str]
  File "<pyshell#1>", line 1, in <module>
    re.Pattern[str]
TypeError: 'type' object is not subscriptable
```
also the docs state that typing.Pattern was deprecated in 3.9 because re.Pattern was made generic in 3.9, just like every other generic alias in typing like Mapping, MutableMapping, Sequence, MutableSequence, AbstractSet, Set, etc etc

davfsa: Oh, im guessing its `from __future__ import annotations` that is making it ignore it then. Because I have never had that error

FasterSpeeding: well we should be using typing.Pattern and if we were really using this approach consistently we'd be using collections.abc and the standard types (e.g. type, tuple) in place of the typing aliases.

davfsa: Thanks for spotting this issue :)

hikari/internal/compatibility.py
```diff
@@ -0,0 +1,47 @@
...
+# Before Python 3.9, you had to use `typing.Pattern` to mark a regex pattern.
+# This was later moved to the `re` module under `re.Pattern` and deprecated
+# `typing.Pattern`.
+#
+# Provides backwards compatibility for: Python 3.8.0 - 3.9.0
+
+if CURRENT_PYTHON_VERSION <= (3, 9, 0):
+    Pattern = typing.Pattern
+else:
+    Pattern = re.Pattern
```
FasterSpeeding, line 47: Do we even need to do this when the typing stuff hasn't been removed or should we also love stuff like typing.(Mutable)Mapping, typing.Dict, typing.Type, typing.(Mutable)Seqience, typing.(Abstract)Set, etc etc here as well

davfsa: If it's marked in the python documentation as `Deprecated as of {version}` and what it has been deprecated for is not in all python versions, then yes

FasterSpeeding: Not at anytime, they specify a timeline for it and even mention it in the docs and this change covers half of the typing module it's not worth the effort to change all of them to this when by the time it's removed well have removed support for 3.8

davfsa: Ah, ok then :+1:

FasterSpeeding: Most of the time when stuff is deprecated there's a timeline given for when it'll be removed, that's just the same way to do stuff

FasterSpeeding (approved):

PR #558, opened 2021-04-23
Improve request rate limiting

davfsa:
- Move lock to per-bucket lock
- Don't acquire global_rate_limit when not using authentication

### Checklist

### DIFF
--- a/hikari/impl/buckets.py
+++ b/hikari/impl/buckets.py
@@ -128,7 +128,7 @@ These headers are:
     limiting occurs in the current window.
 * `X-RateLimit-Bucket`:
     a `builtins.str` containing the initial bucket hash.
-* `x-ratelimit-reset-after`:
+* `X-RateLimit-Reset-After`:
     a `builtins.float` containing the number of seconds when the current rate
     limit bucket will reset with decimal millisecond precision.
 
@@ -166,7 +166,7 @@ on what attributes you send in a JSON or form data payload.
 No information is sent in headers about these specific limits. You will only
 be made aware that they exist once you get ratelimited. In the 429 ratelimited
 response, you will have the `"global"` attribute set to `builtins.False`, and a
-`"reset_after"` attribute that differs entirely to the `x-ratelimit-reset-after`
+`"reset_after"` attribute that differs entirely to the `X-RateLimit-Reset-After`
 header. Thus, it is important to not assume the value in the 429 response
 for the reset time is the same as the one in the bucket headers. Hikari's
 `hikari.api.rest.RESTClient` implementation specifically uses the value furthest
@@ -180,19 +180,6 @@ At the time of writing, the only example of this appears to be on the
 `PATCH /channels/{channel_id}` endpoint. This has a limit of two changes per
 10 minutes. More details about how this is implemented have yet to be
 released or documented...
-
-
-Caveats
--------
-
-These implementations rely on Discord sending consistent buckets back to us.
-
-This also begins to crumble if more than one HTTP client is in use, since
-there is no performent way to communicate shared rate limits between
-distributed applications. The general concept to follow is that if you are
-making repeated API calls, or calls that are not event-based (e.g.
-fetching messages on a timer), then this can be considered a form of API abuse
-and should be used sparingly.
 """
 
 from __future__ import annotations
@@ -205,7 +192,6 @@ import typing
 
 from hikari import errors
 from hikari.impl import rate_limits
-from hikari.internal import aio
 from hikari.internal import routes
 from hikari.internal import time
 from hikari.internal import ux
@@ -241,37 +227,64 @@ class RESTBucket(rate_limits.WindowedBurstRateLimiter):
     which allows dynamically changing the enforced rate limits at any time.
     """
 
-    __slots__: typing.Sequence[str] = ("compiled_route",)
+    __slots__: typing.Sequence[str] = ("_compiled_route", "_max_rate_limit", "_lock")
 
-    compiled_route: typing.Final[routes.CompiledRoute]
-    """The compiled route that this rate limit is covering."""
-
-    def __init__(self, name: str, compiled_route: routes.CompiledRoute) -> None:
+    def __init__(self, name: str, compiled_route: routes.CompiledRoute, max_rate_limit: float) -> None:
         super().__init__(name, 1, 1)
-        self.compiled_route = compiled_route
+        self._compiled_route = compiled_route
+        self._max_rate_limit = max_rate_limit
+        self._lock = asyncio.Lock()
+
+    async def __aenter__(self) -> None:
+        await self.acquire()
+
+    async def __aexit__(
+        self,
+        exc_type: typing.Optional[typing.Type[BaseException]],
+        exc: typing.Optional[BaseException],
+        exc_tb: typing.Optional[types.TracebackType],
+    ) -> None:
+        self._lock.release()
 
     @property
     def is_unknown(self) -> bool:
         """Return `builtins.True` if the bucket represents an `UNKNOWN` bucket."""
         return self.name.startswith(UNKNOWN_HASH)
 
-    def acquire(self) -> asyncio.Future[None]:
+    async def acquire(self) -> None:
         """Acquire time on this rate limiter.
 
         !!! note
             You should afterwards invoke `RESTBucket.update_rate_limit` to
             update any rate limit information you are made aware of.
 
-        Returns
-        -------
-        asyncio.Future[builtins.None]
-            A future that should be awaited immediately. Once the future completes,
-            you are allowed to proceed with your operation.
+        Raises
+        ------
+        hikari.errors.RateLimitTooLongError
+            If the rate limit is longer than `max_rate_limit`.
         """
-        return aio.completed_future(None) if self.is_unknown else super().acquire()
+        await self._lock.acquire()
+
+        if self.is_unknown:
+            return
+
+        now = time.monotonic()
+        retry_after = self.reset_at - now
+
+        if self.is_rate_limited(now) and retry_after > self._max_rate_limit:
+            raise errors.RateLimitTooLongError(
+                route=self._compiled_route,
+                retry_after=retry_after,
+                max_retry_after=self._max_rate_limit,
+                reset_at=self.reset_at,
+                limit=self.limit,
+                period=self.period,
+            )
+
+        await super().acquire()
 
     def update_rate_limit(self, remaining: int, limit: int, reset_at: float) -> None:
-        """Amend the rate limit.
+        """Update the rate limit information.
 
         Parameters
         ----------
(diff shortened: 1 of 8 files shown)


hikari/impl/buckets.py
```diff
@@ -281,12 +276,19 @@ def update_rate_limit(self, remaining: int, limit: int, reset_at: float) -> None
...
+        force : builtins.bool
+            Whether to force updating parameters, ignoring checks.
 
         !!! note
             The `reset_at` epoch is expected to be a `time.monotonic_timestamp`
             monotonic epoch, rather than a `time.time` date-based epoch.
         """
-        self.remaining = remaining
+        # Only set remaining if its less than what we have stored internally.
+        # This is done because this function may be called with outdated values
```
FasterSpeeding, line 287: Won't this mess with the handling of endpoints which have multiple buckets also how will the stats be stored for an endpoint which just went past it's reset time and therefore self.remaining is less than the actual remaining. Also isn't this just as botched equivalent to locking per bucket?

davfsa: I forgot to account for those cases :sweat_smile:. I initially tried to implement per bucket locks, but there is really no "nice" way to do so. I could give it another try and see what I can do

davfsa: > Won't this mess with the handling of endpoints which have multiple buckets

Those are a weird mess which we can't really account for in any way.

> also how will the stats be stored for an endpoint which just went past it's reset time and therefore self.remaining is less than the actual remaining

Endpoints are updated internally when calling `is_ratelimited`, which is done when calling `acquire` on non `UNKNOWN` buckets.

davfsa: Changed to per-bucket lock :+1:

hikari/impl/buckets.py
```diff
@@ -303,6 +333,24 @@ def drip(self) -> None:
...
+        real_bucket_hash: builtins.str
+            The real bucket hash for this bucket.
+
+        Raises
+        ------
+        builtins.RuntimeError
+            If the hash of the bucket is already known.
+        """
+        if not self.is_unknown:
+            raise RuntimeError("Cannot resolve known bucket")
```
FasterSpeeding, line 350: Is there any point in resolving them if it doesn't ensure that ratelimitimg for a single bucket will be handled by and locked the main bucket object now that we’ve discovered the overlap

davfsa: When we make a request to an unknown bucket. This will handle that bucket under an unknown hash. Every other request to that same route will get stopped until the lock is released, after which we will (if nothing went wrong) know what the ratelimits are. We will then resolve the bucket and all new requests and existing requests waiting for the lock will go under that same bucket, respecting the ratelimits.

FasterSpeeding: And if multiple requests are locked to a unknown bucket and it then gets resolved to an already existing bucket youll now have two active request queues for the one bucket

davfsa: It should not resolve to an already existing bucket 🤔(unless im missing something)

FasterSpeeding: Multiple endpoints can have the same bucket hash

davfsa: Oh, ig I didn't account for those. Will fix

FasterSpeeding: This behaviour was being handled in the ver I made of this so could just use that as a reference

hikari/impl/buckets.py
```diff
@@ -590,18 +622,34 @@ def update_rate_limits(
                 limit_header,
                 remaining_header,
             )
-        else:
-            bucket = RESTBucket(real_bucket_hash, compiled_route)
+        except KeyError:
+            try:
+                unknown_bucket_hash = compiled_route.create_real_bucket_hash(UNKNOWN_HASH)
```
FasterSpeeding, line 627: You can't rely on the major param ids being unique, you'll get collisions between unknown ratelimit buckets for channels and guilds with the same id

FasterSpeeding: Not checked this yet but I suspect you'll l also get a ton of collisions between endpoints with no major params with this approach

hikari/impl/rest.py
```diff
@@ -811,7 +815,9 @@ async def _parse_ratelimits(self, compiled_route: routes.CompiledRoute, response
         # isn't some weird edge case here somewhere in Discord's implementation.
         # We can safely retry if this happens.
         if remaining <= 0:
-            _LOGGER.warning("rate limited on bucket %s, maybe you are running more than one bot on this token?", bucket)
+            _LOGGER.warning(
+                "rate limited on bucket %s, maybe you are running more than one bot on this token?", bucket or "None"
```
FasterSpeeding, line 819: This shouldn't be logged if bucket is None

davfsa: Best to always log it. If Discord change the header or have a bug on their side where they stop sending it, both us and bot developers will be confused on why their bots are getting banned for many 429's without any logs

FasterSpeeding: That's just misleading, it only covers bucket ratelimits and to say we were unexpected ratelimited on an endpoint where we can't pre-emptively ratelimit is just wrong. For the cases you just described it'd raise a ratelimited error

FasterSpeeding: That or None is still redundant though

davfsa: It has already been removed. This comment is outdated

hikari/impl/buckets.py
```diff
@@ -590,18 +626,34 @@ def update_rate_limits(
                 limit_header,
                 remaining_header,
             )
-        else:
-            bucket = RESTBucket(real_bucket_hash, compiled_route)
+        except KeyError:
+            try:
```
FasterSpeeding, line 630: just a general question but wouldn't it be cleaner and more explicit to do 
```py
if bucket := ...
```
here instead of the try and accept?

davfsa: try and catching the errors seems to be a bit faster then using a walrus and if/else statements

FasterSpeeding: and is it still faster when an error is raised

davfsa: Yep, both cases

FasterSpeeding: even with all that over head from error handling?

davfsa: Overhead is really performance hurtful when using `contextlib.suppress` (which I found some uses off and need to remove). A bare try/except doesn't seem to have any performance implication and seems to be even faster than using an if/else clause

FasterSpeeding: So you're fine with putting micro-optimisations over readability?

davfsa: This doesn't make the code super unreadable. But I can change it back to if/else statements if you think its better

FasterSpeeding: > This doesn't make the code super unreadable. 
it's more ambiguous as to what case catching KeyError is handling compared to if := since there's multiple function calls in the try section which could all be raising it

hikari/impl/buckets.py
```diff
@@ -303,6 +327,28 @@ def drip(self) -> None:
...
+            If the hash of the bucket is already known.
+        """
+        if not self.is_unknown:
+            raise RuntimeError("Cannot resolve known bucket")
+
+        self.name = real_bucket_hash
+
+
+def _create_unknown_hash(route: routes.CompiledRoute) -> str:
+    return UNKNOWN_HASH + routes.HASH_SEPARATOR + str(hash(route))
```
FasterSpeeding, line 350: What about major params?

hikari/impl/buckets.py
```diff
@@ -591,17 +620,32 @@ def update_rate_limits(
                 remaining_header,
             )
         else:
-            bucket = RESTBucket(real_bucket_hash, compiled_route)
+            unknown_bucket_hash = _create_unknown_hash(compiled_route)
+
+            if bucket := self.real_hashes_to_buckets.pop(unknown_bucket_hash, None):
```
FasterSpeeding, line 625: isn't the whole "real hashes" naming scheme now just a lie here

hikari/impl/buckets.py
```diff
@@ -506,50 +557,32 @@ def acquire(self, compiled_route: routes.CompiledRoute) -> asyncio.Future[None]:
...
         template = compiled_route.route
 
-        if template in self.routes_to_hashes:
+        try:
             bucket_hash = self.routes_to_hashes[template]
-        else:
+            real_bucket_hash = compiled_route.create_real_bucket_hash(bucket_hash)
+        except KeyError:
             bucket_hash = UNKNOWN_HASH
             self.routes_to_hashes[template] = bucket_hash
```
FasterSpeeding, line 574: Shouldn't the generated pseudo hash be used here not "UNKNOWN" since that'd never resolve to a valid bucket?

hikari/impl/buckets.py
```diff
@@ -303,6 +327,28 @@ def drip(self) -> None:
         if not self.is_unknown:
             self.remaining -= 1
 
+    def resolve(self, real_bucket_hash: str) -> None:
```
FasterSpeeding, line 330: isn't this function unnecessary in it's current state?

hikari/impl/buckets.py
```diff
@@ -267,8 +265,34 @@ def acquire(self) -> asyncio.Future[None]:
...
+        Raises
+        ------
+        builtins.RuntimeError
+            If the lock is not acquired.
+        hikari.errors.RateLimitTooLongError
+            If the ratelimit is longer than `max_rate_limit`.
         """
-        return aio.completed_future(None) if self.is_unknown else super().acquire()
+        if not self._lock.locked():
+            raise RuntimeError("Lock is not acquired")
```
FasterSpeeding, line 277: Multiple things.
1: just because the lock is locked doesn't mean this is the place where it's acquired?
2:  Doesn't this break the interface rules of the base ratelimiter since it makes the acquire method almost guaranteed to fail unless something else is also acquiring the ratelimiter with async with syntax at the same time.
3: shouldn't the acquire method be the thing which acquires the lock? isn't the whole async with syntax meant to be a short hand for what could otherwise be achieved through method calls rather than something which needs to directly used to achieve something.

hikari/impl/rate_limits.py
```diff
@@ -308,7 +307,7 @@ def __init__(self, name: str, period: float, limit: int) -> None:
         self.limit = limit
         self.period = period
 
-    def acquire(self) -> asyncio.Future[typing.Any]:
+    def acquire(self) -> asyncio.Future[None]:
```
FasterSpeeding, line 310: random point but the [docs](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future) for asyncio recommend not exposing Future objects in user-facing APIs.
> The rule of thumb is to never expose Future objects in user-facing APIs

hikari/impl/buckets.py
davfsa: The route already includes the major parameters

hikari/impl/buckets.py
davfsa: We can consider unknown hashes "real hashes". We can also store them in a separate dict if that is better with you. I think it will be harder to manage and having another dict for no reason is a waste of resources

hikari/impl/buckets.py
davfsa: Kinda. It is meant to be an interface to updating the buckets name without doing it directly

hikari/impl/rate_limits.py
davfsa: Ig. How else would you implement this?

hikari/impl/buckets.py
davfsa: 1. I don't think there is a way to check that they are both in the same place. The idea behind this is to not allow acquiring the bucket without acquiring the lock before.

2. Yea, kinda. What I disliked about your approach to this was having to change the whole system to match the interface of this specific object.

3. Acquire is a blocking method, meaning that it wont be able to acquire the asyncio lock in any way. Furthermore, using an async context manager allows us to keep the bucket locked until we are done updating it or we error out, in which cause it will just be released.

davfsa: Changed in the next commit

FasterSpeeding (approved):

PR #559, opened 2021-04-23
Fix pipelines to install wheel and setuptools

### LINKED ISSUE #537: aiodns cannot be compiled on python 3.10
Would it be worth it opening an issue here to track how to properly deal with aiodns on 3.10 in the long term even if it ends up being a long hanging one just so we dont forget about it?

_Originally posted by @FasterSpeeding in https://github.com/hikari-py/hikari/issues/516#issuecomment-818643043_

davfsa:
- Remove python version limit on aiodns

Closes #537

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,8 @@ jobs:
 
       - name: Test Installation
         run: |
+          pip install wheel
+
           pip install -e .
           pip uninstall -y hikari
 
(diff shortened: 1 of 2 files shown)


FasterSpeeding:
Should this close #537?

FasterSpeeding (approved):

davfsa:
Yep, it does, forgot to specify

PR #566, opened 2021-04-27
Remove impossible events

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/event_factory.py
+++ b/hikari/impl/event_factory.py
@@ -571,13 +571,7 @@ class EventFactoryImpl(event_factory.EventFactory):
                 message_id=message_id,
             )
 
-        # TODO: check if this can even occur.
-        return reaction_events.DMReactionDeleteAllEvent(
-            app=self._app,
-            shard=shard,
-            channel_id=channel_id,
-            message_id=message_id,
-        )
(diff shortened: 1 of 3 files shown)


hikari/impl/event_factory.py
```diff
@@ -571,13 +571,7 @@ def deserialize_message_reaction_remove_all_event(
...
             )
 
-        # TODO: check if this can even occur.
-        return reaction_events.DMReactionDeleteAllEvent(
-            app=self._app,
-            shard=shard,
-            channel_id=channel_id,
-            message_id=message_id,
-        )
+        raise NotImplementedError("DM reaction all removes are not documented behavior")
```
FasterSpeeding, line 574: Yes it is?

FasterSpeeding:
This just breaks forward compatability

hikari/impl/event_factory.py
davfsa: I haven't been able to trigger any of the two events 🤔

FasterSpeeding: They are documented behaviour though

davfsa: Hmmmm. We could just keep them ig

PR #577, opened 2021-05-02
Add base support for interactions + slash commands

FasterSpeeding:
### Summary
Add base support for interactions + slash commands.

### Checklist


FasterSpeeding:
Some stuff from old pr:
* Ability to go from command object to a populated builder or just accept pure command objects as well on bulk edit commands?
* Check nothing was lost in webhooks module (plus rest logic for editing messages) and mentions module during rebases

hikari/errors.py
```diff
@@ -103,6 +104,17 @@ class HikariInterrupt(KeyboardInterrupt, HikariError):
     """The signal name that was raised."""
 
 
+@attr.define(auto_exc=True, repr=False, weakref_slot=False)
+class ComponentAlreadyRunningError(HikariError):
```
FasterSpeeding, line 108: Just merge this with ComponentNotRunningError and rename it to "ComponentStateConflictError"

hikari/impl/special_endpoints.py
```diff
@@ -653,3 +656,123 @@ async def __anext__(self) -> audit_logs.AuditLog:
...
+@attr_extensions.with_copy
+@attr.s(kw_only=False, slots=True, weakref_slot=False)
+class CommandBuilder(special_endpoints.CommandBuilder):
+    """Standard implementation of `hikari.api.special_endpoints.CommandBuilder`."""
+
+    # Required arguments.
+    _name: str = attr.ib()
+    _description: str = attr.ib()
+
+    # TODO: does ID actually matter here?
```
FasterSpeeding, line 748: this one little TODO which almost got away

hikari/impl/entity_factory.py
```diff
@@ -1680,6 +1902,14 @@ def _deserialize_sticker(self, payload: data_binding.JSONObject) -> message_mode
             format_type=message_models.StickerFormatType(payload["format_type"]),
         )
 
+    def _deserialize_interaction(self, payload: data_binding.JSONObject) -> message_models.MessageInteraction:
```
FasterSpeeding, line 1905: Different name?

hikari/interactions/commands.py
```diff
@@ -0,0 +1,777 @@
...
+@attr_extensions.with_copy
+@attr.define(hash=False, kw_only=True, weakref_slot=False)
+class CommandInteractionOption:
+    """Represents the options passed for a command interaction."""
+
+    name: str = attr.field(repr=True)
+    """Name of this option."""
+
+    type: OptionType = attr.field(repr=True)
+    """Type of this option."""
```
FasterSpeeding, line 341: Should be union with int

hikari/internal/ed25519.py
```diff
@@ -0,0 +1,109 @@
...
+        _verify_key(public_key)
+        return _build_verifier(_ed25519.VerifyingKey(public_key).verify, _ed25519.BadSignatureError)
+
+    build_fast_ed25519_verifier = _build_fast_ed25519_verifier
+
+except ImportError:
+    build_fast_ed25519_verifier = None
+
+build_ed25519_verifier: typing.Final[VerifyBuilderT] = build_fast_ed25519_verifier or build_slow_ed25519_verifier
+"""Main implementation of `VerifyBuilderT` used within Hikari."""
```
davfsa, line 109: As I said before, could we just merge all these into 1?

```suggestion
def build_ed25519_verifier(public_key: bytes, /) -> VerifierT:
    """`VerifyBuilderT` implementation which will always be present."""
    _verify_key(public_key)
    return _build_verifier(_ed25519.VerifyingKey(public_key).verify, _ed25519.BadSignatureError)
```
and above (in the imports)
```py
try:
    import ed25519 as _ed25519 # type: ignore[import]
    
except ImportError:
    # Fallback to the pure python package if speedups not available
    from pure25519 import ed25519_oop as _ed25519 # type: ignore[import]
```

davfsa (commented):
I have still some files to go through, but ima comment these so we can start discussing them :)

davfsa:
Also, something I just remembered, we should probably also add a simple example under `examples/` and a little working script in the README

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,498 @@
...
+            reuse_address=reuse_address,
+            reuse_port=reuse_port,
+            socket=socket,
+            shutdown_timeout=shutdown_timeout,
+            ssl_context=ssl_context,
+        )
+
+        if close_passed_executor and self._executor:
+            self._executor.shutdown(wait=True)
+            self._executor = None
```
davfsa, line 404:
```suggestion
        try:
            self._server.run(
                asyncio_debug=asyncio_debug,
                backlog=backlog,
                close_loop=close_loop,
                coroutine_tracking_depth=coroutine_tracking_depth,
                enable_signal_handlers=enable_signal_handlers,
               host=host,
                port=port,
                path=path,
                reuse_address=reuse_address,
                reuse_port=reuse_port,
                socket=socket,
                shutdown_timeout=shutdown_timeout,
                ssl_context=ssl_context,
            )

        finally:
            if close_passed_executor and self._executor:
                self._executor.shutdown(wait=True)
                self._executor = None
```

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,498 @@
...
+        """Open this REST server and block until it closes.
+
+        Other Parameters
+        ----------------
+        asyncio_debug : builtins.bool
+            Defaults to `builtins.False`. If `builtins.True`, then debugging is
+            enabled for the asyncio event loop in use.
+        backlog : builtins.int
+            The number of unaccepted connections that the system will allow before
+            refusing new connections.
```
davfsa, line 325: Would this make more sense to rename to `simultaneous_connections`?

FasterSpeeding: this isn't about active connections, this about how many unaccepted connections will be queued in a backlog before it starts instantly failing new connections

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,558 @@
...
+        runner = self._runner
+        future = self._future
+        self._runner = None
+        await runner.shutdown()
+        await runner.cleanup()
+        future.set_result(None)
+
+    async def join(self) -> None:
+        if not self._runner:
+            raise errors.ComponentStateConflictError("Cannot wait for an inactive interaction server to join")
```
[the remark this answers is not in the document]
FasterSpeeding: This would infer that "joining" it actively does anything as opposed to being a passive operation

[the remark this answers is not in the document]
davfsa: > 
> 
> This would infer that "joining" it actively does anything as opposed to being a passive operation

Oh, true. I was thinking in the bot context. My bad

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,558 @@
...
+
+    def __init__(
+        self,
+        *,
+        dumps: aiohttp.typedefs.JSONEncoder = data_binding.dump_json,
+        entity_factory: entity_factory_.EntityFactory,
+        event_factory: event_factory_.EventFactory,
+        loads: aiohttp.typedefs.JSONDecoder = data_binding.load_json,
+        rest_client: rest_client_.RESTClient,
+        public_key: typing.Optional[bytes] = None,
```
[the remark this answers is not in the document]
FasterSpeeding: When has that style ever been enforced for implementation inits, didn't they always prefer alphabetical order over ordering based on requiredness for key-word only arguments where order doesn't actually matter at runtime like in shard?

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,558 @@
...
+        application: typing.Union[applications.Application, applications.AuthorizationApplication]
+        async with self._application_fetch_lock:
+            if self._verify:
+                return self._verify
+
+            try:
+                application = (await self._rest_client.fetch_authorization()).application
+
+            except errors.UnauthorizedError:
+                application = await self._rest_client.fetch_application()
```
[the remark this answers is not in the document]
FasterSpeeding: One only works with bearer authorization, one only works with bot authorization; there's zero overlap between them

hikari/impl/entity_factory.py
```diff
@@ -470,7 +476,9 @@ def deserialize_audit_log(self, payload: data_binding.JSONObject) -> audit_log_m
                         new_value = value_converter(new_value) if new_value is not None else None
                         old_value = value_converter(old_value) if old_value is not None else None
 
-                    elif not isinstance(key, audit_log_models.AuditLogChangeKey):
+                    elif _LOGGER.isEnabledFor(logging.DEBUG) and not isinstance(
+                        key, audit_log_models.AuditLogChangeKey
+                    ):
```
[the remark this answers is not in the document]
FasterSpeeding: A what about avoiding making multiple unnecessary instance checks every time an entity is deserialized and b isEnabledFor isn't exclusively used for what you described, e.g. :

```json
                # Assume exception for now.
                ex = self.exception()
                self.logger.warning(
                    "encountered unexpected error: %s",
                    ex,
                    exc_info=ex if self.logger.isEnabledFor(logging.DEBUG) else None,
                )
                raise errors.GatewayError("Unexpected websocket exception from gateway") from ex
```

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,558 @@
...
+        typing.Type[interaction_server.InteractionT],
+        interaction_server.ListenerT[interaction_server.InteractionT, interaction_server.ResponseT],
+    ]
+
+
+_LOGGER: typing.Final[logging.Logger] = logging.getLogger("hikari.interaction_server")
+
+# Internal interaction and interaction response types.
+_PING_INTERACTION_TYPE: typing.Final[int] = 1
+_PONG_RESPONSE_TYPE: typing.Final[int] = 1
```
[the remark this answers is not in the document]
FasterSpeeding: yes

hikari/api/interaction_server.py
```diff
@@ -0,0 +1,214 @@
...
+    #     self,
+    #     interaction_type: typing.Type[commands.CommandInteraction],
+    #     listener: typing.Optional[
+    #         MainListenerT[commands.CommandInteraction, special_endpoints.InteractionResponseBuilder]
+    #     ],
+    #     /,
+    #     *,
+    #     replace: bool = False,
+    # ) -> None:
+    #     raise NotImplementedError
```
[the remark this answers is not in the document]
FasterSpeeding: you can't actually have only 1 override, those were just put in provisionally and won't be enabled until the system's expanded later on if this style holds out down the line

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,558 @@
...
+        event_factory: event_factory_.EventFactory,
+        loads: aiohttp.typedefs.JSONDecoder = data_binding.load_json,
+        rest_client: rest_client_.RESTClient,
+        public_key: typing.Optional[bytes] = None,
+    ) -> None:
+        self._application_fetch_lock = asyncio.Lock()
+        self._dumps = dumps
+        self._entity_factory = entity_factory
+        self._event_factory = event_factory
+        self._future: asyncio.Future[None] = asyncio.Future()
```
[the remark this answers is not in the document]
FasterSpeeding: Implementation detail wise an event is just a collection of Futures instead of just one future and that's how it'd be implemented in most scenarios so I don't really see the difference here myself but ok will change

hikari/impl/interaction_server.py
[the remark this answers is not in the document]
davfsa: Yep, that is what I was missing :)

hikari/internal/ed25519.py
FasterSpeeding: It's internal detail so nobody should be touching this stuff outside of the library itself and treating each one separately just makes stuff like testing that we're using both properly easier and debugging easier

.flake8
```diff
@@ -23,11 +23,12 @@ ignore =
...
 # D106, D104, D103, D101, D100: Missing docstring in public function/class/module (ignore in tests)
 per-file-ignores =
-    hikari/__init__.py:              F401,F403
-    hikari/events/__init__.py:       F401,F403
-    hikari/internal/routes.py:       FS003
-    hikari/internal/time.py:         FS003
-    tests/hikari/*:                  FS003,S106,D106,D104,D103,D101,D100
+    hikari/__init__.py:                 F401,F403
+    hikari/events/__init__.py:          F401,F403
+    hikari/interactions/__init__.py:    F401,F403
```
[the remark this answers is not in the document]
FasterSpeeding: doesn't seem necessary to ignore it in `hikari/api/__init__.py` or  `hikari/api/__internal__.py`

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,558 @@
...
+# Header keys and values
+_X_SIGNATURE_ED25519_HEADER: typing.Final[str] = "X-Signature-Ed25519"
+_X_SIGNATURE_TIMESTAMP_HEADER: typing.Final[str] = "X-Signature-Timestamp"
+_CONTENT_TYPE_KEY: typing.Final[str] = "Content-Type"
+_USER_AGENT_KEY: typing.Final[str] = "User-Agent"
+_JSON_CONTENT_TYPE: typing.Final[str] = "application/json"
+_TEXT_CONTENT_TYPE: typing.Final[str] = "text/plain"
+
+# Constant response payloads
+_PONG_PAYLOAD: typing.Final[str] = data_binding.dump_json({"type": _PONG_RESPONSE_TYPE})
```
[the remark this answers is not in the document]
FasterSpeeding: That's not where interning would benefit anything:
```
Enter string in the table of “interned” strings and return the interned string – which is string itself or a copy. Interning strings is useful to gain a little performance on dictionary lookup – if the keys in a dictionary are interned, and the lookup key is interned, the key comparisons (after hashing) can be done by a pointer compare instead of a string compare. Normally, the names used in Python programs are automatically interned, and the dictionaries used to hold module, class or instance attributes have interned keys.
```
and strings are implicitly interned under cpython anyway

hikari/internal/ed25519.py
davfsa: Should we also change it here then?

https://github.com/hikari-py/hikari/blob/9b194734db0d746c59be5286469a9e2b06ec032b/hikari/internal/time.py#L90-L101

hikari/api/special_endpoints.py
```diff
@@ -517,3 +531,299 @@ def add_stage_channel(
...
+
+
+CommandResponseTypes = typing.Union[
+    "typing.Literal[bases.ResponseType.SOURCED_RESPONSE]",
+    "typing.Literal[bases.ResponseType.DEFERRED_SOURCED_RESPONSE]",
+]
+"""Type hints of the valid response types for a command interaction."""
+
+
+CommandResponseBuilderT = typing.TypeVar("CommandResponseBuilderT", bound="CommandResponseBuilder")
```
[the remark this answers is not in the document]
FasterSpeeding: Won't work for CommandResponseTypes since that's included in `__all__` for the sake of documentation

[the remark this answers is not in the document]
FasterSpeeding: Also flake8 still complains if those bounds aren't strings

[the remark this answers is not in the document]
davfsa: What does it complain about?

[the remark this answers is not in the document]
FasterSpeeding: The referenced variables being undefined

hikari/internal/ed25519.py
FasterSpeeding: Might be worth changing at a later time but that's outside the scope of this PR

hikari/impl/entity_factory.py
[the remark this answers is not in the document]
davfsa: > 
> 
> what about avoiding making multiple unnecessary instance checks every time an entity is deserialized? Also isEnabledFor isn't exclusively used for what you described within the library e.g. :
> 
> ```json
>                 # Assume exception for now.
>                 ex = self.exception()
>                 self.logger.warning(
>                     "encountered unexpected error: %s",
>                     ex,
>                     exc_info=ex if self.logger.isEnabledFor(logging.DEBUG) else None,
>                 )
>                 raise errors.GatewayError("Unexpected websocket exception from gateway") from ex
> ```

We use it there to not have logging generate the traceback string and any other info it needs for the exception.

As per the mater in hand, you are right. Even tho this is not called in every entity, it might get called a bit more than we want it to. But I still think that an isinstance check of a class (in this case an enum) will still be faster than checking if the logger is enabled for that level as there are lot of comparisons that happen inside that function

hikari/impl/interaction_server.py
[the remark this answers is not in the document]
davfsa: Also, something that came to mind. Wouldn't it be better to check what token we were provided and to the action based on that?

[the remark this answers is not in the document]
FasterSpeeding: > Because, at the moment, we might be raising UnauthorizedErrors every time we need the application without any need, which might lead to Discord bans.

It makes two requests max for the whole runtime of a program, that's not gonna be an issue.

> Also, something that came to mind. Wouldn't it be better to check what token we were provided and to the action based on that? 

We just don't have access to that information here.

FasterSpeeding:
Don't like how the semantics of interaction_bases.ResponseType will shift as more interaction types are added

FasterSpeeding:
explicitly set charset?

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,499 @@
...
+    import ssl
+
+    from hikari import applications
+    from hikari.api import entity_factory as entity_factory_
+    from hikari.api import rest as rest_
+    from hikari.api import special_endpoints
+    from hikari.interactions import bases as interaction_bases
+
+
+class RESTBot(traits.InteractionServerAware, interaction_server_.InteractionServer):
```
davfsa, line 51: This class doesnt seem to provide any information above the interaction server. Could we just move the little logic there is here there and make that the "main" class?

hikari/impl/rest_bot.py
```diff
@@ -363,7 +363,17 @@ def run(
...
-            asyncio.get_event_loop().create_task(
+            # get_event_loop will error under oddly specific cases such as if set_event_loop has been called before even
+            # if it was just called with None or if it's called on a thread which isn't the main Thread so it's easier
+            # and more consistent to just explicitly make a new loop.
+            try:
+                loop = asyncio.get_running_loop()
+
+            except RuntimeError:
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
```
davfsa, line 374: Should this be in `hikari/impl/bot.py` too?

davfsa:
Looks great so far, thank you for working on this :+1:

hikari/impl/rest_bot.py
FasterSpeeding: This is just following the lib's standard pattern where logic which brings together multiple component implementations to make a full working client is kept to "apps" and "bots" with the individual implementations of different components taking in other components as arguments.
This also keeps concepts like printing banners and checking versions specifically to "bots" instead of extending the scope of individual component implementations.

This pattern as a whole leaves the implementation classes in a more flexible state than the bot classes.

hikari/impl/rest_bot.py
FasterSpeeding: Yes but likely not in this pr

hikari/impl/rest_bot.py
davfsa, line 51: I feel like `RESTBot` will cause confusion with `RESTApp` :confused: 

Maybe?
```suggestion
class InteractionBot(traits.InteractionServerAware, interaction_server_.InteractionServer):
```

FasterSpeeding: Why would people think that RESTApp is a bot?

FasterSpeeding: also random point, would it be worth it renaming "Bot" to something like "GatewayBot"?

hikari/impl/rest_bot.py
```diff
@@ -456,6 +474,7 @@ async def start(
             For more information on the other parameters such as defaults see
             AIOHTTP's documentation.
         """
+        # self._server.start should raise errors.ComponentStateConflictError if it's already running
```
davfsa, line 477: This will still spawn check_for_updates task, which we don't want to do if there is a ComponentStateConflictError

hikari/impl/bot.py
```diff
@@ -75,7 +75,7 @@
 _LOGGER: typing.Final[logging.Logger] = logging.getLogger("hikari.bot")
 
 
-class BotApp(traits.BotAware):
+class BotApp(traits.GatewayBotAware):
```
davfsa, line 78: This naming makes it really confusing now

hikari/api/shard.py
```diff
@@ -151,10 +151,6 @@ async def join(self) -> None:
         websocket runtime itself.
         """
 
-    @abc.abstractmethod
-    async def start(self) -> None:
-        """Start the shard, wait for it to become ready."""
-
```
davfsa, line 157: Why was this removed?

hikari/impl/rest_bot.py
davfsa: That rename would just cause more naming issues than fixing them

FasterSpeeding: wdym fixing them

hikari/impl/rest_bot.py
FasterSpeeding: There isn't any defined behaviour for how far it will have gone or what will have been done before that error is raised and check for updates isn't mutating so i don't see any harm in it as an artifact

hikari/impl/bot.py
FasterSpeeding: what's confusing about it

hikari/api/shard.py
FasterSpeeding: Because it's implementation detail, the only place where a sub-component will be started is in a place where internal detail is being used in-order to initialise and start it, like within bot app. Similar to how the rest interface doesn't have a start method but some implementation may require that it is explicitly "started"

davfsa: It is not. `start` and `join` can be used instead of `run` if you want to have more control over what it does.

hikari/impl/rest_bot.py
davfsa: There is, as check_for_updates will have already been called if the parameter was passed

hikari/api/shard.py
FasterSpeeding: not a single interface in hikari.api defines a run method(?)

hikari/impl/bot.py
davfsa: The fact that there are 100 different names for the same things. All this name refactoring doesn't bring anything to the pr, it just confuses naming even more

FasterSpeeding: so you're saying BotAware should be renamed to BotApp?🤔

hikari/impl/rest_bot.py
FasterSpeeding: Idk what you're getting at with that by yes it will be called every time you call the function

hikari/api/shard.py
FasterSpeeding: > It is not. start and join can be used instead of run if you want to have more control over what it does.

and how is this relevant to the public abstract interface for a Shard instance?

davfsa: You kinda need to call `start` before `join` to start the shard :thinking:

FasterSpeeding: and the manager won't be calling it based on the abc since it'd just built the implementation class

hikari/impl/bot.py
FasterSpeeding: Still don't see how there are 100 different names for the same thing here though. With a "bot" just being an autonomous program that's separate from Discord and can interact with their users which can be run by itself without the need to introduce any other components as it'll build them itself in this lib with a gateway bot enabling this kind of functionality over by combining multiple sub-components (shards, rest client and multiple dispatch) and receiving gateway events based on user action while a rest bot enables this functionality with only stateless restful sub-components (rest client and rest server) (with the rest server also handling single dispatch which is a design limitation of the rest only flow) and receiving RESTful webhook events based on user action.

hikari/api/shard.py
FasterSpeeding: > You kinda need to call start before join to start the shard 🤔

also what if it's a shard implementation which just wraps around a receiver and never needs to be started itself

hikari/api/interaction_server.py
```diff
@@ -107,34 +107,10 @@ class InteractionServer(abc.ABC):
...
-            Whether this interaction server is active
-        """
-
-    @abc.abstractmethod
-    async def close(self) -> None:
-        """Gracefully close the server and any open connections."""
-
-    @abc.abstractmethod
-    async def join(self) -> None:
-        """Wait for the process to halt before continuing."""
```
davfsa: Why tho? This provides a nice interface people can extend upon. If they dont want to inherit it, they can just have it return when called

hikari/events/interaction_events.py
```diff
@@ -0,0 +1,125 @@
...
+from hikari.internal import attr_extensions
+
+if typing.TYPE_CHECKING:
+    from hikari import traits
+    from hikari.api import shard as gateway_shard
+    from hikari.interactions import bases as interaction_bases
+    from hikari.interactions import commands
+
+
+class CommandEvent(shard_events.ShardEvent, abc.ABC):
```
FasterSpeeding, line 49: These are sent for other bot's commands as well

FasterSpeeding:
This probably shouldn't be merged before #582 since this introduces cache getter methods in the style introduced in that PR

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,518 @@
...
+
+        Users should consider this an advice to the application on whether it is
+        safe to show colours if possible or not. Since some terminals can be
+        awkward or not support features in a standard way, the option to
+        explicitly disable this is provided. See `force_color` for an
+        alternative.
+    application : typing.Optional[hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialApplication]]
+        Object or ID of the application this bot instance should be associated
+        with. If left as `builtins.None` then the client will try to work this
+        value out based on `token`.
```
davfsa, line 83: Was just going through the pr looking for something. Found this un-used docstring

hikari/impl/entity_factory.py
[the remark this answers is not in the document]
FasterSpeeding: > But I still think that an isinstance check of a class (in this case an enum) will still be faster than checking if the logger is enabled for that level as there are lot of comparisons that happen inside that function

That stuff is all cached though, most calls will just be a dict lookup

hikari/api/special_endpoints.py
```diff
@@ -23,7 +23,13 @@
 """Special additional endpoints used by the REST API."""
 from __future__ import annotations
 
-__all__: typing.List[str] = ["TypingIndicator", "GuildBuilder"]
+__all__: typing.List[str] = [
+    "CommandBuilder",
+    "TypingIndicator",
+    "GuildBuilder",
+    "CommandResponseBuilder",
+    "InteractionResponseBuilder",
```
FasterSpeeding, line 31: Is special endpoints the right place for these response builders or might it start to get messy in the future?

hikari/impl/bot.py
```diff
@@ -1049,3 +1052,20 @@ def _validate_activity(activity: undefined.UndefinedNoneOr[presences.Activity])
...
+    """
+
+    # This is only overridden at runtime in-order to preserve this class'
+    # inherited init pattern from `GatewayBot` without repeating it
+    if not typing.TYPE_CHECKING:
+
+        def __init__(self, *args: typing.Any, **kwargs: typing.Any) -> None:
+            super().__init__(*args, **kwargs)
+            # This is called after the super call as __init__ may alter the warnings config
+            deprecation.warn_deprecated(BotApp, stack_level=3)
```
davfsa, line 1071:
```suggestion
@deprecation.deprecated(version="2.0.0.dev102", alternative="GatewayBot")
class BotApp(GatewayBot):
    """Alias for `GatewayBot`."""
```

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,514 @@
...
+        builtins.bool
+            Whether this interaction server is active
+        """
+        return self._runner is not None
+
+    def _get_event(self) -> asyncio.Event:
+        if self._event is None:
+            self._event = asyncio.Event()
+
+        return self._event
```
davfsa, line 199: Could this lead to weird state errors?

hikari/impl/interaction_server.py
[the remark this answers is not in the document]
davfsa: Unresolving this as it can be changed once #627 is merged

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,514 @@
...
+            return aiohttp.web.Response(
+                status=_PAYLOAD_TOO_LARGE_STATUS,
+                body=b"Payload too large",
+                content_type=_TEXT_CONTENT_TYPE,
+                charset=_UTF_8_CHARSET,
+            )
+
+        if not body:
+            user_agent = request.headers.get(_USER_AGENT_KEY, "NONE")
+            _LOGGER.info("Received a body-less request (UA %r)", user_agent)
```
davfsa, line 271: Could these be debug instead of info?

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,514 @@
...
+                body=b"Missing or invalid required request signature header(s)",
+                content_type=_TEXT_CONTENT_TYPE,
+                charset=_UTF_8_CHARSET,
+            )
+
+        try:
+            body = await request.read()
+
+        except aiohttp.web.HTTPRequestEntityTooLarge:
+            return aiohttp.web.Response(
```
davfsa, line 262: Logging could be added here too (and the other error above)

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,514 @@
...
+            payload = self._loads(body.decode("utf-8"))
+            interaction_type = int(payload["type"])
+
+        except (data_binding.JSONDecodeError, ValueError, TypeError) as exc:
+            _LOGGER.error("Received a request with an invalid JSON body", exc_info=exc)
+            return _Response(_BAD_REQUEST_STATUS, b"Invalid JSON body")
+
+        except KeyError as exc:
+            _LOGGER.error("Missing 'type' field in received JSON payload", exc_info=exc)
+            return _Response(_BAD_REQUEST_STATUS, b"Missing required 'type' field in payload")
```
davfsa, line 343: Could this spam the console if Discord try to trigger this?

hikari/impl/interaction_server.py
```diff
@@ -0,0 +1,514 @@
...
+        ],
+        /,
+        *,
+        replace: bool = False,
+    ) -> None:
+        if listener:
+            if not replace and interaction_type in self._listeners:
+                raise TypeError(f"Listener already set for {interaction_type.__name__}")
+
+            self._listeners[interaction_type] = listener  # type: ignore[assignment]
```
davfsa, line 511: Why the type ignore?

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,510 @@
...
+        self._rest = rest_impl.RESTClientImpl(
+            entity_factory=self._entity_factory,
+            executor=self._executor,
+            http_settings=self._http_settings,
+            max_rate_limit=max_rate_limit,
+            proxy_settings=self._proxy_settings,
+            rest_url=rest_url,
+            token=token,
+            token_type=token_type,
+        )
```
davfsa, line 212: Cache #660

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,510 @@
...
+
+        !!! note
+            `force_color` will always take precedence over `allow_color`.
+        """
+        ux.print_banner(banner, allow_color, force_color)
+
+    async def close(self) -> None:
+        # self._server.start should raise errors.ComponentStateConflictError if it's not running
+        await self._server.close()
+        await self._rest.close()
```
davfsa, line 281: Do not really understand this comment here

hikari/impl/rest_bot.py
```diff
@@ -0,0 +1,510 @@
...
+
+    async def join(self) -> None:
+        await self._server.join()
+
+    async def on_interaction(self, body: bytes, signature: bytes, timestamp: bytes) -> interaction_server_.Response:
+        return await self._server.on_interaction(body, signature, timestamp)
+
+    def run(
+        self,
+        asyncio_debug: typing.Optional[bool] = None,
```
davfsa, line 291:
```suggestion
        asyncio_debug: bool = False,
```

hikari/impl/rest_bot.py
davfsa: What I mean is, we will always create the task for `check_for_updates`, even if there is a StateConflicError, which may lead to the task falling out of scope and erroring out

hikari/interactions/commands.py
```diff
@@ -0,0 +1,838 @@
...
+            Mentioning everyone, roles, or users in message edits currently
+            will not send a push notification showing a new mention to people
+            on Discord. It will still highlight in their chat as if they
+            were mentioned, however.
+
+        !!! note
+            There is currently no documented way to clear attachments or edit
+            attachments from a previously sent message on Discord's API. To
+            do this, delete the message and re-send it. This also applies
+            to embed attachments.
```
davfsa, line 702: There is now :)

hikari/interactions/commands.py
```diff
@@ -0,0 +1,838 @@
...
+            If an internal error occurs on Discord while handling the request.
+        """
+        return await self.app.rest.fetch_channel(self.channel_id)
+
+    def get_channel(self) -> typing.Optional[channels.GuildChannel]:
+        """Get the guild channel this was triggered in from the cache.
+
+        !!! note
+            This will always return `builtins.None` for interactions triggered
+            in a DM channel.
```
davfsa, line 827: #660 :)

hikari/webhooks.py
```diff
@@ -59,96 +60,48 @@ class WebhookType(int, enums.Enum):
...
-            This exists purely for consistency. Webhooks do not receive events
-            from the gateway, and without some bot backend to support it, will
-            not be able to detect mentions of their webhook.
-
-        Example
+        Returns
         -------
+        hikari.snowflakes.Snowflake
+            The ID used to execute this entity as a webhook.
+        """
```
davfsa, line 89: In what case could this be different to `.id`?

hikari/impl/interaction_server.py
FasterSpeeding: For the same reason the type ignores are in the event dispatcher, because that just ain't how the variance of callables works

hikari/webhooks.py
FasterSpeeding: For interactions (?)

hikari/impl/interaction_server.py
davfsa: Should this be something we fix? I have seen a lof of these

hikari/__init__.py
```diff
@@ -22,7 +22,7 @@
 # SOFTWARE.
 """A sane Python framework for writing modern Discord bots.
 
-To get started, you will want to initialize an instance of `BotApp`
+To get started, you will want to initialize an instance of `GatewayBot`
 for writing a bot, or `RESTApp` if you only need to use the REST API.
```
davfsa, line 26: Shouldnt more info on RESTBot be added here along with an example in the examples folder + README?

hikari/impl/rest_bot.py
FasterSpeeding: this is copy pasted from the gateway bot impl

davfsa: Should be changed on both then. Makes no sense to allow for `None`. I can open another pr if you want

hikari/impl/bot.py
FasterSpeeding: That would both be a breaking change and also break the typing for anybody who's accessing stuff on the BotApp class like unbound methods

hikari/impl/interaction_server.py
FasterSpeeding: Nope

hikari/impl/interaction_server.py
FasterSpeeding: Not like it should be silent about receiving a ton of invalid payloads which as far as we're concerned are from Discord themselves, it's a lot spammier if it it manages to get down to the except Exception wrapping deserialization

hikari/impl/interaction_server.py
FasterSpeeding: There's not really any way to "fix" it per-say, best you can do is just replace `typing.Callable[[Event], ...]` with `typing.Callable[[Any], ...]` in the typing for the listeners Dict to get rid of the type ignores

hikari/impl/rest_bot.py
FasterSpeeding: it's meant to be .close not .start, that's a typo

hikari/interactions/commands.py
FasterSpeeding: RESTBot is probably going to stay cache-less and that new store isn't overly helpful here anyway

hikari/__init__.py
FasterSpeeding: can be added in another PR

hikari/api/interaction_server.py
FasterSpeeding: Because how it's run/started/stopped is implementation detail, can't write an interface which is flexible enough to allow for stuff like requiring that they pass an active ASGI server to the "start" method can you and restricting how a sub-component is managed isn't overly helpful when sub-components should only ever be started/run by their parent app/bot

hikari/api/interaction_server.py
```diff
@@ -0,0 +1,202 @@
...
+            Instructions on how the REST server calling this should respond to
+            the interaction request.
+        """
+
+    # # This overload cannot be included until component interactions are implemented
+    # @typing.overload
+    # def get_listener(
+    #     self, interaction_type: typing.Type[commands.CommandInteraction], /
+    # ) -> typing.Optional[ListenerT[commands.CommandInteraction, _MessageResponseBuilderT]]:
+    #     ...
```
davfsa, line 138: Because of stuff like this, should we merge this into a feature branch, push all the documentation updates as well as the components there and then merge to master?

davfsa (commented):
LGTM

hikari/api/interaction_server.py
FasterSpeeding: Seems kinda unnecessary to me tbh (the component stuff can just be PR'ed into master as soon as this is merged anyway) plus that'd merge even more changes into what is already going to be a giant single commit which we'd probably want to avoid doing any more of lol.

davfsa: I don't mind having huge branch merges as long as we break them down into smaller commits, which will make sure we have been able to review all the changes and make sure they will fit nicely into master, which I'm not so sure from this.

FasterSpeeding: yeah but all PR merges are squashed so a feature branch would just lead to all the commits to it being squashed into one

davfsa (approved):
LGTM, just merge conflicts

FasterSpeeding:
@davfsa rebased and added in small thing which was breaking (BotAware as a deprecated alias for GatewayBotAware) 👍

davfsa (approved):

PR #580, opened 2021-05-04
Handle InvalidStateError on waiter dispatch

FasterSpeeding:
### Summary
Handle InvalidStateError on waiter dispatch

### Checklist

### DIFF
--- a/hikari/impl/event_manager_base.py
+++ b/hikari/impl/event_manager_base.py
@@ -223,9 +223,13 @@ class EventManagerBase(event_manager.EventManager):
                 for callback in self._listeners[cls]:
                     tasks.append(self._invoke_callback(callback, event))
 
-            if cls in self._waiters:
-                waiter_set = self._waiters[cls]
-                for predicate, future in tuple(waiter_set):
+            if cls not in self._waiters:
+                continue
+
+            waiter_set = self._waiters[cls]
+            for waiter in tuple(waiter_set):
+                predicate, future = waiter
+                if not future.cancelled():
                     try:
                         result = predicate(event)
                         if not result:
@@ -235,7 +239,7 @@ class EventManagerBase(event_manager.EventManager):
                     else:
                         future.set_result(event)
 
-                    waiter_set.remove((predicate, future))
+                waiter_set.remove(waiter)
 
         return asyncio.gather(*tasks) if tasks else aio.completed_future()
 
--- a/hikari/impl/event_manager_base.py
+++ b/hikari/impl/event_manager_base.py
@@ -229,7 +229,7 @@ class EventManagerBase(event_manager.EventManager):
             waiter_set = self._waiters[cls]
             for waiter in tuple(waiter_set):
                 predicate, future = waiter
-                if not future.cancelled():
+                if not future.done():
                     try:
                         result = predicate(event)
                         if not result:


hikari/impl/event_manager_base.py
```diff
@@ -223,9 +223,13 @@ def dispatch(self, event: event_manager.EventT_inv) -> asyncio.Future[typing.Any
...
-            if cls in self._waiters:
-                waiter_set = self._waiters[cls]
-                for predicate, future in tuple(waiter_set):
+            waiter_set = self._waiters.pop(cls, None)
+            if not waiter_set:
+                continue
+
+            for waiter in waiter_set:
+                predicate, future = waiter
+                try:
```
davfsa, line 232: Always popping the waiters will not have the effect we want. The way this currently works is that the waiters wont be removed if the result of the predicate is not trully.

This code will remove the waiter always, regardless if the future gets completed or if the event is the one that we want.

Furthermore, the bug that you are trying to fix lies somewhere else as, if we cleanup everything correctly, there should not be an invalid state

davfsa (changes requested):

hikari/impl/event_manager_base.py
FasterSpeeding: ah ok

hikari/impl/event_manager_base.py
```diff
@@ -235,7 +239,13 @@ def dispatch(self, event: event_manager.EventT_inv) -> asyncio.Future[typing.Any
                     else:
                         future.set_result(event)
 
-                    waiter_set.remove((predicate, future))
+                except asyncio.InvalidStateError:
+                    pass
```
davfsa, line 243: As I said in the other comment, we shouldn't be needing to handle this error, as we should never encounter it if we remove the future after it is completed.

hikari/impl/event_manager_base.py
```diff
@@ -223,9 +223,13 @@ def dispatch(self, event: event_manager.EventT_inv) -> asyncio.Future[typing.Any
...
 
-            if cls in self._waiters:
-                waiter_set = self._waiters[cls]
-                for predicate, future in tuple(waiter_set):
+            if cls not in self._waiters:
+                continue
+
+            waiter_set = self._waiters[cls]
+            for waiter in tuple(waiter_set):
+                predicate, future = waiter
```
davfsa, line 231: Why unpack it here and not in the `for` loop?

hikari/impl/event_manager_base.py
FasterSpeeding: Welp if you fine with event dispatch being stopped by cancelled futures then you do you ig 🤷‍♀️

davfsa: The bug is still there, I just dont think that this is the way to go about fixing it. This will just suppress it, not actually fix it. Unless there is another case when a future will be cancelled that is out of our control (which is then causing this issue), I don't think this is the way to fix it

FasterSpeeding: Well without that try except other listeners may miss an event

davfsa: Turns out I'm just being dumb and this is the best way to do it. Might also be best to check `if future.done()` and not even run the predicate, just remove the waiter from the mapping instead of catching the error.

davfsa (dismissed):

davfsa (approved):

PR #582, opened 2021-05-06
Move away from hiding cache calls behind properties

FasterSpeeding:
### Summary
Move away from hiding cache calls behind properties
* This mainly applies to events (plus Guild, Member and Message)
* Add in "app" properties to applicable event objects to replace the app attrs field
* rename some cache getting properties to `get_{original_name}`

Since previously we've already had the standard of having cache calls behing get_x methods with the idea that a property/instance variable should remain relatively constant while the results of get methods has no guarantee of being constant, this is being used in this PR. The idea here is to make it more explicit when a cache call is being made vs when a field being present is dependent on API cases within the scope of the object that's currently being operated on to help keep semantics simple and explicit. This also helps avoid issues where a seemingly simple property may be making multiple non-lazy cache calls to build a dict where multiple accesses to the property would lead to extra unnecessary work being done.

This also has a fix for an internal bug which likely stemmed from this approach of hiding cache calls behind properties

### Checklist

### DIFF
--- a/hikari/events/guild_events.py
+++ b/hikari/events/guild_events.py
@@ -83,22 +83,6 @@ class GuildEvent(shard_events.ShardEvent, abc.ABC):
             The ID of the guild that relates to this event.
         """
 
-    @property
-    def guild(self) -> typing.Optional[guilds.GatewayGuild]:
-        """Get the cached guild that this event relates to, if known.
-
-        If not known, this will return `builtins.None` instead.
-
-        Returns
-        -------
-        typing.Optional[hikari.guilds.GatewayGuild]
-            The guild this event relates to, or `builtins.None` if not known.
-        """
-        if not isinstance(self.app, traits.CacheAware):
-            return None
-
-        return self.app.cache.get_available_guild(self.guild_id) or self.app.cache.get_unavailable_guild(self.guild_id)
-
     async def fetch_guild(self) -> guilds.RESTGuild:
         """Perform an API call to get the guild that this event relates to.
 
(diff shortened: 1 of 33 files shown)


hikari/events/guild_events.py
```diff
@@ -319,17 +321,31 @@ class GuildUpdateEvent(GuildEvent):
...
+    def app(self) -> traits.RESTAware:
+        # <<inherited docstring from Event>>.
+        return self.guild.app
+
     @property
     def guild_id(self) -> snowflakes.Snowflake:
         # <<inherited docstring from GuildEvent>>.
         return self.guild.id
 
+    def get_guild(self) -> typing.Optional[guilds.GatewayGuild]:
```
FasterSpeeding, line 334: Remove this

hikari/events/guild_events.py
```diff
@@ -228,11 +224,20 @@ class GuildAvailableEvent(GuildVisibilityEvent):
...
+        # <<Inherited docstring from GuildEvent>>
```
FasterSpeeding, line 238: Remove overriding additions like this

hikari/guilds.py
```diff
@@ -945,6 +945,218 @@ def make_icon_url(self, *, ext: typing.Optional[str] = None, size: int = 4096) -
...
+    def get_emojis(self) -> typing.Mapping[snowflakes.Snowflake, emojis_.KnownCustomEmoji]:
+        """Get the emojis cached for the guild.
+
+        Returns
+        -------
+        typing.Mapping[hikari.snowflakes.Snowflake, Member]
+            A mapping of emoji IDs to objects of the emojis cached for the guild.
+        """
+        if not isinstance(self.app, traits.CacheAware):
+            return {}
```
FasterSpeeding, line 971: Dont we have an empty view class somewhere?

FasterSpeeding:
This PR skips out some message and message event methods which will likely be changed in a bigger refactor of the message objects structure

FasterSpeeding:
Pretty sure you missed Bot.me

davfsa (commented):
Looks good, I'm thinking if we should merge this into another branch before master and then push the full "breaking changes" with messages included

hikari/messages.py
```diff
@@ -619,7 +620,7 @@ class PartialMessage(snowflakes.Unique):
     `type` is `MessageType.REPLY` and `builtins.None`, the message was deleted.
     """
 
-    @property
+    @property  # TODO: this breaks from the new convention for cache methods
```
FasterSpeeding, line 623: This logic is kinda off since it doesn't distinguish between the guild_id being unknown and the guild_id being known to be None

davfsa (dismissed):
This is going to be a bit of a breaking change for a lot of people :P

davfsa (approved):

Issue #37, opened 2020-08-20
Guilds methods


davfsa:
Methods and properties should be added to guild models where appropriate.

These may take the shape of the following, but this is just a recommendation and should be at the developer's discretion.

Also, these should be tested and fully documented.

- `GuildWidget#fetch_channel`
- `Member`
    - `#fetch_user`
    - `#fetch_roles`
    - ~`#ban`~
    - `#kick`
    - `#add_roles`
    - `#remove_roles`
- `MemberPresence`
    - `#fetch_user`
    - `#fetch_member`
- `GuildIntegration`
    - `#fetch_user`
    - `#fetch_member` (?)
    - `#sync` (?)
- `GuildMemberBan`
    - Rename to `GuildBan`
    - `#unban`
- `GuildPreview`
    - `#fetch_guild`
- `Guild`
    - `#edit`
    - `#ban`
    - `#kick`
    - `#unban`
    - `#fetch_owner`
    - `#fetch_afk_channel`
    - `#fetch_roles` async iter
    - `#fetch_emojis`
    - `#fetch_widget_channel`
    - `#fetch_system_channel`
    - `#fetch_rules_channel`
    - `#fetch_public_updates_channel`
    - `#update_voice_state`
    - `$text_channels`
    - `$voice_channels`
    - `$categories`
    - `$news_channels`
    - `$store_channels`
    - `#add_channel`
    - `#add_category` (redundant?)
    - `#remove_channel`
    - `#remove_category` (redundant?)

ahnaf-zamil:
Lemme try and see what I can do

ashwinvin:
Seems like pretty easy thing to do. Can I attempt at doing this?

PR #591, opened 2021-05-09
Add guild methods

ashwinvin:
### Summary
This PR adds the methods listed in #37 by using the bot's REST client.

### Checklist

- [x] I have made unittests according to the code I have added.

### Related issues
#37


hikari/guilds.py
```diff
@@ -79,6 +79,7 @@
     from hikari import permissions as permissions_
     from hikari import presences as presences_
     from hikari import voices as voices_
+    from hikari.internal import time as time_
```
davfsa, line 82: This can be `from hikari.internal import time`, no need for the alias. The reason why the other imports have it is because they collide with the naming of the arguments, which they don't in this case.

davfsa (commented):
Looks good, just a little remark. Request my review when its ready 👍

hikari/guilds.py
```diff
@@ -79,6 +80,7 @@
     from hikari import permissions as permissions_
     from hikari import presences as presences_
     from hikari import voices as voices_
+    from hikari.internal import time as time
```
FasterSpeeding, line 83: the as part of this is redundant

ashwinvin: oops didn't notice it 😅

tests/hikari/test_guilds.py
```diff
@@ -762,3 +763,172 @@ def test_get_my_member(self, model):
...
+    async def test_unban(self, model):
+        model.app.rest.unban_user = mock.AsyncMock()
+        await model.unban(4321, reason="Comeback!!")
+
+        model.app.rest.unban_user.assert_awaited_once_with(123, 4321, reason="Comeback!!")
+
+    @pytest.mark.asyncio
+    async def test_edit(self, model):
+        model.app.rest.edit_guild = mock.AsyncMock()
+        await model.edit(
```
davfsa, line 791: This should be checking the return too

tests/hikari/test_guilds.py
```diff
@@ -762,3 +763,172 @@ def test_get_my_member(self, model):
...
+            public_updates_channel=undefined.UNDEFINED,
+            preferred_locale="us-en",
+            reason="beep boop",
+        )
+
+    @pytest.mark.asyncio
+    async def test_fetch_emojis(self, model):
+        model.app.rest.fetch_guild_emojis = mock.AsyncMock()
+
+        await model.fetch_emojis()
```
davfsa, line 825: Checking return here too

tests/hikari/test_guilds.py
```diff
@@ -762,3 +763,172 @@ def test_get_my_member(self, model):
...
+
+        model.app.rest.fetch_guild_emojis.assert_awaited_once_with(model.id)
+
+    @pytest.mark.asyncio
+    async def test_fetch_emoji(self, model):
+        model.app.rest.fetch_emoji = mock.AsyncMock()
+
+        await model.fetch_emoji(349)
```
davfsa, line 833: As well as here

tests/hikari/test_guilds.py
```diff
@@ -762,3 +763,172 @@ def test_get_my_member(self, model):
...
+
+        model.app.rest.fetch_emoji.assert_awaited_once_with(model.id, 349)
+
+    @pytest.mark.asyncio
+    async def test_create_category(self, model):
+        model.app.rest.create_guild_category = mock.AsyncMock()
+
+        await model.create_category("very cool category", position=2)
```
davfsa, line 841: Ima point this one out too and just stop cause it will get boring, xD

davfsa (changes requested):

ashwinvin:
Rebased !

tests/hikari/test_guilds.py
```diff
@@ -753,7 +753,85 @@ def test_make_banner_url_when_hash(self, model):
...
+
+        afk_channel = await model.fetch_afk_channel()
+
+        assert afk_channel is model.app.rest.fetch_channel.return_value
+        model.app.rest.fetch_channel.assert_awaited_once_with(1234)
+
+        model.afk_channel_id = None
 
+        afk_none_case = await model.fetch_afk_channel()
+        assert None is afk_none_case
```
davfsa: This should be a different test case all together

tests/hikari/test_guilds.py
```diff
@@ -558,6 +754,94 @@ def test_make_banner_url_when_no_hash(self, model):
...
+        model.app.rest.fetch_member.assert_awaited_once_with(123, 1111)
+        assert owner is model.app.rest.fetch_member.return_value
+
+    @pytest.mark.asyncio
+    async def test_fetch_widget_channel(self, model):
+        model.app.rest.fetch_channel = mock.AsyncMock()
+
+        widget_channel = await model.fetch_widget_channel()
+
+        assert widget_channel is model.app.rest.fetch_channel.return_value
```
davfsa, line 781: You can combine these lines. I'm going to guess this was done as you saw some test cases storing the return value in `returned` before checking. This is done due to black formatting being hard to read in asserts which are really long

```suggestion
        assert await model.fetch_widget_channel() is model.app.rest.fetch_channel.return_value
```

ashwinvin: Yea, I saw the other tests and with it. I will change these

tests/hikari/test_guilds.py
ashwinvin: Are you suggesting to combine these type of tests into one function?

davfsa: No, have 2 tests. 1 for `fetch_afk_channel` when None and one when its set:

```py
async def fetch_afk_channel():
    ...

async def fetch_afk_channel_when_None():
    ...
```

You can have a look at the existing tests for reference

ashwinvin: Ah, will change these too

hikari/guilds.py
```diff
@@ -851,7 +1132,7 @@ class WelcomeScreen:
 
 @attr_extensions.with_copy
 @attr.define(hash=False, kw_only=True, weakref_slot=False)
-class GuildMemberBan:
+class Ban:
```
davfsa, line 1135: Maybe best to rename to `GuildBan` instead.

ashwinvin: Ah, I will change that.

hikari/guilds.py
```diff
@@ -263,12 +263,14 @@ class GuildWidget:
...
-    async def fetch_channel(self) -> channels_.PartialChannel:
+    async def fetch_channel(self) -> typing.Optional[channels_.PartialChannel]:
         """Fetch the widget channel.
 
+        This will be `builtins.None` if not set.
+
         Returns
         -------
-        hikari.channels.PartialChannel
+        typing.Optional[hikari.channels.PartialChannel]
```
ashwinvin, line 273: Should I type this as `channels.GuildTextChannel` since guild widgets can only be set to a text channel?

davfsa: They can actually be set to any channel (including voice). Best to keep it as PartialChannel

ashwinvin: Ah, my bad :)

ashwinvin:
I think I have implemented every thing specified in the issue.

tests/hikari/test_guilds.py
```diff
@@ -558,6 +864,87 @@ def test_make_banner_url_when_no_hash(self, model):
...
+        model.app.rest.fetch_channel = mock.AsyncMock()
+
+        assert await model.fetch_public_updates_channel() is model.app.rest.fetch_channel.return_value
+        model.app.rest.fetch_channel.assert_awaited_once_with(99699)
+
+    @pytest.mark.asyncio
+    async def test_fetch_public_updates_channel_when_None(self, model):
+        model.public_updates_channel_id = None
+
+        assert None is await model.fetch_public_updates_channel()
```
davfsa, line 933:
```suggestion
        assert await model.fetch_public_updates_channel() is None
        model.app.rest.fetch_channel.assert_not_called()
```

tests/hikari/test_guilds.py
```diff
@@ -558,6 +864,87 @@ def test_make_banner_url_when_no_hash(self, model):
...
+        model.app.rest.fetch_channel = mock.AsyncMock()
+
+        assert await model.fetch_afk_channel() is model.app.rest.fetch_channel.return_value
+        model.app.rest.fetch_channel.assert_awaited_once_with(1234)
+
+    @pytest.mark.asyncio
+    async def test_fetch_afk_channel_when_None(self, model):
+        model.afk_channel_id = None
+
+        assert None is await model.fetch_afk_channel()
```
davfsa, line 946:
```suggestion
        assert await model.fetch_afk_channel() is None
        model.app.rest.fetch_channel.assert_not_called()
```

hikari/guilds.py
```diff
@@ -1438,6 +2922,121 @@ def channels(self) -> typing.Mapping[snowflakes.Snowflake, channels_.GuildChanne
 
         return self.app.cache.get_guild_channels_view_for_guild(self.id)
 
+    @property
+    def text_channels(self) -> typing.Mapping[snowflakes.Snowflake, channels_.GuildTextChannel]:
```
davfsa, line 2926: @FasterSpeeding wanted to move away from hiding cache calls behind properties. Might be worth leaving this part out till we decide what to do

davfsa (changes requested):

hikari/guilds.py
ashwinvin: Ah ok!

hikari/guilds.py
```diff
@@ -262,6 +263,64 @@ class GuildWidget:
...
+            `hikari.channels.GroupDMChannel`,
+            `hikari.channels.GuildTextChannel`,
+            `hikari.channels.GuildVoiceChannel`,
+            `hikari.channels.GuildStoreChannel`,
+            `hikari.channels.GuildNewsChannel`.
+
+            Likewise, the `hikari.channels.GuildChannel` can be used to
+            determine if a channel is guild-bound, and
+            `hikari.channels.TextChannel` can be used to determine
+            if the channel provides textual functionality to the application.
```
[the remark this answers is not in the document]
ashwinvin: I asked about this a while ago and some said to not change the doc . Anyways will change these ..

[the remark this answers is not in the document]
davfsa: @ashwinvin Yea, that was my bad, xD. You suggested to make it only text channels, but then I realized it can be any `GuildChannel`, so i forgot to tell you to change it 😓

tests/hikari/test_guilds.py
```diff
@@ -228,6 +280,58 @@ async def test_ban(self, model):
...
+    async def test_edit(self, model):
+        model.app.rest.edit_member = mock.AsyncMock()
+        edit = await model.edit(
+            nick="Kitten", roles=[123, 432, 345], mute=False, deaf=True, voice_channel=4321245, reason="I'm God"
+        )
+
+        model.app.rest.edit_member.assert_awaited_once_with(
+            456,
+            123,
+            nick="Kitten",
```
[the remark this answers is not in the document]
ashwinvin: just for laughs xd

hikari/guilds.py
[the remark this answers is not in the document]
ashwinvin: No problem :)

tests/hikari/test_guilds.py
```diff
@@ -558,6 +864,87 @@ def test_make_banner_url_when_no_hash(self, model):
...
+        model.app.rest.fetch_channel = mock.AsyncMock()
+
+        assert await model.fetch_system_channel() is model.app.rest.fetch_channel.return_value
+        model.app.rest.fetch_channel.assert_awaited_once_with(123888)
+
+    @pytest.mark.asyncio
+    async def test_fetch_system_channel_when_None(self, model):
+        model.system_channel_id = None
+
+        assert None is await model.fetch_system_channel()
```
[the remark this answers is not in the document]
davfsa: These were marked as resolved but not actually fixed

tests/hikari/test_guilds.py
davfsa: Same here

tests/hikari/test_guilds.py
davfsa: And here

tests/hikari/test_guilds.py
```diff
@@ -558,6 +864,87 @@ def test_make_banner_url_when_no_hash(self, model):
...
+        model.app.rest.fetch_channel = mock.AsyncMock()
+
+        assert await model.fetch_widget_channel() is model.app.rest.fetch_channel.return_value
+        model.app.rest.fetch_channel.assert_awaited_once_with(192729)
+
+    @pytest.mark.asyncio
+    async def test_fetch_widget_channel_when_None(self, model):
+        model.widget_channel_id = None
+
+        assert None is await model.fetch_widget_channel()
```
davfsa, line 894:
```suggestion
        assert await model.fetch_widget_channel() is None
        model.app.rest.fetch_channel.assert_not_called()
```

tests/hikari/test_guilds.py
```diff
@@ -558,6 +864,87 @@ def test_make_banner_url_when_no_hash(self, model):
...
+        model.app.rest.fetch_channel = mock.AsyncMock()
+
+        assert await model.fetch_rules_channel() is model.app.rest.fetch_channel.return_value
+        model.app.rest.fetch_channel.assert_awaited_once_with(123445)
+
+    @pytest.mark.asyncio
+    async def test_fetch_rules_channel_when_None(self, model):
+        model.rules_channel_id = None
+
+        assert None is await model.fetch_rules_channel()
```
davfsa, line 907:
```suggestion
        assert await model.fetch_rules_channel() is None
        model.app.rest.fetch_channel.assert_not_called()
```

hikari/guilds.py
```diff
@@ -1338,6 +2544,271 @@ def get_emoji(
...
+        return await self.app.rest.fetch_channel(self.system_channel_id)
+
+    async def fetch_rules_channel(self) -> typing.Optional[channels_.PartialChannel]:
+        """Fetch the channel where guilds display rules and guidelines.
+
+        If the `GuildFeature.COMMUNITY` feature is not defined, then this is `builtins.None`.
+
+        Returns
+        -------
+        typing.Optional[hikari.channels.PartialChannel]
```
davfsa, line 2738:
```suggestion
    async def fetch_rules_channel(self) -> typing.Optional[channels_.GuildTextChannel]:
        """Fetch the channel where guilds display rules and guidelines.

        If the `GuildFeature.COMMUNITY` feature is not defined, then this is `builtins.None`.

        Returns
        -------
        typing.Optional[hikari.channels.GuildTextChannel]
```

hikari/guilds.py
```diff
@@ -1338,6 +2544,271 @@ def get_emoji(
...
+
+    async def fetch_system_channel(self) -> typing.Optional[channels_.PartialChannel]:
+        """Fetch the system channel or `builtins.None` if it is not enabled.
+
+        Welcome messages and Nitro boost messages may be sent to this channel.
+
+        Returns
+        -------
+        typing.Optional[hikari.channels.PartialChannel]
+            The channel that the widget is linked to.
```
davfsa, line 2699:
```suggestion
    async def fetch_system_channel(self) -> typing.Optional[channels_.GuildTextChannel]:
        """Fetch the system channel.

        Returns
        -------
        typing.Optional[hikari.channels.GuildTextChannel]
            The channel system channel for this guild or `builtins.None` if not
            enabled.
```

hikari/guilds.py
```diff
@@ -1338,6 +2544,271 @@ def get_emoji(
...
+    async def fetch_afk_channel(self) -> typing.Optional[channels_.PartialChannel]:
+        """Fetch the channel that AFK voice users get sent to.
+
+        If this information is unavailable or this is not enabled for the guild then
+        this will be `builtins.None`.
+
+        Returns
+        -------
+        typing.Optional[hikari.channels.PartialChannel]
+            The channel that the widget is linked to.
```
davfsa, line 2659:
```suggestion
    async def fetch_afk_channel(self) -> typing.Optional[channels_.GuildVoiceChannel]:
        """Fetch the channel that AFK voice users get sent to.

        Returns
        -------
        typing.Optional[hikari.channels.GuildVoiceChannel]
            The AFK channel or `builtins.None` if not enabled.
```

hikari/guilds.py
```diff
@@ -1338,6 +2544,271 @@ def get_emoji(
...
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
+        hikari.errors.InternalServerError
+            If an internal error occurs on Discord while handling the request.
+        """
+        return await self.app.rest.fetch_member(self.id, self.owner_id)
+
+    async def fetch_widget_channel(self) -> typing.Optional[channels_.PartialChannel]:
```
davfsa, line 2607:
```suggestion
    async def fetch_widget_channel(self) -> typing.Optional[channels_.GuildChannel]:
```

hikari/guilds.py
```diff
@@ -945,6 +1299,858 @@ def make_icon_url(self, *, ext: typing.Optional[str] = None, size: int = 4096) -
...
+        """
+        return await self.app.rest.delete_channel(category)
+
+    async def fetch_guild(self) -> RESTGuild:
+        """Fetch the guild.
+
+        Returns
+        -------
+        hikari.guilds.RESTGuild
+            The requested guild.
```
davfsa, line 2128:
```suggestion
    async def fetch_guild(self) -> RESTGuild:
        """Fetch the updated guild object.

        Returns
        -------
        hikari.guilds.RESTGuild
            The requested guild.
```
#

@FasterSpeeding Should we have something like `fetch_self` on all objects

davfsa (changes requested):

tests/hikari/test_guilds.py
ashwinvin: forgot about these.. sorry 😓

ashwinvin:
Seems like mypy is complaining after changing the return types for the fetch_* methods

davfsa:
> 
> 
> Seems like mypy is complaining after changing the return types for the fetch_* methods

You will need to write assertions after the fetch call. You can check other `fetch_*` calls that do the same

hikari/guilds.py
FasterSpeeding: Well it doesn't apply to all models and in some cases it might introduce weird semantics like a fetch self on partial user would trickle down to a fetch self on team members which can't exactly be directly fetched so that'd likely be left as the inherited method unlike on member where it'd be overridden to fetch that member. But makes sense to try to carry on with that style where possible

hikari/guilds.py
```diff
@@ -495,6 +539,37 @@ async def fetch_self(self) -> Member:
...
+            rate-limits automatically. This includes most bucket-specific
+            rate-limits and global rate-limits. In some rare edge cases,
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
+        hikari.errors.InternalServerError
+            If an internal error occurs on Discord while handling the request.
+        """
+        member = await self.fetch_self()
```
FasterSpeeding, line 570: Member.roles is a property which retrieves the member's roles from the cache based on member.role_ids,this should be directly fetching all the roles based on member.guild_id and then filtering them based on member.role_ids

FasterSpeeding (changes requested):

hikari/guilds.py
```diff
@@ -2642,7 +2642,9 @@ async def fetch_widget_channel(self) -> typing.Optional[channels_.GuildChannel]:
         if not self.widget_channel_id:
             return None
 
-        return await self.app.rest.fetch_channel(self.widget_channel_id)
+        widget_channel = await self.app.rest.fetch_channel(self.widget_channel_id)
+        assert widget_channel is None or isinstance(widget_channel, channels_.GuildChannel)
```
davfsa, line 2646: No need to test this here (in any of the cases bellow either). `fetch_channel` will never actually return `None`

davfsa (changes requested):

tests/hikari/test_guilds.py
```diff
@@ -558,6 +864,92 @@ def test_make_banner_url_when_no_hash(self, model):
...
+
+    @pytest.mark.asyncio
+    async def test_fetch_system_channel_when_None(self, model):
+        model.system_channel_id = None
+
+        assert await model.fetch_system_channel() is None
+
+    @pytest.mark.asyncio
+    async def test_fetch_public_updates_channel(self, model):
+        mock_channel = mock.Mock(spec_set=channels_.GuildTextChannel)
```
davfsa, line 927: As I said, no need to spec_set these. A simple spec should be enough

```suggestion
        mock_channel = mock.Mock(channels_.GuildTextChannel)
```
(applies to all other mocks using spec_set too)

hikari/guilds.py
FasterSpeeding: This still applies to all of these channel methods, they shouldn't be asserting that it only be asserting that they're X not X or None since the fetch method will never return None

hikari/guilds.py
```diff
@@ -1338,6 +2544,274 @@ def get_emoji(
     def get_role(self, role: snowflakes.SnowflakeishOr[PartialRole]) -> typing.Optional[Role]:
         """Get a role from the cache by it's ID."""
 
+    async def fetch_roles(self) -> typing.Sequence[Role]:
```
FasterSpeeding, line 2547: shouldn't fetch_roles be on partial guild?

FasterSpeeding (changes requested):

davfsa:
@ashwinvin You got merge conflicts + what Faster pointed out

hikari/guilds.py
```diff
@@ -889,6 +1213,36 @@ class GuildMemberBan:
     user: users.User = attr.field(repr=True)
     """The object of the user this ban targets."""
 
+    async def fetch_user(self) -> users.User:
+        """Fetch the banned user.
+
```
FasterSpeeding, line 1218: Isn't this method redundant, last time i checked people can already do `await ban.user.fetch_self())

hikari/guilds.py
FasterSpeeding: but yeah this should probably be called fetch_self if we're being consistent

hikari/guilds.py
ashwinvin: Yes, I too think its redundant. But it was mentioned in #37, that's why I added it.

davfsa: Nah, this is redundant and can be removed. I moved the tickets from gitlab, never actually checked the content.

hikari/guilds.py
```diff
@@ -289,6 +290,49 @@ class GuildWidget:
     is_enabled: bool = attr.field(repr=True)
     """Whether this embed is enabled."""
 
+    async def fetch_channel(self) -> typing.Optional[channels_.PartialChannel]:
+        """Fetch the widget channel.
```
FasterSpeeding, line 294: Would be more concise to have this return GuildChannel

hikari/guilds.py
```diff
@@ -977,16 +1304,898 @@ def make_icon_url(self, *, ext: typing.Optional[str] = None, size: int = 4096) -
...
+            bitrate=bitrate,
+            permission_overwrites=permission_overwrites,
+            region=region,
+            category=category,
+            reason=reason,
+        )
+
+    async def delete_channel(
+        self, channel: snowflakes.SnowflakeishOr[channels_.PartialChannel]
+    ) -> channels_.PartialChannel:
```
FasterSpeeding, line 2045: GuildChannel for the return type

hikari/guilds.py
```diff
@@ -977,16 +1304,898 @@ def make_icon_url(self, *, ext: typing.Optional[str] = None, size: int = 4096) -
...
+            If an internal error occurs on Discord while handling the request.
+
+        !!! note
+            For Public servers, the set 'Rules' or 'Guidelines' channels and the
+            'Public Server Updates' channel cannot be deleted.
+        """
+        return await self.app.rest.delete_channel(channel)
+
+    async def delete_category(
+        self, category: snowflakes.SnowflakeishOr[channels_.GuildCategory]
```
FasterSpeeding, line 2088: This method probably isn't neccessary, delete_channel can just be used for this

hikari/guilds.py
```diff
@@ -977,16 +1304,898 @@ def make_icon_url(self, *, ext: typing.Optional[str] = None, size: int = 4096) -
...
+    ) -> RESTGuild:
+        """Edits the guild.
+
+        Parameters
+        ----------
+        name : hikari.undefined.UndefinedOr[builtins.str]
+            If provided, the new name for the guild.
+        verification_level : hikari.undefined.UndefinedOr[hikari.guilds.GuildVerificationLevel]
+            If provided, the new verification level.
+        default_message_notifications : hikari.undefined.UndefinedOr[hikari.guilds.GuildMessageNotificationsLevel]
```
FasterSpeeding, line 1481: You can ignore this comment since it's outside the scope of this PR, this is more just a reminder that these should be updated to also accept their str or int equivalent for the relevant RESTClient methods

hikari/presences.py
```diff
@@ -297,3 +299,63 @@ class MemberPresence:
 
     client_status: ClientStatus = attr.field(eq=False, hash=False, repr=False)
     """Platform-specific user-statuses."""
+
+    async def fetch_user(self) -> users.PartialUser:
```
FasterSpeeding, line 303:
```suggestion
    async def fetch_user(self) -> users.User:
```

hikari/guilds.py
ashwinvin: Removing that would leave newbies confused as they mostly don't read the docs and also make the class feel _incomplete_.

davfsa: It just mimics behavior, there is no need for it.

ashwinvin: I will remove it then.

FasterSpeeding (approved):

davfsa (approved):

PR #592, opened 2021-05-09
Add stage channel methods 

ashwinvin:
### Summary
Add  stage channel methods for RESTClient and GuildBuilder

### Checklist

- [x] I have made unittests according to the code I have added

### DIFF
--- a/tests/hikari/impl/test_rest.py
+++ b/tests/hikari/impl/test_rest.py
@@ -2648,6 +2648,39 @@ class TestRESTClientImplAsync:
             reason="because we need one",
         )
 
+    async def test_create_guild_stage_channel(self, rest_client):
+        guild = StubModel(123)
+        channel = mock.Mock(channels.GuildStageChannel)
+        category_channel = StubModel(789)
+        overwrite1 = StubModel(987)
+        overwrite2 = StubModel(654)
+        rest_client._create_guild_channel = mock.AsyncMock(return_value=channel)
+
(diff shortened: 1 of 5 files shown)


davfsa:
@ashwinvin Looks good. Could you mind adding testing for the rest endpoints. I will do the special endpoints ones in the future :)

ashwinvin:
Sure!

davfsa (approved):

tests/hikari/impl/test_rest.py
```diff
@@ -2648,6 +2648,39 @@ async def test_create_guild_voice_channel(self, rest_client):
...
+            "general",
+            position=1,
+            user_limit=60,
+            bitrate=64,
+            permission_overwrites=[overwrite1, overwrite2],
+            category=category_channel,
+            region="Doge Moon",
+            reason="When doge == 1$",
+        )
+        assert returned == channel
```
FasterSpeeding, line 2670: couldn't `is` be used here to check?

FasterSpeeding: oh didn't notice that, might as well just leave that style then ig

FasterSpeeding (approved):

tests/hikari/impl/test_rest.py
davfsa: These should def be using `is`. I will have to fix the other tests

PR #599, opened 2021-05-11
Convert Message.link property to Message.make_link function

FasterSpeeding:
### Summary
Convert Message.link property to Message.make_link function

This change is necessary since the lib can't always work out whether a message is in a DM or guild and the old Message.link property would default to assuming it's DM which is incorrect behaviour. This would have primarily effected applications which are REST-only or cache-less/not caching channels.

### Checklist

### DIFF
--- a/hikari/messages.py
+++ b/hikari/messages.py
@@ -640,16 +640,26 @@ class PartialMessage(snowflakes.Unique):
 
         return channel.guild_id
 
-    @property
-    def link(self) -> str:
-        """Jump link to the message.
+    def make_link(self, guild: typing.Optional[snowflakes.SnowflakeishOr[guilds.PartialGuild]] = None) -> str:
+        """Generate a jump link to this message.
+
+        Other Parameters
+        ----------------
+        guild : typing.Union[hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialGuild]]
+            Object or ID of the guild this message is in. If not provided or then
+            this will return a DM message link.
+
+            !!! note
+                This parameter is necessary since `PartialMessage.guild_id`
+                isn't returned by the REST API regardless of whether the message
+                is in a DM or not.
 
         Returns
         -------
         builtins.str
             The jump link to the message.
         """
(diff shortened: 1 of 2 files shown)


hikari/messages.py
```diff
@@ -640,16 +640,26 @@ def guild_id(self) -> typing.Optional[snowflakes.Snowflake]:
...
+                isn't returned by the REST API regardless of whether the message
+                is in a DM or not.
 
         Returns
         -------
         builtins.str
             The jump link to the message.
         """
-        guild_id_str = "@me" if self.guild_id is None else str(self.guild_id)
+        guild_id_str = "@me" if guild is None else str(int(guild))
```
davfsa, line 662: A better way of doing this might be to ignore the `guild` argument when the internal `guild_id` value is set. If the internal value is not set and `guild` is not provided, then raise an error saying that we cannot figure out the link to the message. 

*Note: This will require defaulting the `guild` argument to undefined*

FasterSpeeding: I kind of feel like that'd mush up the semantics of this function by overcomplicating it with factors outside the scope of the function. Since instead of the two cases of pass None for a dm link and a snowflake for a guild link (could make the argument required if that helps) you’d now have to
. Explicitly pass None for all dm messages
. Pass a snowflake for some partial guild messages received over the gateway
. Pass a snowflake for guild messages received over REST when the bot either isn’t caching channels or isn’t receiving guild channel events (which wouldn’t necessarily be obvious in scopes where the method’s being called)
. Pass nothing fo full guild messages received over the gateway

In the future users might be using a method which tries to work out the guild a message is in based on cached data (this will be similar to the current guild id property except it might distinguish between the cases of nothing found cached and the cached channel not having a guild id by returning a union of [UndefinedType, None, GatewayGuild]) before calling this with them deciding whether to make a rest call to get the guild id or just fail when it returns undefined.

davfsa (dismissed):

davfsa (dismissed):

davfsa (approved):

PR #606, opened 2021-05-13
Let OwnGuildIterator rely on the default limit

FasterSpeeding:
### Summary
Let OwnGuildIterator rely on the default limit
This means that it'll now return 200 guilds per request rather than 100.

### Checklist


davfsa (dismissed):
LGTM. Linting is the only issue :)

davfsa (approved):

PR #610, opened 2021-05-15
Remove git sha1 detection on HEAD

davfsa:
### Checklist


hikari/internal/ux.py
```diff
@@ -166,20 +167,28 @@ def print_banner(package: typing.Optional[str], allow_color: bool, force_color:
...
     filtered_system_bits = (s.strip() for s in system_bits if s.strip())
 
+    path = os.path.abspath(os.path.dirname(about.__file__))
+    sha1 = about.__git_sha1__
+    if sha1.casefold() == "head":
+        try:
+            # S603 - subprocess call
+            # S607 - Starting a subprocess with a partial executable path
+            sha1 = subprocess.check_output(  # noqa S603 S607
+                ["git", "rev-parse", "HEAD"], cwd=path, stderr=subprocess.DEVNULL
```
FasterSpeeding, line 180: and what if this is say installed in another git repo cause for example it's a virtual environment which was created in a cloned bot repo's top level folder and their pip is targeting master

davfsa: I don't think there is any way to fix against that, so probably best to remove this code all together, as in its other form it doesn't work either and (with the fixes) still breaks under the conditions you said

FasterSpeeding (approved):

PR #611, opened 2021-05-15
"Freeze" data classes and routes + some slotting fixes/refactor 

FasterSpeeding:
### Summary
Turns out previous assumptions on performance implications of freezing attr data classes were incorrect and the documentation on https://www.attrs.org/en/stable/how-does-it-work.html?highlight=frozen#slotted-classes is misleading. While frozen slotted classes are slower to initiate than frozen dict classes, there isn't any difference between frozen slotted classes and non-frozen slotted classes as far as this goes.

### Other features this PR introduces 
* A user ID to DM ID cache store to replace the previous caching of DM channel IDs on user objects which is now not possible with objects being immutable.
* A switch over to preferring tuples in the entity and event factory over lists to avoid passing around mutable state as much as possible

### Removals
* Removed usage of copy in the cache in most places as this is now unnecessary with objects being immutable.
* Removed hikari.iternal.attr_extensions as this only existed as an impl detail aid to the usage of copy in the cache.
* The guild info member_count, joined_at and is_large are no longer persisted between updates in the cache due to the limitation of models being immutable (this could be changed as to update the model by calling `object.__setattr__(new_guild, "field_name", old_state)` but that'd be unorthodox at best)

### Changes this PR doesn't introduce but marks as TODO:
* Possibly splitting out the Embed class into an immutable received embed object and a mutable embed builder. This means that a lot of the embed objects for now will not be defined as frozen for now.

### Barely in scope fixes/changes introduced
* Some event models were missing slot declarations and usages of attrs.define on abcs with no actual fields has been further removed in places in favour of explicitly declaring an empty init.

### Checklist


hikari/applications.py
```diff
@@ -272,7 +270,7 @@ class OwnConnection:
     """The visibility of the connection."""
 
 
-@attr.define(hash=True, kw_only=True, weakref_slot=False)
+@attr.define(frozen=True, hash=True, kw_only=True, weakref_slot=False)
```
davfsa, line 273: Instead of this, could we use `attr.frozen`? I was thinking about this when we changed to the next gen API

FasterSpeeding: sure

davfsa:
Btw, should exceptions also be frozen?

FasterSpeeding:
Gonna close this for now, since https://github.com/python-attrs/attrs/pull/817 likely brings back the performance difference between default parameter define and default parameter frozen classes

PR #619, opened 2021-05-21
Remove connector factory and connector ownership

davfsa:
### Summary
The reason behind this change is because libraries like Django or Quart create a new event loops per request (as they are threaded), making it kind of weird to `RESTApp`, as you would have to re-create per request, which is not ideal.

Before:
```py
async def some_view():
    rest_app = hikari.RESTApp()
    async with rest_app.acquire():
        ...
        
async def some_view2():
    rest_app = hikari.RESTApp()
    async with rest_app.acquire():
        ...
```

Now:
```py
rest_app = hikari.RESTApp()

async def some_view():
    async with rest_app.acquire():
        ...

async def some_view2():
    async with rest_app.acquire():
        ...
```

### Changes
  - Move TCP cache to per RESTClientImpl
  - Remove close and context managers from RESTApp

### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -29,7 +29,6 @@ RESTful functionality.
 from __future__ import annotations
 
 __all__: typing.List[str] = [
-    "BasicLazyCachedTCPConnectorFactory",
     "ClientCredentialsStrategy",
     "RESTApp",
     "RESTClientImpl",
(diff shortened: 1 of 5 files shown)


FasterSpeeding:
The REST client (along with a lot of asyncio stuff) isn't thread safe and last time I checked ClientSession attaches itself to a loop so I don't exactly see how this would ever work while being shared between threads like that

davfsa:
> 
> 
> The REST client (along with a lot of asyncio stuff) isn't thread safe and last time I checked ClientSession attaches itself to a loop so I don't exactly see how this would ever work while being shared between threads like that

This removes that exactly. It de-attaches the TCP connector from the RESTApp and moves it to the RESTClient object, which will later be event loop bound, just like the ClientSession.

hikari/impl/rest.py
```diff
@@ -357,21 +310,6 @@ def __init__(
...
-        # would potentially end up using the wrong event loop and aiohttp
-        # would then fail when creating an HTTP request.
-        if connector_factory is None:
-            connector_factory = BasicLazyCachedTCPConnectorFactory(self._http_settings)
-            connector_owner = True
-
-        self._connector_factory = connector_factory
-        self._connector_owner = connector_owner
-        self._event_loop: typing.Optional[asyncio.AbstractEventLoop] = None
         self._executor = executor
```
FasterSpeeding, line 313: do executors guarantee thread safety?

davfsa: Generally, yes:

https://stackoverflow.com/questions/45286344/is-python-threadpoolexecutor-thread-safe

FasterSpeeding (approved):

PR #621, opened 2021-05-26
Set repository version to latest and update on new release

davfsa:
### Summary
Installing hikari from the repository does not work anymore as of [PEP 517](https://www.python.org/dev/peps/pep-0517/). Due to this we have relied on editable mode (`-e`) to bypass that. With [PEP 660](https://www.python.org/dev/peps/pep-0660/), right around the corner, this will break, so we need a better alternative.

This PR will fix that by setting a new default version that will update with every release.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,10 +33,10 @@ jobs:
         run: |
           pip install wheel
 
-          pip install -e .
+          pip install .
           pip uninstall -y hikari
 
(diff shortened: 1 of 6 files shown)


FasterSpeeding (approved):

davfsa:
@FasterSpeeding Fixed some bugs and improved some other parts. Mind having another look when you can?

FasterSpeeding (approved):

PR #622, opened 2021-05-27
Replace Guild.is_nsfw with Guild.nsfw_level

davfsa:
### Checklist

*Note: Even tho this is marked as breaking, this is just because it will break any bot pulling from master. `is_nsfw` never got to be deployed in a proper version of the lib*

### DIFF
--- a/hikari/guilds.py
+++ b/hikari/guilds.py
@@ -40,6 +40,7 @@ __all__: typing.List[str] = [
     "GuildPremiumTier",
     "GuildPreview",
     "GuildMemberBan",
+    "GuildNSFWLevel",
     "Member",
     "Integration",
     "IntegrationAccount",
@@ -244,6 +245,23 @@ class GuildVerificationLevel(int, enums.Enum):
 
     VERY_HIGH = 4
     """Must have a verified phone number."""
+
+
+@typing.final
+class GuildNSFWLevel(int, enums.Enum):
+    """Represents the NSFW level of a guild."""
+
+    DEFAULT = 0
+    """Guild has not been categorized yet."""
+
+    EXPLICIT = 1
+    """Guild contains explicit NSFW content."""
+
+    SAFE = 2
+    """Guild is safe of NSFW content."""
+
+    AGE_RESTRICTED = 3
+    """Guild may contain NSFW content."""
 
 
 @attr_extensions.with_copy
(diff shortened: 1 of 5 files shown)


hikari/guilds.py
```diff
@@ -246,6 +247,23 @@ class GuildVerificationLevel(int, enums.Enum):
...
+    """Represents the NSFW level of a guild."""
+
+    DEFAULT = 0
+    """Default level of NSFW (per channel)."""
+
+    EXPLICIT = 1
+    """Guild contains explicit NSFW content."""
+
+    SAFE = 2
+    """Guild is safe of NSFW content."""
```
FasterSpeeding, line 261: So does this mean that there's no channels marked as nsfw in the guild or that it's not age restricted?

hikari/guilds.py
```diff
@@ -246,6 +247,23 @@ class GuildVerificationLevel(int, enums.Enum):
...
+
+    AGE_RESTRICTED = 2
```
FasterSpeeding, line 263: shouldn't this be 3?

hikari/guilds.py
```diff
@@ -246,6 +247,23 @@ class GuildVerificationLevel(int, enums.Enum):
     """Must have a verified phone number."""
 
 
+@typing.final
+class GuildNSFWLevel(int, enums.Enum):
+    """Represents the NSFW level of a guild."""
+
+    DEFAULT = 0
+    """Default level of NSFW (per channel)."""
```
FasterSpeeding, line 255: what does this mean?

hikari/guilds.py
```diff
@@ -246,6 +247,23 @@ class GuildVerificationLevel(int, enums.Enum):
...
+
+    EXPLICIT = 1
+    """Guild contains explicit NSFW content."""
+
+    SAFE = 2
+    """Guild is safe of NSFW content."""
+
+    AGE_RESTRICTED = 2
+    """Guild is age restricted due to NSFW content."""
```
FasterSpeeding, line 264: Could this go into more detail about this actually means and maybe link to a support article about it?

hikari/guilds.py
davfsa: Yep, my bad

hikari/guilds.py
davfsa: At the current time, there is no way of setting this (although the [article](https://support.discord.com/hc/en-us/articles/1500005389362-NSFW-Server-Designation) does mention this will come in the future)

I will rewrite the docstrings for all of them to see if I can make them clear :)

davfsa:
@FasterSpeeding Changed all the docstring to try and make them clearer

FasterSpeeding (approved):

PR #627, opened 2021-05-30
Add a token type property to the rest interface to allow for token introspection

FasterSpeeding:
### Summary
Add a token type property to the rest interface to allow for token introspection

### Checklist

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -64,7 +64,7 @@ class TokenStrategy(abc.ABC):
     @property
     @abc.abstractmethod
     def token_type(self) -> typing.Union[applications.TokenType, str]:
-        """Type of token this token strategy returns.
+        """Type of token this strategy returns.
 
         Returns
         -------
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -61,9 +61,20 @@ class TokenStrategy(abc.ABC):
 
     __slots__: typing.Sequence[str] = ()
 
+    @property
+    @abc.abstractmethod
+    def token_type(self) -> typing.Union[applications.TokenType, str]:
+        """Type of token this token strategy returns.
+
+        Returns
+        -------
+        typing.Union[hikari.applications.TokenType, builtins.str]
+            The type of token this strategy returns.
+        """
+
     @abc.abstractmethod
     async def acquire(self, client: RESTClient) -> str:
-        """Acquire an authorization token (with the prefix).
+        """Acquire an authorization token (including the prefix).
 
         Returns
         -------
@@ -100,16 +111,16 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
 
     @property
     @abc.abstractmethod
-    def token(self) -> typing.Union[str, TokenStrategy, None]:
-        """Token this client is using for most requests.
+    def token_type(self) -> typing.Union[str, applications.TokenType, None]:
+        """Type of token this client is using for most requests.
 
         Returns
         -------
-        typing.Union[builtins.str, TokenStrategy, builtins.None]
-            The token this client is using for most requests.
+        typing.Union[builtins.str, hikari.applications.TokenType, builtins.None]
+            The type of token this client is using for most requests.
 
-            If this is `builtins.None` then this client will only work for some
-            endpoints such as public and webhook ones.
+            If this is `builtins.None` then this client will likely only work
+            for some endpoints such as public and webhook ones.
         """
 
     @abc.abstractmethod
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -98,6 +98,20 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
 
     __slots__: typing.Sequence[str] = ()
 
+    @property
+    @abc.abstractmethod
+    def token(self) -> typing.Union[str, TokenStrategy, None]:
+        """Token this client is using for most requests.
+
+        Returns
+        -------
+        typing.Union[builtins.str, TokenStrategy, builtins.None]
+            The token this client is using for most requests.
+
+            If this is `builtins.None` then this client will only work for some
+            endpoints such as public and webhook ones.
+        """
+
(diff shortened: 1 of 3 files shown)


hikari/api/rest.py
```diff
@@ -96,6 +96,20 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
...
+        """Token this client is using for most requests.
+
+        Returns
+        -------
+        typing.Union[builtins.str, TokenStrategy, builtins.None]
+            The token this client is using for most requests.
+
+            If this is `builtins.None` then this client will only work for some
+            endpoints such as public and webhook ones.
+        """
```
davfsa, line 111: I dont think this is a good thing to expose in any way

FasterSpeeding: Why not.

davfsa: Tokens are private and not something that should be in the public interface. I know we already store it in `_token`, but that is private internal detail, not something that we should expose

FasterSpeeding: Why shouldn't they be exposed and if a REST client doesn't have a token it can return None here

FasterSpeeding: Still not sure exactly what you're getting at with that private stuff but if you think keeping tokens "private" in this language is going to help with security you're just wrong, most people will be targetting the main implementation and it doesn't take a genius to just do `rest._token`. This doesn't change the fact that people can access the token, this just adds a more formalised way for people who want to introspect it.
Plus fwiw this is the only Discord API lib i've seen that doesn't have a way to get the rest client's active token

davfsa (approved):

hikari/api/rest.py
```diff
@@ -96,6 +96,20 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
 
     __slots__: typing.Sequence[str] = ()
 
+    @property
+    @abc.abstractmethod
+    def token(self) -> typing.Union[str, TokenStrategy, None]:
+        """Token this client is using for most requests.
+
```
davfsa, line 103:
```suggestion
        """Token this client is using for most requests.

        !!! warning
            This should be kept private at all times!
```

davfsa (changes requested):

davfsa:
I'm just going to leave this here as we were discussing it in the server.

Would exposing the token type only be good enough? You can act on that information, no need to expose the whole token

FasterSpeeding:
I'll refactor this to only return the "token type" at one point, will probably have it return Bearer if the token type is a strategy or otherwise will have to add a token type property to token strategy which this relays but will have to spend a bit more time to consider this

davfsa (dismissed):
LGMT :thumbsup:

hikari/impl/rest.py
```diff
@@ -431,15 +488,21 @@ def __init__(
         self._http_settings = http_settings
         self._proxy_settings = proxy_settings
 
-        self._token: typing.Union[str, rest_api.TokenStrategy, None]
+        self._token: typing.Union[str, rest_api.TokenStrategy, None] = None
+        self._token_type: typing.Optional[str] = None
```
davfsa, line 492: Are we sure we want to allow passing an empty token type?

FasterSpeeding: This only checks that token_type is or isn't None in certain cases to ensure that they passed a token type, these checks aren't here to make sure that they pass what we think is a correct token type for their scenario and if we really wanted to verify they were doing that we'd be checking against specific constants but that's not as flexible and may not account for scenarios we're not aware of down the line

davfsa: Sure ig. I just dont think we should allow not passing the type of the token

FasterSpeeding: Weird point but fun fact if they passed prefix + the whole token in as token prefix then passed " " or "" for the actual token discord would accept that as valid auth

hikari/impl/rest.py
```diff
@@ -326,17 +330,74 @@ def http_settings(self) -> config.HTTPSettings:
...
+            `token`.
+
+        Returns
+        -------
+        RESTClientImpl
+            An instance of the REST client.
+
+        Raises
+        ------
+        ValueError
```
davfsa, line 389:
```suggestion
        builtins.ValueError
```

hikari/impl/rest.py
```diff
@@ -376,11 +437,21 @@ class RESTClientImpl(rest_api.RESTClient):
...
+        This should be left as `builtins.None` when either
+        `hikari.api.rest.TokenStrategy` or `builtins.None` is passed for
+        `token`.
     rest_url : builtins.str
         The HTTP API base URL. This can contain format-string specifiers to
         interpolate information such as API version in use.
+
+    Raises
+    ------
+    ValueError
```
davfsa, line 452:
```suggestion
    builtins.ValueError
```

davfsa (commented):
LGTM, just two docstring fixes

hikari/impl/rest.py
```diff
@@ -182,6 +182,10 @@ def scopes(self) -> typing.Sequence[typing.Union[applications.OAuth2Scope, str]]
         """
         return self._scopes
 
+    @property
+    def token_type(self) -> typing.Union[applications.TokenType, str]:
+        return applications.TokenType.BEARER
```
davfsa, line 187:
```suggestion
    @property
    def token_type(self) -> applications.TokenType:
        return applications.TokenType.BEARER
```

hikari/impl/rest.py
FasterSpeeding: These do seem to be resolving without prefixing them as `builtins.{0}` though

davfsa: Ik, but we use them everywhere else because some of them dont seem to resolve (I have already tried to fix this several times). It is purely for consistency sake

FasterSpeeding: Do these cases where it doesn't resolve even still exist though

davfsa (approved):

PR #629, opened 2021-05-30
Make voice and rest components multi-use

FasterSpeeding:
### Summary
* This adds an is_alive property to Voice component along with ComponentStateConflictError cases in the main implementation to avoid erronous behaviour from connections being added after it stops tracking voice events.
* ~~This also explicitly raises ComponentStateConflictError in BotApp if it's already been used instead of letting the rest client raise it on fetch_gateway_bot and documents the fact that BotApp is one use~~

### Checklist

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -85,10 +85,6 @@ class TokenStrategy(abc.ABC):
             prefix.
         """
 
-    @abc.abstractmethod
-    async def close(self) -> None:
-        """Close this access token handler."""
-
     def invalidate(self, token: typing.Optional[str]) -> None:
         """Invalidate the cached token in this handler.
 
@@ -110,6 +106,11 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
     """Interface for functionality that a REST API implementation provides."""
 
     __slots__: typing.Sequence[str] = ()
+
+    @property
+    @abc.abstractmethod
+    def is_alive(self) -> bool:
+        """Whether this component is alive."""
 
     @property
     @abc.abstractmethod
--- a/hikari/impl/event_manager.py
+++ b/hikari/impl/event_manager.py
@@ -32,6 +32,7 @@ import random
 import typing
 
 from hikari import channels
+from hikari import errors
 from hikari import intents as intents_
 from hikari import presences
 from hikari import snowflakes
@@ -54,6 +55,21 @@ def _fixed_size_nonce() -> str:
     head = time.monotonic_ns().to_bytes(8, "big")
     tail = random.getrandbits(92).to_bytes(12, "big")
     return base64.b64encode(head + tail).decode("ascii")
+
+
+async def _request_guild_members(
+    shard: gateway_shard.GatewayShard,
+    guild: guilds.PartialGuild,
+    *,
+    include_presences: bool,
+    nonce: str,
+) -> None:
+    try:
+        await shard.request_guild_members(guild, include_presences=include_presences, nonce=nonce)
+
+    # Ignore errors raised by a shard shutting down
+    except errors.ComponentStateConflictError:
+        pass
 
 
 class EventManagerImpl(event_manager_base.EventManagerBase):
@@ -167,8 +183,8 @@ class EventManagerImpl(event_manager_base.EventManagerBase):
                 # We create a task here instead of awaiting the result to avoid any rate-limits from delaying dispatch.
                 nonce = f"{shard.id}.{_fixed_size_nonce()}"
                 event.chunk_nonce = nonce
-                coroutine = shard.request_guild_members(
-                    event.guild, include_presences=bool(presences_declared), nonce=nonce
+                coroutine = _request_guild_members(
+                    shard, event.guild, include_presences=bool(presences_declared), nonce=nonce
                 )
                 asyncio.create_task(coroutine, name=f"{shard.id}:{event.guild.id} guild create members request")
 
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -424,7 +424,7 @@ class RESTApp(traits.ExecutorAware):
 
 @attr.define()
 class _LiveAttributes:
-    """Fields which are only present within `RESTClientImpl` while it's "live".
+    """Fields which are only present within `RESTClientImpl` while it's "alive".
 
     !!! note
         This must be started within an active asyncio event loop.
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -424,6 +424,12 @@ class RESTApp(traits.ExecutorAware):
 
 @attr.define
 class _LiveAttributes:
+    """Fields which are only present within `RESTClientImpl` while it's "live".
+
+    !!! note
+        This must be started within an active asyncio event loop.
+    """
+
     buckets: buckets_.RESTBucketManager
     client_session: aiohttp.ClientSession
     closed_event: asyncio.Event
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -474,8 +474,8 @@ class _LiveAttributes:
         self.is_closing = True
         self.closed_event.set()
         self.buckets.close()
-        await self.client_session.close()
         self.global_rate_limit.close()
+        await self.client_session.close()
         await self.tcp_connector.close()
 
     def still_alive(self) -> _LiveAttributes:
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -671,8 +671,6 @@ class RESTClientImpl(rest_api.RESTClient):
     ) -> typing.Union[None, data_binding.JSONObject, data_binding.JSONArray]:
         # Make a ratelimit-protected HTTP request to a JSON endpoint and expect some form
         # of JSON response.
-        live_attributes = self._get_live_attributes()
-
         headers = data_binding.StringMapBuilder()
         headers.setdefault(_USER_AGENT_HEADER, _HTTP_USER_AGENT)
 
@@ -695,11 +693,11 @@ class RESTClientImpl(rest_api.RESTClient):
         while True:
             try:
                 uuid = time.uuid()
-                async with live_attributes.buckets.acquire(compiled_route):
+                async with self._get_live_attributes().buckets.acquire(compiled_route):
                     # Buckets not using authentication still have a global
                     # rate limit, but it is different from the token one.
                     if not no_auth:
-                        await live_attributes.global_rate_limit.acquire()
+                        await self._get_live_attributes().global_rate_limit.acquire()
 
                     if _LOGGER.isEnabledFor(ux.TRACE):
                         _LOGGER.log(
@@ -713,7 +711,7 @@ class RESTClientImpl(rest_api.RESTClient):
                         start = time.monotonic()
 
                     # Make the request.
-                    response = await live_attributes.client_session.request(
+                    response = await self._get_live_attributes().client_session.request(
                         compiled_route.method,
                         url,
                         headers=headers,
@@ -739,7 +737,7 @@ class RESTClientImpl(rest_api.RESTClient):
                         )
 
                     # Ensure we are not rate limited, and update rate limiting headers where appropriate.
-                    await self._parse_ratelimits(compiled_route, response, live_attributes)
+                    await self._parse_ratelimits(compiled_route, response, self._get_live_attributes())
 
                 # Don't bother processing any further if we got NO CONTENT. There's not anything
                 # to check.
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -430,12 +430,13 @@ class _LiveAttributes:
         This must be started within an active asyncio event loop.
     """
 
-    buckets: buckets_.RESTBucketManager
-    client_session: aiohttp.ClientSession
-    closed_event: asyncio.Event
+    buckets: buckets_.RESTBucketManager = attr.field()
+    client_session: aiohttp.ClientSession = attr.field()
+    closed_event: asyncio.Event = attr.field()
     # We've been told in DAPI that this is per token.
-    global_rate_limit: rate_limits.ManualRateLimiter
-    tcp_connector: aiohttp.TCPConnector
+    global_rate_limit: rate_limits.ManualRateLimiter = attr.field()
+    tcp_connector: aiohttp.TCPConnector = attr.field()
+    is_closing: bool = attr.field(default=False, init=False)
 
     @classmethod
     def build(
@@ -470,11 +471,19 @@ class _LiveAttributes:
         )
 
     async def close(self) -> None:
+        self.is_closing = True
         self.closed_event.set()
         self.buckets.close()
         await self.client_session.close()
         self.global_rate_limit.close()
         await self.tcp_connector.close()
+
+    def still_alive(self) -> _LiveAttributes:
+        """Chained method used to Check if `close` has been called before using this object's resources."""
+        if self.is_closing:
+            raise errors.ComponentStateConflictError("The REST client was closed mid-request")
+
+        return self
 
 
 class RESTClientImpl(rest_api.RESTClient):
@@ -671,6 +680,7 @@ class RESTClientImpl(rest_api.RESTClient):
     ) -> typing.Union[None, data_binding.JSONObject, data_binding.JSONArray]:
         # Make a ratelimit-protected HTTP request to a JSON endpoint and expect some form
         # of JSON response.
+        live_attributes = self._get_live_attributes()
         headers = data_binding.StringMapBuilder()
         headers.setdefault(_USER_AGENT_HEADER, _HTTP_USER_AGENT)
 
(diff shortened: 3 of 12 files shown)


davfsa:
Wouldn't it be better to correctly cleanup everything to make sure they can be used again. I can now see the point you made on Discord about the rest client. We want to keep it a once time use thing, so we could just re-create the object in `Bot.start` (or we can think of a better way to do this)

FasterSpeeding:
@davfsa made some changes to this, currently the lazy behaviour of the voice and rest component standard implementations is being preserved but a "start" method is only added to the standard voice implementation in favour of implicitly re-starting it. The only changes this makes to the interfaces is adding "is_alive" properties to rest and voice components + removing "close" from the token strategy.

Not really overly happy with the implicit restarting behaviour in the REST component since it kinda defeats the purpose of having a "close" method and makes it more like a "clear" method. A part of me just wants to get rid of the implicit lazy starting all together and have a start, stop method approach for the main implementations without having the start and stop methods on the interfaces as to allow other implementations to be lazy (since there are definitely use cases for one-use lazily started REST clients just a bot ain't it).

FasterSpeeding:
This does also have the side effect of meaning that cached rate limit data (such as routes to hashes) isn't preserved between restarts but tbh these hashes aren't guaranteed to be constant (e.g. they could change after discord pushes changes to their REST api) and I don't think rediscovering them should hurt too much but this could switch to an approach where RESTBucketManager is "cleared" rather than closed if we want to preserve that data

FasterSpeeding:
@davfsa btw do you have any comments for the 2 previous messages?

hikari/api/rest.py
```diff
@@ -70,10 +70,6 @@ async def acquire(self, client: RESTClient) -> str:
             prefix.
         """
 
-    @abc.abstractmethod
-    async def close(self) -> None:
```
FasterSpeeding, line 74: Just a note for why I'm removing this, since this is intended for use with the OAuth2 flow and the refresh tokens Discord provides either don't expire or have a large enough expire time that nobody's noticed yet there's no need for any implementations of this to do anything which would need to be "closed" (i.e. have a background task)

hikari/impl/rest.py
```diff
@@ -473,6 +514,14 @@ async def close(self) -> None:
         # https://github.com/aio-libs/aiohttp/issues/1925#issuecomment-715977247
         await asyncio.sleep(0.25)
 
+    def _get_live_attributes(self) -> _LiveAttributes:
+        if not self._live_attributes:
+            self._live_attributes = _LiveAttributes.build(
+                self._max_rate_limit, self._http_settings, self._proxy_settings
+            )
+
+        return self._live_attributes
```
davfsa, line 523: Why cant this be in the `__init__`?

davfsa:
> Not really overly happy with the implicit restarting behaviour in the REST component since it kinda defeats the purpose of having a "close" method and makes it more like a "clear" method (one that closes all active connections without stopping further ones from being started)

When we close the REST component, it becomes fully closed and there is no way to re-open it unless you create a new RESTImpl. This could be kept that way or have a reset state function


> This does currently have the side effect that cached rate limit data (such as routes to hashes) isn't preserved between rest client restarts but tbh these hashes aren't guaranteed to be constant (e.g. they could change after discord pushes changes to their REST API) and I don't think rediscovering them should hurt too much but this could switch to an approach where RESTBucketManager is "cleared" rather than closed if we want to preserve that data

This isnt really important if we have to re-populate as it would be the equivalent of starting the bot again.

hikari/impl/rest.py
FasterSpeeding: it builds objects which can only be built when within a live event loop (ClientSession, Event, RESTBucketManager and ManualRatelimiter)

FasterSpeeding:
> When we close the REST component, it becomes fully closed and there is no way to re-open it unless you create a new RESTImpl. This could be kept that way or have a reset state function

That's basically just a start method though

FasterSpeeding:
What use case does having the main REST implementation start-up lazily benefit anyway?

hikari/impl/rest.py
```diff
@@ -351,6 +348,61 @@ def acquire(
...
+    ) -> _LiveAttributes:
+        """Build a live attributes object.
+
+        !!! warning
+            This can only be called when the current thread has an active
+            asyncio loop.
+        """
+        # This asserts that this is called within an active event loop.
+        asyncio.get_running_loop()
+        tcp_connector = net.create_tcp_connector(http_settings)
```
FasterSpeeding, line 378: does TCP connector have to be handled this way or can it be persisted between starts and initiated in `RESTClientImpl.__init__`?

FasterSpeeding: Tbf don't really care

hikari/impl/rest.py
```diff
@@ -351,6 +348,61 @@ def acquire(
...
+            client_session=client_session,
+            closed_event=asyncio.Event(),
+            global_rate_limit=rate_limits.ManualRateLimiter(),
+            tcp_connector=tcp_connector,
+        )
+
+    async def close(self) -> None:
+        self.buckets.close()
+        await self.client_session.close()
+        self.closed_event.set()
```
FasterSpeeding, line 401: call closed_event.set first

davfsa:
Been thinking a bit, maybe REST start/close will fix a lot of other issues we are having atm, so wouldnt be bad to implement it

hikari/impl/rest.py
```diff
@@ -588,37 +647,6 @@ def __exit__(
     ) -> None:
         return None
```
davfsa, line 648: Closing here? (tbf, this just hit me as I was reviewing the pr)

hikari/impl/rest.py
```diff
@@ -676,8 +703,7 @@ async def _request(
                         start = time.monotonic()
 
                     # Make the request.
-                    session = self._acquire_client_session()
-                    response = await session.request(
+                    response = await live_attributes.client_session.request(
```
davfsa, line 706: What will happen if the live attributes are closed before this is done?

hikari/impl/shard.py
```diff
@@ -524,6 +524,12 @@ async def get_user_id(self) -> snowflakes.Snowflake:
             raise RuntimeError("user_id was not known, this is probably a bug")
         return self._user_id
 
+    def _get_ws(self) -> _GatewayTransport:
+        if not self._ws:
+            raise errors.ComponentStateConflictError("Shard is shutting down")
+
+        return self._ws
```
davfsa, line 531: Would a better approach for this be the WS be raising the `ComponentStateConflictError`?

Then this would leave all the `self._get_ws()` calls to 1 which will only check if the shard has an available ws and use it. Then it is up to the ws to handle it being on or not.

hikari/impl/voice.py
```diff
@@ -63,16 +63,40 @@ def __init__(self, app: traits.BotAware) -> None:
...
+            raise errors.ComponentStateConflictError("Component cannot be used while it's not alive")
+
+    async def _disconnect(self) -> None:
         if self._connections:
             _LOGGER.info("shutting down %s voice connection(s)", len(self._connections))
+            # We rely on the assumption that _on_connection_close will be called here rather than explicitly
+            # emptying self._connections.
             await asyncio.gather(*(c.disconnect() for c in self._connections.values()))
 
+    async def disconnect(self) -> None:
```
davfsa, line 83: Why 2 functions? Cant we just keep it to 1?

hikari/impl/rest.py
FasterSpeeding: It is closed in `__aexit__`, `__exit__` only exists as a part of `__enter__` to give better errors if someone tries to use this as a sync context manager

hikari/impl/rest.py
FasterSpeeding: Changed this so that _get_live_attributes is called multiple times in _request to account for it being closed mid-request

hikari/impl/voice.py
FasterSpeeding: Added an _is_closing field to avoid calls being made to the client while it's closing but _disconnect is still needed to avoid a state conflict error being raised when this is called in close

hikari/impl/shard.py
FasterSpeeding: That honestly seems like a more complex fix than the one currently present in this PR since it'd probably involve some sort of restructuring

hikari/impl/event_manager.py
```diff
@@ -56,6 +57,21 @@ def _fixed_size_nonce() -> str:
...
+    guild: guilds.PartialGuild,
+    *,
+    include_presences: bool,
+    nonce: str,
+) -> None:
+    try:
+        await shard.request_guild_members(guild, include_presences=include_presences, nonce=nonce)
+
+    # Ignore errors raised by a shard shutting down
+    except errors.ComponentStateConflictError:
```
davfsa, line 71: Do we want to make more precise errors for each component that derive from `ComponentStateConflictError`?

hikari/impl/rest.py
```diff
@@ -425,6 +422,61 @@ def acquire(
...
+            global_rate_limit=rate_limits.ManualRateLimiter(),
+            tcp_connector=tcp_connector,
+        )
+
+    async def close(self) -> None:
+        self.closed_event.set()
+        self.buckets.close()
+        await self.client_session.close()
+        self.global_rate_limit.close()
+        await self.tcp_connector.close()
```
davfsa, line 477: Cant we have all this logic inside the `RESTClientImpl`?

hikari/impl/shard.py
davfsa: Considering we are already using our own `_GatewayTransport`, I dont think it will be that hard, as just all the methods in it will have to check if it alive before executing (as we do in the other components)

hikari/impl/voice.py
```diff
@@ -63,16 +63,46 @@ def __init__(self, app: traits.GatewayBotAware) -> None:
...
+        self._check_if_alive()
+        await self._disconnect()
+
     async def close(self) -> None:
-        await self.disconnect()
+        self._check_if_alive()
+        self._is_closing = True
         self._app.event_manager.unsubscribe(voice_events.VoiceEvent, self._on_voice_event)
+        await self._disconnect()
```
davfsa, line 95: Unsubscribing from the event should go after we call `_disconnect`

hikari/impl/voice.py
davfsa: We could also change it to allow close to always be called, even if the component is not running, which would probably be best now that I think about it

hikari/impl/voice.py
FasterSpeeding: Why

hikari/impl/event_manager.py
FasterSpeeding: what's the use case for that

FasterSpeeding:
> We could also change it to allow close to always be called, even if the component is not running, which would probably be best now that I think about it

why

hikari/impl/voice.py
davfsa: Hmmm, true. Probably best to do it this way then

hikari/impl/event_manager.py
davfsa: Will probably clarify all these excepts, as it would be something like `errors.ShardStateConflictError`

hikari/impl/voice.py
davfsa: Something that just popped my mind, we dont actually have a disconnect a specific channels, we only have a full disconnect from all channels. Should we change disconnect to allow passing a channel to disconnect from and moving the closing all connections to `close`?

(This would be done in another pr, since it is a bit out of scope for this one)

hikari/impl/rest.py
FasterSpeeding: Sure but I prefer the separation of having the logic which manages those fields be in the same place they're stored, kinda helps with readability imo tbh

hikari/impl/voice.py
FasterSpeeding: Right now they could just get the specific connection handler from VoiceClient.connections and then call disconnect on that which should have the same behaviour

hikari/impl/event_manager.py
FasterSpeeding: Eh, at least under the current interface these errors don't really ever bobble up like that and unless there's a reason to want to differentiate between them at runtime (can't really think of one myself) the error's message should already be enough to clarify where it's from. What we could do with however is added documentation on places where it might be raised since right now we're lacking on that

hikari/impl/rest.py
davfsa: Having all the `_get_live_attributes()` calls doesnt look nice to read (imo)

hikari/impl/voice.py
davfsa: Ok, after this is merged I will open a pr to do this 👍

FasterSpeeding:
> Considering we are already using our own _GatewayTransport, I dont think it will be that hard, as just all the methods in it will have to check if it alive before executing (as we do in the other components)

Keep in mind _GatewayTransport does just extend aiohttp.ClientWebSocketResponse to add some extra logic for handling stuff like serialization, and connection and in itself might already have an error it raises when certain calls are made and it's considered closed

hikari/impl/shard.py
FasterSpeeding: Keep in mind _GatewayTransport does just extend aiohttp.ClientWebSocketResponse to add some extra logic for handling stuff like serialization, and connection and in itself might already have an error it raises when certain calls are made and it's considered closed

hikari/impl/rest.py
FasterSpeeding: Less calls can be made to that if LiveAttributes itself internally tracks whether it's been closed or not

FasterSpeeding: Fwiw I've pushed another approach which lets LiveAttributes track whether it's been told to close or not internally, this comes with the added benefit of letting us distinguish between when a call fails because the client was closed and when a call fails because the client was closed as the request was being made in the error message

hikari/impl/shard.py
davfsa: Considering it is private implementation and we only make use of the methods in _GatewayTransport, we could add the validation there

FasterSpeeding: And then you've also gotta refactor how the Shard object handles the ws and add stuff like is alive methods to the gateway transport etc etc and check to see whether its reusable to see if these refactors should be replacing it on restart or not and if it is reusable we'll also have to add a start method to it and it's all just a lot of extra effort for no added benefit

hikari/impl/rest.py
```diff
@@ -693,11 +703,11 @@ async def _request(
         while True:
             try:
                 uuid = time.uuid()
-                async with self._get_live_attributes().buckets.acquire(compiled_route):
+                async with live_attributes.still_alive().buckets.acquire(compiled_route):
```
davfsa, line 706: Tbh, this is still kinda meh. Would doing this in the  `__getattribute__` in live_attributes work?

```py
    def __getattribute__(self, attr: str) -> typing.Any:
        if super().__getattribute__("is_closing"):
            raise errors.ComponentStateConflictError("The REST client was closed mid-request")

        return super().__getattribute__(attr)

Comment (FasterSpeeding) [reply hikari/impl/rest.py:706]: sounds really implicit and bad for readability

Comment (davfsa) [reply hikari/impl/rest.py:706]: Having a `still_alive` check everywhere I find kind of unreadable (considering that _LiveAttributes could manage this with no need). A comment stating could be added on to specify this behavior

Comment (FasterSpeeding) [reply hikari/impl/rest.py:706]: > (considering that _LiveAttributes could manage this with no need)

Just because you can do something doesn't mean you should and are you really in-favour of stuff like having properties raise errors that aren't AttributeError, would you introduce that to the public api?

Comment (davfsa) [reply hikari/impl/rest.py:706]: > > (considering that _LiveAttributes could manage this with no need)
> 
> Just because you can do something doesn't mean you should and are you really in-favour of stuff like having properties raise errors that aren't AttributeError, would you introduce that to the public api?

No but, as I said, this is private detail, so we are kind of excused by doing something like this here. Another idea I just thought of is making properties for them and making the checks there (this should have been the first idea, sorry for that)

Comment (FasterSpeeding) [reply hikari/impl/rest.py:706]: > Another idea I just thought of is making properties for them and making the checks there (this should have been the first idea, sorry for that)

That's a lot of boilerplate and still as implicit and unreadable due to how it hides a function call which raises behind a property

Comment (davfsa) [reply hikari/impl/rest.py:706]: What about get methods that check if still alive and return the object?

```py
def get_buckets() -> ...:
   self._still_alive()
   return self._buckets
```

FasterSpeeding: how would you avoid making unnecessary calls to the check if it's already been verified that's it's active since the last yield to the event loop with that pattern

davfsa: Tbh, it is always better to be safe than sorry. But ig you are right.

hikari/impl/shard.py
```diff
@@ -527,6 +527,12 @@ async def get_user_id(self) -> snowflakes.Snowflake:
             raise RuntimeError("user_id was not known, this is probably a bug")
         return self._user_id
 
+    def _get_ws(self) -> _GatewayTransport:
+        if not self._ws:
+            raise errors.ComponentStateConflictError("Shard is shutting down")
```
davfsa, line 532:
```suggestion
            raise errors.ComponentStateConflictError("Shard is closed")
```

FasterSpeeding: If it was already closed when whatever call triggered this was started then _check_if_alive should've raised before it got this far

davfsa (approved):

PR #636, opened 2021-06-01
Start filtering events before unmarshalling

FasterSpeeding:
### Summary
This change looks into how we may be able to avoid deserializing event data when it isn't being listened to or used to set state to lower the inherent strain of shard dispatch. While intents can help with this by lowering the amount of events a bot receives, intents can't always be used to prevent unwanted events as some events just aren't covered by it and some of it's categories (such as GUILDS) are broad enough to have the potential to still lead to unnecessary events being received.

Another benefit this has over intents is that while intents are behaviour that's statically defined for the entire runtime of a shard, the behaviour of this filtering is dynamic and will shift based on what's being listened for and cached.

Due to the broad scope of the GUILD_CREATE and GUILD_UPDATE events, the raw consumer methods for these events were modified to internally filter based on the cache as well as also filtering based on registered event listeners (this was done in-favour of the generic filtering system used for other raw event consumers). This internal filtering allows the different entities present in these events to only be unmarshalled if they're specifically being cached when there's no listeners registered for the relevant guild events.

### Checklist

### DIFF
--- a/hikari/api/cache.py
+++ b/hikari/api/cache.py
@@ -92,13 +92,7 @@ class Cache(abc.ABC):
     @property
     @abc.abstractmethod
     def settings(self) -> config.CacheSettings:
-        """Get the configured settings for this cache.
-
-        Returns
-        -------
-        hikari.config.CacheSettings
-            The configured settings for this cache.
-        """
+        """Get the configured settings for this cache."""
 
     @abc.abstractmethod
     def get_dm_channel_id(
--- a/hikari/api/cache.py
+++ b/hikari/api/cache.py
@@ -32,6 +32,7 @@ from hikari import iterators
 
 if typing.TYPE_CHECKING:
     from hikari import channels
+    from hikari import config
     from hikari import emojis
     from hikari import guilds
     from hikari import invites
@@ -87,6 +88,17 @@ class Cache(abc.ABC):
     """
 
     __slots__: typing.Sequence[str] = ()
+
+    @property
+    @abc.abstractmethod
+    def settings(self) -> config.CacheSettings:
+        """Get the configured settings for this cache.
+
+        Returns
+        -------
+        hikari.config.CacheSettings
+            The configured settings for this cache.
+        """
 
     @abc.abstractmethod
     def get_dm_channel_id(
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -58,6 +58,7 @@ class Event(abc.ABC):
     __slots__: typing.Sequence[str] = ()
 
     __subclasses: typing.ClassVar[typing.Set[typing.Type[Event]]]
+    __on_new_subclass: typing.Final[typing.ClassVar[typing.Set[typing.Callable[[typing.Type[Event]], None]]]] = set()
 
     def __init_subclass__(cls) -> None:
         super().__init_subclass__()
@@ -73,6 +74,9 @@ class Event(abc.ABC):
                     # This should only ever apply to the base Event class
                     parent_cls.__subclasses = {parent_cls, cls}
 
+        for callback in cls.__on_new_subclass:
+            callback(cls)
+
     @property
     @abc.abstractmethod
     def app(self) -> traits.RESTAware:
@@ -85,8 +89,19 @@ class Event(abc.ABC):
         """
 
     @classmethod
-    def subclasses(cls) -> typing.Set[typing.Type[Event]]:
-        return cls.__subclass
+    def subclasses(cls) -> typing.AbstractSet[typing.Type[Event]]:
+        """Set of the event class' subclasses."""
+        return cls.__subclasses
+
+    @classmethod
+    def on_new_subclass(cls, listener: typing.Callable[[typing.Type[Event]], None], /) -> None:
+        """Add a callback to be called whenever a new `Event` subclass is made."""
+        cls.__on_new_subclass.add(listener)
+
+    @classmethod
+    def remove_on_new_subclass(cls, listener: typing.Callable[[typing.Type[Event]], None], /) -> None:
+        """Remove a new subclass callback."""
+        cls.__on_new_subclass.remove(listener)
 
 
 def get_required_intents_for(event_type: typing.Type[Event]) -> typing.Collection[intents.Intents]:
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -57,25 +57,21 @@ class Event(abc.ABC):
 
     __slots__: typing.Sequence[str] = ()
 
-    __subclasses: typing.ClassVar[typing.Set[typing.Type[Event]]]
-    __on_new_subclass: typing.Final[typing.ClassVar[typing.Set[typing.Callable[[typing.Type[Event]], None]]]] = set()
+    __dispatches: typing.ClassVar[typing.Tuple[typing.Type[Event], ...]]
 
     def __init_subclass__(cls) -> None:
         super().__init_subclass__()
-        # For convenience's sake we include a class in the set of it's subclasses.
-        # This also matches how issubclass(Foo, Foo) returns True
-        cls.__subclasses = {cls}
+        # hasattr doesn't work with protected variables in this case so we use a try except.
+        # we need to set Event's __dispatches when the first subclass is made as Event itself cannot
+        # be included in a set literal on itself due to not existing yet.
+        try:
+            Event.__dispatches
+        except AttributeError:
+            Event.__dispatches = (Event,)
 
-        for parent_cls in cls.mro():
-            if parent_cls is not cls and issubclass(parent_cls, Event):
-                try:
-                    parent_cls.__subclasses.add(cls)
-                except AttributeError:
-                    # This should only ever apply to the base Event class
-                    parent_cls.__subclasses = {parent_cls, cls}
+        mro = cls.mro()
 
-        for callback in cls.__on_new_subclass:
-            callback(cls)
+        cls.__dispatches = tuple(mro[: mro.index(Event) + 1])
 
     @property
     @abc.abstractmethod
@@ -89,19 +85,9 @@ class Event(abc.ABC):
         """
 
     @classmethod
-    def subclasses(cls) -> typing.AbstractSet[typing.Type[Event]]:
-        """Set of the event class' subclasses."""
-        return cls.__subclasses
-
-    @classmethod
-    def on_new_subclass(cls, listener: typing.Callable[[typing.Type[Event]], None], /) -> None:
-        """Add a callback to be called whenever a new `Event` subclass is made."""
-        cls.__on_new_subclass.add(listener)
-
-    @classmethod
-    def remove_on_new_subclass(cls, listener: typing.Callable[[typing.Type[Event]], None], /) -> None:
-        """Remove a new subclass callback."""
-        cls.__on_new_subclass.remove(listener)
+    def dispatches(cls) -> typing.Sequence[typing.Type[Event]]:
+        """Sequence of the event classes this class is dispatched as."""
+        return cls.__dispatches
 
 
 def get_required_intents_for(event_type: typing.Type[Event]) -> typing.Collection[intents.Intents]:
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -61,7 +61,7 @@ class Event(abc.ABC):
 
     def __init_subclass__(cls) -> None:
         super().__init_subclass__()
-        # hasattr doesn't work with protected variables in this case so we use a try except.
+        # hasattr doesn't work with private variables in this case so we use a try except.
         # We need to set Event's __dispatches when the first subclass is made as Event itself cannot
         # be included in a tuple literal on itself due to not existing yet.
         try:
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -62,8 +62,8 @@ class Event(abc.ABC):
     def __init_subclass__(cls) -> None:
         super().__init_subclass__()
         # hasattr doesn't work with protected variables in this case so we use a try except.
-        # we need to set Event's __dispatches when the first subclass is made as Event itself cannot
-        # be included in a set literal on itself due to not existing yet.
+        # We need to set Event's __dispatches when the first subclass is made as Event itself cannot
+        # be included in a tuple literal on itself due to not existing yet.
         try:
             Event.__dispatches
         except AttributeError:
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -70,7 +70,8 @@ class Event(abc.ABC):
             Event.__dispatches = (Event,)
 
         mro = cls.mro()
-        cls.__dispatches = tuple(mro[: mro.index(Event) + 1])
+        # Non-event classes should be ignored.
+        cls.__dispatches = tuple(cls for cls in mro if issubclass(cls, Event))
 
     @property
     @abc.abstractmethod
@@ -85,7 +86,7 @@ class Event(abc.ABC):
 
     @classmethod
     def dispatches(cls) -> typing.Sequence[typing.Type[Event]]:
-        """Sequence of the event classes this class is dispatched as."""
+        """Sequence of the event classes this event is dispatched as."""
         return cls.__dispatches
 
 
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -62,7 +62,7 @@ class Event(abc.ABC):
     def __init_subclass__(cls) -> None:
         super().__init_subclass__()
         # hasattr doesn't work with private variables in this case so we use a try except.
-        # We need to set Event's __dispatches when the first subclass is made as Event itself cannot
+        # We need to set Event's __dispatches when the first subclass is made as Event cannot
         # be included in a tuple literal on itself due to not existing yet.
         try:
             Event.__dispatches
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -70,6 +70,7 @@ class Event(abc.ABC):
             Event.__dispatches = (Event,)
 
         mro = cls.mro()
+        # We don't have to explicitly include Event here as issubclass(Event, Event) returns True.
         # Non-event classes should be ignored.
         cls.__dispatches = tuple(cls for cls in mro if issubclass(cls, Event))
 
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -57,6 +57,22 @@ class Event(abc.ABC):
 
     __slots__: typing.Sequence[str] = ()
 
+    __subclasses: typing.ClassVar[typing.Set[typing.Type[Event]]]
+
+    def __init_subclass__(cls) -> None:
+        super().__init_subclass__()
+        # For convenience's sake we include a class in the set of it's subclasses.
+        # This also matches how issubclass(Foo, Foo) returns True
+        cls.__subclasses = {cls}
+
+        for parent_cls in cls.mro():
+            if parent_cls is not cls and issubclass(parent_cls, Event):
+                try:
+                    parent_cls.__subclasses.add(cls)
+                except AttributeError:
+                    # This should only ever apply to the base Event class
+                    parent_cls.__subclasses = {parent_cls, cls}
+
     @property
     @abc.abstractmethod
     def app(self) -> traits.RESTAware:
@@ -67,6 +83,10 @@ class Event(abc.ABC):
         hikari.traits.RESTAware
             The REST-aware app trait.
         """
+
+    @classmethod
+    def subclasses(cls) -> typing.Set[typing.Type[Event]]:
+        return cls.__subclass
 
 
 def get_required_intents_for(event_type: typing.Type[Event]) -> typing.Collection[intents.Intents]:
--- a/hikari/events/base_events.py
+++ b/hikari/events/base_events.py
@@ -70,7 +70,6 @@ class Event(abc.ABC):
             Event.__dispatches = (Event,)
 
         mro = cls.mro()
-
         cls.__dispatches = tuple(mro[: mro.index(Event) + 1])
 
     @property
--- a/hikari/impl/cache.py
+++ b/hikari/impl/cache.py
@@ -770,7 +770,8 @@ class CacheImpl(cache.MutableCache):
         return copy.copy(self._me)
 
     def set_me(self, user: users.OwnUser, /) -> None:
-        self._me = copy.copy(user)
+        if self._is_cache_enabled_for(config.CacheComponents.ME):
+            self._me = copy.copy(user)
 
     def update_me(
         self, user: users.OwnUser, /
--- a/hikari/impl/cache.py
+++ b/hikari/impl/cache.py
@@ -771,15 +771,18 @@ class CacheImpl(cache.MutableCache):
 
     def set_me(self, user: users.OwnUser, /) -> None:
         if self._is_cache_enabled_for(config.CacheComponents.ME):
+            _LOGGER.debug("setting my user to %s", user)
             self._me = copy.copy(user)
 
     def update_me(
         self, user: users.OwnUser, /
     ) -> typing.Tuple[typing.Optional[users.OwnUser], typing.Optional[users.OwnUser]]:
-        _LOGGER.debug("setting my user to %s", user)
+        if not self._is_cache_enabled_for(config.CacheComponents.ME):
+            return None, None
+
         cached_user = self.get_me()
         self.set_me(user)
-        return cached_user, self._me
+        return cached_user, self.get_me()
 
     def _build_member(
         self,
--- a/hikari/impl/event_manager.py
+++ b/hikari/impl/event_manager.py
@@ -52,6 +52,7 @@ from hikari.internal import time
 from hikari.internal import ux
 
 if typing.TYPE_CHECKING:
+    from hikari import emojis as emojis_
     from hikari import guilds
     from hikari import invites
     from hikari import voices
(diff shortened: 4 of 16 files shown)


hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,130 @@
...
+
+    def decorator(method: UnboundMethodT[EventManagerBaseT], /) -> UnboundMethodT[EventManagerBaseT]:
+        setattr(method, _CACHE_RESOURCE_ATTRIBUTE, cache_resource)
+        setattr(method, _EVENT_TYPES_ATTRIBUTE, event_types)
+        return method
+
+    return decorator
+
+
+@attr.frozen()
```
davfsa, line 106:
```suggestion
@attr.define()
```

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,130 @@
...
         for name, member in inspect.getmembers(self):
             if name.startswith("on_"):
-                self._consumers[name[3:]] = member
+                member = typing.cast("MethodT", member)
+                cache_resource = getattr(member, _CACHE_RESOURCE_ATTRIBUTE, undefined.UNDEFINED)
+                event_types = getattr(member, _EVENT_TYPES_ATTRIBUTE, undefined.UNDEFINED)
+                cache_resource = typing.cast("undefined.UndefinedOr[config.CacheComponents]", cache_resource)
+                event_types = typing.cast(
+                    "undefined.UndefinedOr[typing.Sequence[typing.Type[base_events.Event]]]", event_types
+                )
```
davfsa, line 137: are these `typing.cast` necessary or can we just type them as being that?
```suggestion
                member = typing.cast("MethodT", member)
                cache_resource: undefined.UndefinedOr[config.CacheComponents] = getattr(member, _CACHE_RESOURCE_ATTRIBUTE, undefined.UNDEFINED)
                event_types: undefined.UndefinedOr[typing.Sequence[typing.Type[base_events.Event]]] = getattr(member, _EVENT_TYPES_ATTRIBUTE, undefined.UNDEFINED)
```

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,130 @@
...
+        setattr(method, _EVENT_TYPES_ATTRIBUTE, event_types)
+        return method
+
+    return decorator
+
+
+@attr.frozen()
+class _Consumer:
+    callback: ConsumerT
+    cache: undefined.UndefinedOr[config.CacheComponents]
```
davfsa, line 109:
```suggestion
    required_cache_components: undefined.UndefinedOr[config.CacheComponents]
```

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,130 @@
...
     is the raw event name being dispatched in lower-case.
     """
 
-    __slots__: typing.Sequence[str] = ("_app", "_listeners", "_consumers", "_waiters")
+    __slots__: typing.Sequence[str] = ("_app", "_dispatches_for_cache", "_listeners", "_consumers", "_waiters")
 
     def __init__(self, app: traits.BotAware) -> None:
         self._app = app
-        self._consumers: typing.Dict[str, ConsumerT] = {}
+        self._dispatches_for_cache: typing.Dict[_Consumer, bool] = {}
```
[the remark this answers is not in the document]
FasterSpeeding: it's not cleared on every listener/waiter removal/addition though, only when the last listener/waiter for a type is removed or the first listener/waiter for a type is added

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,130 @@
...
+
+    def _enabled_for_event(self, event_type: typing.Type[base_events.Event], /) -> bool:
+        for cls in event_type.dispatches():
+            if cls in self._listeners or cls in self._waiters:
+                return True
+
+        return False
+
+    # This returns int rather than bool to avoid unnecessary bool casts
+    def _enabled_for_consumer(self, consumer: _Consumer) -> int:
```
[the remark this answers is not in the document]
FasterSpeeding:
```py
return (
    consumer.cache is undefined.UNDEFINED
    or consumer.cache != config.CacheComponents.NONE
    and consumer.cache & self._app.cache.settings.components
)
```
resolves to `Union[builtins.bool, hikari.config.CacheComponents]`

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,130 @@
...
+        # If undefined then we can only safely assume that this does link to registered listeners.
+        if consumer.event_types is undefined.UNDEFINED:
+            return True
+
+        if (cached_value := self._dispatches_for_cache.get(consumer, ...)) is True:
+            return True
+
+        if cached_value is ...:
```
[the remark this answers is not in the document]
FasterSpeeding: if it's not cached (`...`) then we want to calculate the actual value, otherwise the cached value is `False` and we want to skip straight ahead to the cache components check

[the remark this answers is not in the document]
davfsa: Cached value can only ever be a bool tho :thinking:

hikari/impl/event_manager_base.py
FasterSpeeding: Frozen is used for it's hashability and only a couple dozen of these are built at start-up so there's no real performance concern about the usage

hikari/impl/event_manager_base.py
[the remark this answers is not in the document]
FasterSpeeding: you're missing the default in `self._dispatches_for_cache.get(consumer, ...)`

hikari/impl/event_manager_base.py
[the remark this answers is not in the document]
davfsa: > ```python
> return (
>     consumer.cache is undefined.UNDEFINED
>     or consumer.cache != config.CacheComponents.NONE
>     and consumer.cache & self._app.cache.settings.components
> )
> ```
> 
> resolves to `Union[builtins.bool, hikari.config.CacheComponents]`

Wouldnt that have to be
```py
return (
    consumer.cache is undefined.UNDEFINED
    or consumer.cache != config.CacheComponents.NONE
    and (consumer.cache & self._app.cache.settings.components) == consumer.cache
)
```
instead to later resolve to a `bool`?

hikari/impl/event_manager_base.py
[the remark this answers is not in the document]
davfsa: Ik, which is why I'm saying there is no need for it. It could just be `self._dispatches_for_cache.get(consumer)` and you use the default as `None`, no?

hikari/impl/event_manager_base.py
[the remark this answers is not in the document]
FasterSpeeding:
```py
(consumer.cache & self._app.cache.settings.components) == consumer.cache
```
For this filtering it should return true if any of the bits are enabled, not just if all of them are enabled.

hikari/impl/event_manager_base.py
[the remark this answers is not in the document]
FasterSpeeding: Either works, they're both standard singletons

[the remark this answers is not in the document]
davfsa: I just mean because it will make it easier to read

hikari/impl/event_manager_base.py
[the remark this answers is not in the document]
FasterSpeeding: > I do not see the benefit of using this versus always checking the cache components.

Also this isn't caching the result of the cache component check, this is caching the result of iterating over all the event classes in an event class' mro to see if any have a registered listener or waiter.

hikari/impl/cache.py
```diff
@@ -725,7 +725,8 @@ def get_me(self) -> typing.Optional[users.OwnUser]:
         return copy.copy(self._me)
 
     def set_me(self, user: users.OwnUser, /) -> None:
-        self._me = copy.copy(user)
+        if self._is_cache_enabled_for(config.CacheComponents.ME):
+            self._me = copy.copy(user)
```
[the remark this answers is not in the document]
FasterSpeeding: added

hikari/impl/event_manager_base.py
FasterSpeeding: Added

hikari/impl/event_manager_base.py
FasterSpeeding: renamed the field but not to "required..." since it's not quite a requirement

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,135 @@
...
+                    return True
+
+            self._enabled_consumers_cache[consumer] = False
+
+        # If cache_components is UNDEFINED then we have to fall back to assuming that the consumer might set state.
+        # If cache_components is NONE then it doesn't make set state calls.
+        return (
+            consumer.cache_components is undefined.UNDEFINED
+            or consumer.cache_components != config.CacheComponents.NONE
+            and consumer.cache_components & self._app.cache.settings.components
```
davfsa, line 175:
```suggestion
            and (consumer.cache_components & self._app.cache.settings.components) != 0
```

The fact that this returns `int` makes it really weird :confused:

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,135 @@
...
+            return True
+
+        if (cached_value := self._enabled_consumers_cache.get(consumer)) is True:
+            return True
+
+        if cached_value is None:
+            for event_type in consumer.event_types:
+                if event_type in self._listeners or event_type in self._waiters:
+                    self._enabled_consumers_cache[consumer] = True
+                    return True
```
davfsa, line 166: Wouldn't this cause problems if the `event_type` in the waiters is removed right after this check? We will be deserializing objects that we don't need to deserialize (which is what this pr is trying to prevent).

Maybe we will have to remove the consumer entry from `_enabled_consumers_cache` when deleted off waiters or listeners

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +56,135 @@
...
-        self._consumers: typing.Dict[str, ConsumerT] = {}
+        self._consumers: typing.Dict[str, _Consumer] = {}
+        self._enabled_consumers_cache: typing.Dict[_Consumer, bool] = {}
         self._listeners: ListenerMapT[base_events.Event] = {}
         self._waiters: WaiterMapT[base_events.Event] = {}
 
         for name, member in inspect.getmembers(self):
             if name.startswith("on_"):
-                self._consumers[name[3:]] = member
+                member = typing.cast(MethodT, member)
```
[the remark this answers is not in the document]
FasterSpeeding: you can't re-annotate a variable which has already been declared

hikari/impl/event_manager_base.py
FasterSpeeding: This is a single thread environment and consume_raw_event never yields to the event loop so that'd never happen

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +58,155 @@
...
     """
 
-    __slots__: typing.Sequence[str] = ("_app", "_listeners", "_consumers", "_waiters")
+    __slots__: typing.Sequence[str] = ("_app", "_consumers", "_enabled_consumers_cache", "_listeners", "_waiters")
 
     def __init__(self, app: traits.BotAware) -> None:
         self._app = app
-        self._consumers: typing.Dict[str, ConsumerT] = {}
+        self._consumers: typing.Dict[str, _Consumer] = {}
+        self._enabled_consumers_cache: typing.Dict[_Consumer, bool] = {}
```
davfsa, line 152: Could this cache be replaced with just `_Consumer` holding whether it is enabled or not? 

I just thought of this, but I dont know how well it will work.
Essentially, each consumer holds a value of how many listeners it has (0 for none) and whether the cache is enabled for it (this will be setup when the consumer is created)
Then, to check if we should dispatch or not, we check if the cache for it is enabled and if not, check the number of listeners it has.

FasterSpeeding: > Could this cache be replaced with just _Consumer holding whether it is enabled or not?

The thing with this idea is when registering listeners and waiters we don't easily know which consumer(s) which are relevant for the provided Event and to modify the cached value we'd have to somehow work out all the consumers an event class is relevant to which is possible but would add in a noticeable amount of complexity.

Chances are this'd mean we'd have to iterate over a resource each time listeners or waiters are added (likely just iterating over all the registered consumers and checking their `.event_types` to see if there's a match in them or another resource which stores something like `Dict[Set[Type[Event]], List[_Consumer]]`, really depends on which one ends up meaning less iteration and idk what the dynamics would be for that off the top of my head lol) but just thinking about this just gives me shivers.

hikari/api/cache.py
```diff
@@ -78,6 +79,17 @@ class Cache(abc.ABC):
 
     __slots__: typing.Sequence[str] = ()
 
+    @property
+    @abc.abstractmethod
+    def settings(self) -> config.CacheSettings:
```
FasterSpeeding, line 84: split into base config with only the flags and then an extended one with client specific data for specific resources

hikari/events/base_events.py
```diff
@@ -52,10 +52,28 @@
...
+        try:
+            Event.__dispatches
+        except AttributeError:
+            Event.__dispatches = (Event,)
+
+        mro = cls.mro()
+        # We don't have to explicitly include Event here as issubclass(Event, Event) returns True.
+        # Non-event classes should be ignored.
+        cls.__dispatches = tuple(cls for cls in mro if issubclass(cls, Event))
+
```
davfsa, line 76: I am starting to think this is not a good idea. Building the list when we need it (which is not a lot) seems to be much more efficient (memory) than storing it in all the event classes

hikari/impl/cache.py
```diff
@@ -725,15 +725,19 @@ def get_me(self) -> typing.Optional[users.OwnUser]:
         return copy.copy(self._me)
 
     def set_me(self, user: users.OwnUser, /) -> None:
-        self._me = copy.copy(user)
+        if self._is_cache_enabled_for(config.CacheComponents.ME):
+            _LOGGER.debug("setting my user to %s", user)
```
davfsa, line 729: This debug was in `update_me` because we call it on `on_ready`

```py
    @event_manager_base.filtered(shard_events.ShardReadyEvent, config.CacheComponents.ME)
    async def on_ready(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
        """See https://discord.com/developers/docs/topics/gateway#ready for more info."""
        # TODO: cache unavailable guilds on startup, I didn't bother for the time being.
        event = self._app.event_factory.deserialize_ready_event(shard, payload)

        if self._cache:
            self._cache.update_me(event.my_user)

        await self.dispatch(event)
```

hikari/impl/event_manager.py
```diff
@@ -113,82 +136,168 @@ async def on_channel_delete(self, shard: gateway_shard.GatewayShard, payload: da
...
         await self.dispatch(event)
 
+    @event_manager_base.filtered((channel_events.GuildPinsUpdateEvent, channel_events.DMPinsUpdateEvent))
     async def on_channel_pins_update(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
         """See https://discord.com/developers/docs/topics/gateway#channel-pins-update for more info."""
         # TODO: we need a method for this specifically
         await self.dispatch(self._app.event_factory.deserialize_channel_pins_update_event(shard, payload))
 
+    # Internal granularity is preferred for GUILD_CREATE over decorator based filtering due to its large cache scope.
     async def on_guild_create(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
```
davfsa, line 146: These could actually hold the `filtered` metadata with their correct settings. And just add a comment saying that the function will later decrease the scope of what is cached + deserialized based on the information

hikari/impl/event_manager.py
```diff
@@ -113,82 +136,168 @@ async def on_channel_delete(self, shard: gateway_shard.GatewayShard, payload: da
...
+                    self._cache.set_presence(presence)
+
+            if voice_states:
+                self._cache.clear_voice_states_for_guild(guild_id)
+                for voice_state in voice_states.values():
+                    self._cache.set_voice_state(voice_state)
+
+        recv_chunks = self._enabled_for_event(shard_events.MemberChunkEvent) or self._cache_enabled_for(
+            config.CacheComponents.MEMBERS
+        )
```
davfsa, line 222: What if we spin up the MemberChunkEvent listener when we receive the GuildAvailable?

hikari/impl/event_manager.py
```diff
@@ -113,82 +136,168 @@ async def on_channel_delete(self, shard: gateway_shard.GatewayShard, payload: da
...
-
-        if self._cache:
-            self._cache.update_guild(event.guild)
+        enabled_for_event = self._enabled_for_event(guild_events.GuildUpdateEvent)
+
+        if not enabled_for_event and self._cache:
+            _LOGGER.log(ux.TRACE, "Skipping on_guild_update raw dispatch due to lack of any registered listeners")
+            event: typing.Optional[guild_events.GuildUpdateEvent] = None
+            gd = self._app.entity_factory.deserialize_gateway_guild(payload)
+            emojis = gd.emojis() if self._cache_enabled_for(config.CacheComponents.GUILDS) else None
```
davfsa, line 251:
```suggestion
            emojis = gd.emojis() if self._cache_enabled_for(config.CacheComponents.EMOJIS) else None
```

hikari/impl/event_manager.py
```diff
@@ -113,82 +136,168 @@ async def on_channel_delete(self, shard: gateway_shard.GatewayShard, payload: da
...
 
-        await self.dispatch(event)
+            coroutine = shard.request_guild_members(guild_id, include_presences=bool(presences_declared), nonce=nonce)
+            asyncio.create_task(coroutine, name=f"{shard.id}:{guild_id} guild create members request")
 
+        if event:
+            await self.dispatch(event)
+
+    # Internal granularity is preferred for GUILD_UPDATE over decorator based filtering due to its large cache scope.
     async def on_guild_update(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
```
davfsa, line 243: Same as above for on_guild_create

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +58,161 @@
...
+
 def _default_predicate(_: event_manager.EventT_inv) -> bool:
     return True
 
 
+def filtered(
+    event_types: typing.Union[typing.Type[base_events.Event], typing.Sequence[typing.Type[base_events.Event]]],
+    cache_components: config.CacheComponents = config.CacheComponents.NONE,
+    /,
+) -> typing.Callable[[UnboundMethodT[EventManagerBaseT]], UnboundMethodT[EventManagerBaseT]]:
```
davfsa, line 97: Could this be changed to mark consumers instead?

hikari/impl/event_manager_base.py
```diff
@@ -50,46 +58,161 @@
...
+    return decorator
+
+
+@attr.frozen()
+class _Consumer:
+    callback: ConsumerT = attr.ib()
+    """The callback function for this consumer."""
+
+    cache_components: undefined.UndefinedOr[config.CacheComponents] = attr.ib()
+    """Bitfield of the cache components this consumer makes modifying calls to, if set."""
```
davfsa, line 134: What about storing if the cache is enabled or not in the `__init__`?

hikari/impl/event_manager_base.py
```diff
@@ -151,14 +275,14 @@ def get_listeners(
     ) -> typing.Collection[event_manager.CallbackT[event_manager.EventT_co]]:
         if polymorphic:
             listeners: typing.List[event_manager.CallbackT[event_manager.EventT_co]] = []
-            for subscribed_event_type, subscribed_listeners in self._listeners.items():
-                if issubclass(subscribed_event_type, event_type):
-                    listeners += subscribed_listeners
+            for cls in event_type.dispatches():
+                listeners.extend(self._listeners[cls])
+
```
davfsa, line 280: This will raise `KeyError` as not all the event_types will be in the listeners:
```suggestion
            for cls in event_type.dispatches():
                if (items := self._listeners.get(cls)):
                    listeners.extend(items)

```

hikari/impl/event_manager_base.py
```diff
@@ -151,14 +275,14 @@ def get_listeners(
...
             return listeners
+
         else:
-            items = self._listeners.get(event_type)
-            if items is not None:
-                return items[:]
+            if items := self._listeners.get(event_type):
+                return items.copy()
 
             return []
```
davfsa, line 287: This else statement is redundant

hikari/events/base_events.py
FasterSpeeding: > Building the list when we need it (which is not a lot)
this list is used for every event dispatch though 🤔

davfsa: Completely missed that, nvm :+1:

hikari/impl/cache.py
FasterSpeeding: Seems a bit weird to have that one specific Logger though when this could also be applied to most of the cache set methods

hikari/impl/event_manager_base.py
FasterSpeeding: For that _clear_enabled_ cache would have to iterate over all the registered consumers each time it's called

hikari/impl/event_manager.py
FasterSpeeding: Seems unnecessary to duplicate steps like that for every guild create and update event dispatch

hikari/impl/event_manager_base.py
FasterSpeeding: That doesn't really bring any benefit as a change (plus it seems a bit odd to enforce external filtering when not all event consumers need to be externally filtered) (and fwiw that'd be a breaking change for anybody who's overriding methods on this class through inheritance)

hikari/impl/event_manager_base.py
davfsa: > For that _clear_enabled_ cache would have to iterate over all the registered consumers each time it's called

Not exactly, it would just simplify this piece of code:
```py
        # If cache_components is NONE then it doesn't make any altering state calls.
        return (
            consumer.cache_components != config.CacheComponents.NONE
            and (consumer.cache_components & self._app.cache.settings.components) != 0
        )
```

hikari/impl/event_manager.py
FasterSpeeding: What would the benefit be to that

davfsa: I mean, what if the users register the event listener when they receive the `GuildAvailableEvent`. This will cause it to not be registered before doing this check and cause the event to never be dispatched

*Really should have worded that correctly, sorry for that*

hikari/impl/event_manager_base.py
FasterSpeeding: That code's already going to be changed multiple times in the future, already changed it locally a while back since the != NONE bit is redundant and it'll have to be changed in the future to check if self._cache before checking the cache settings

FasterSpeeding: Also are you sure the cache settings are always going to be static a cross implementations

davfsa: The way everything is setup right now, it should be. I also dont see any reason to change the cache on the fly

hikari/impl/event_manager.py
FasterSpeeding: Well subscribing an event listener wouldn't yield to the event loop and neither does anything in the consumer functions so that race condition would only come around if they're subscribing the listener from another thread and this already isnt documented as being thread safe so...

hikari/impl/event_manager_base.py
FasterSpeeding: > The way everything is setup right now, it should be. 

Isn't that just implementation detail of the standard implementation though

hikari/api/cache.py
FasterSpeeding: I'ma do this in another PR

hikari/impl/event_manager_base.py
```diff
@@ -233,37 +259,137 @@ def _assert_is_listener(parameters: typing.Iterator[inspect.Parameter], /) -> No
...
+
+    Other Parameters
+    ----------------
+    cache_components : hikari.config.CacheComponents
+        Bitfield of the cache components this event may make altering calls to.
+        This defaults to `hikari.config.CacheComponents.NONE`.
+    """
+    if isinstance(event_types, typing.Sequence):
+        # dict.fromkeys is used to remove any duplicate entries here
+        event_types = tuple(dict.fromkeys(itertools.chain.from_iterable(e.dispatches() for e in event_types)))
```
davfsa, line 287: Using a set instead of a tuple here can help with base events, as they dispatch a lot of other events

(Same applies for `.dispatches()`)

hikari/impl/event_manager_base.py
davfsa: Using the set that I mentioned above, it should be quite easy to avoid all the loops, as all lookups to the event types will be O(1).

I don't like the temporary cache approach as it might be rendered useless in a lot of occasions (like a lot of waiters being constantly added and removed, causing the cache to be cleared all the time). I would like to try out the approach of making the consumers hold the information and just fetch it from there when needed, if that is fine with you.

FasterSpeeding: > Using the set that I mentioned above, it should be quite easy to avoid all the loops, as all lookups to the event types will be O(1).

And how would you deal with the fact that when registering or unregistering a listener or waiter you'd have to iterate over every event it's event dispatches for to change the state for any relevant consumer which is dispatching that event which would look something like

```py
for event_type in event.dispatches():
    for consumer in self._event_to_consumer[event_type]:
        consumer.listeners += 1
```

or

```py
for event_type in event.dispatches():
    for consumer in self._consumers:
        if event_type in consumer.event_types:
            consumer.listeners += 1
```

when adding a listener/waiter and -= when removing a listener or waiter

hikari/impl/event_manager_base.py
FasterSpeeding: That'd erm use noticeable more memory and be slower to iterate over which is the main thing we do with it

hikari/impl/event_manager_base.py
FasterSpeeding: > I don't like the temporary cache approach as it might be rendered useless in a lot of occasions (like a lot of waiters being constantly added and removed, causing the cache to be cleared all the time)

So in the case where a bot is registering 1 and only 1 waiter for an event type between every single event dispatch and that waiter presumably gets unregistered when the next event is received?

FasterSpeeding:
@davfsa i might officially put this PR on ice for a bit cause there's other things we should prioritise over an optimisation

davfsa:
> @davfsa i might officially put this PR on ice for a bit cause there's other things we should prioritise over an optimisation

Sounds good to me

davfsa:
Merging this to a separate branch so that I can add my work on top of it

Issue #35, opened 2020-08-20
Channels methods.


davfsa:
Methods and properties should be added to channels.py classes.

These are just general ideas, and should be down to the discretion of the developer, specifically.

Also these should be fully documented and tested as usual.

- `TextChannel`
    - `fetch_history`
    - `last_message`
    - `typing` async context manager
    - `trigger_typing`
    - `send`
    - `fetch_message`
- `GuildChannel#fetch_parent`
- `GuildChannel#add_overwrite`
- `GuildChannel#remove_overwrite`
- `GuildChannel#edit_overwrite`
- `GuildChannel#edit`
- `GuildChannel#delete`
- `GuildChannel#fetch_guild`
- `GuildTextChannel#fetch_pins`

PR #637, opened 2021-06-01
Add channel methods

ashwinvin:
### Summary
This pr adds the methods specified #35

### Checklist

### Related issues
#35

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -493,6 +493,33 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             If an internal error occurs on Discord while handling the request.
         """
 
+    @typing.overload
+    @abc.abstractmethod
+    async def edit_permission_overwrites(
+        self,
+        channel: snowflakes.SnowflakeishOr[channels_.GuildChannel],
+        target: typing.Union[channels_.PermissionOverwrite, users.PartialUser, guilds.PartialRole],
+        *,
+        allow: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
+        deny: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
+        reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
+    ) -> None:
+        """Edit permissions for a target entity."""
+
+    @typing.overload
+    @abc.abstractmethod
+    async def edit_permission_overwrites(
+        self,
+        channel: snowflakes.SnowflakeishOr[channels_.GuildChannel],
+        target: snowflakes.Snowflakeish,
+        *,
+        target_type: typing.Union[channels_.PermissionOverwriteType, int],
+        allow: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
+        deny: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
+        reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
+    ) -> None:
+        """Edit permissions for a given entity ID and type."""
+
     @abc.abstractmethod
     async def edit_permission_overwrites(
         self,
@@ -501,7 +528,7 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
             snowflakes.Snowflakeish, users.PartialUser, guilds.PartialRole, channels_.PermissionOverwrite
         ],
         *,
-        target_type: undefined.UndefinedOr[typing.Union[channels_.PermissionOverwriteType, str]] = undefined.UNDEFINED,
+        target_type: undefined.UndefinedOr[typing.Union[channels_.PermissionOverwriteType, int]] = undefined.UNDEFINED,
         allow: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
         deny: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
         reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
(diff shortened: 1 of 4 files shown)


hikari/channels.py
```diff
@@ -732,6 +962,237 @@ def shard_id(self) -> typing.Optional[int]:
...
+            however, Discord implements other undocumented rules for
+            rate-limiting, such as limits per attribute. These cannot be
+            detected or handled normally by Hikari due to their undocumented
+            nature, and will trigger this exception if they occur.
+        hikari.errors.InternalServerError
+            If an internal error occurs on Discord while handling the request.
+        """
+        return await self.app.rest.fetch_guild(self.guild_id)
+
+    async def edit_overwrite(
```
ashwinvin, line 1003: I don't think we need to add `add_overwrite`, I think its better to document the behavior in `edit_overwrite`.

hikari/channels.py
```diff
@@ -346,7 +384,7 @@ class TextChannel(PartialChannel, abc.ABC):
     __slots__: typing.Sequence[str] = ()
 
     # TODO: add examples to this and the REST method this invokes.
-    def history(
+    def fetch_history(
```
davfsa, line 387: Why rename this? `channel.history()` looks better :)

FasterSpeeding: If we ever add a method which gets the messages cached for this channel the name "fetch_history" gives us a distinction would be nice to have under the naming scheme present in #582

hikari/api/rest.py
```diff
@@ -472,33 +472,6 @@ async def edit_voice_state(
...
 
-    @typing.overload
-    @abc.abstractmethod
-    async def edit_permission_overwrites(
-        self,
-        channel: snowflakes.SnowflakeishOr[channels_.GuildChannel],
-        target: typing.Union[channels_.PermissionOverwrite, users.PartialUser, guilds.PartialRole],
-        *,
-        allow: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
-        deny: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
```
FasterSpeeding, line 483: Why were these overloads removed?

hikari/api/rest.py
```diff
@@ -507,7 +480,7 @@ async def edit_permission_overwrites(
             snowflakes.Snowflakeish, users.PartialUser, guilds.PartialRole, channels_.PermissionOverwrite
         ],
         *,
-        target_type: undefined.UndefinedOr[channels_.PermissionOverwriteType] = undefined.UNDEFINED,
+        target_type: undefined.UndefinedOr[typing.Union[channels_.PermissionOverwriteType, str]] = undefined.UNDEFINED,
```
FasterSpeeding, line 483: The implementation's typing for this needs to be updated as well, along with the type in the docstring.

hikari/channels.py
davfsa: Hmmm, true. Will mark this as a breaking change then

hikari/api/rest.py
davfsa: This was edited to match the implementation type, which it didnt match. It is true that the typehint has to be updated as well

hikari/api/rest.py
davfsa: I told him to. They were causing an error where there were too many Unions and mypy would just error. This seems to have fixed it. This overloads were also not providing any extra information

FasterSpeeding: > This overloads were also not providing any extra information

The overloads were specifying the behaviour around inferring the target type from target where if a target is an int rather than an entity then target_type needs to be provided

hikari/api/rest.py
FasterSpeeding, line 483: also target_type should be int not str

FasterSpeeding: the too many unions behaviour seems to go away if that's correctly typed as [int | PermissionOverwriteType] and this does look a bit geh but if you assert that the expected different signatures are matched like the following then mypy doesn't complain 
```py
        if target_type is undefined.UNDEFINED:
            assert not isinstance(target, int)
            return await self.app.rest.edit_permission_overwrites(
                self.id, target, allow=allow, deny=deny, reason=reason
            )

        assert isinstance(target, int)
        return await self.app.rest.edit_permission_overwrites(
            self.id, target, target_type=target_type, allow=allow, deny=deny, reason=reason
        )
```

hikari/api/rest.py
FasterSpeeding: the too many unions behaviour seems to go away if that's correctly typed as [int | PermissionOverwriteType] (that likely gets rid of the union all together from MyPy's perspective as PermissionOverwriteType inherits from int) and this does look a bit geh but if you assert that the expected different signatures are matched like the following then mypy doesn't complain (would also want to copy the overloads over to the method on GuildChannel as well).

```py
        if target_type is undefined.UNDEFINED:
            assert not isinstance(
                target, int
            ), "Cannot determine the type of the target to update. Try specifying 'target_type' manually."
            return await self.app.rest.edit_permission_overwrites(
                self.id, target, allow=allow, deny=deny, reason=reason
            )

        # While the interface specifies that int should be provided in this case, implementation detail may allow a
        # role, user or overwrite object to be passed in this case.
        target = typing.cast(int, target)
        return await self.app.rest.edit_permission_overwrites(
            self.id, target, target_type=target_type, allow=allow, deny=deny, reason=reason
        )
```

(to give a quick POC)

FasterSpeeding (approved):

FasterSpeeding:
Needs rebasing but otherwise LGTM

ashwinvin:
Rebased everything.

FasterSpeeding:
The linting pipeline is failing on this

FasterSpeeding (approved):

Issue #626, opened 2021-05-28
Improve loop acquisition in `hikari/impl/bot.py`


davfsa:
Should this be in `hikari/impl/bot.py` too?

_Originally posted by @davfsa in https://github.com/hikari-py/hikari/pull/577#discussion_r641811548_

PR #641, opened 2021-06-03
Improve loop acquisition in BotApp.run

FasterSpeeding:
### Summary
Improve loop acquisition in BotApp.run

### Checklist

### Related issues
Fixes: #626

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -576,7 +576,15 @@ class BotApp(traits.BotAware):
         if shard_ids is not None and shard_count is None:
             raise TypeError("'shard_ids' must be passed with 'shard_count'")
 
-        loop = asyncio.get_event_loop()
+        # get_event_loop will error under oddly specific cases such as if set_event_loop has been called before even
+        # if it was just called with None or if it's called on a thread which isn't the main Thread.
+        try:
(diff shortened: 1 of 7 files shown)


hikari/impl/bot.py
```diff
@@ -576,7 +576,15 @@ def run(
...
 
-        loop = asyncio.get_event_loop()
+        # get_event_loop will error under oddly specific cases such as if set_event_loop has been called before even
+        # if it was just called with None or if it's called on a thread which isn't the main Thread.
+        try:
+            loop = asyncio.get_event_loop()
+
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
```
davfsa, line 586: Just to continue the discussion we were having before, wouldn't it be better to have run actually take control of the event loop?

The way asyncio.run manages it is making sure there is no running event loop and then creating a new one. They also set the running event loop to `None`.

*This could be done in the form of an option passed into `run` that toggles this (manage_loop?), and then we could deprecate and eventually remove `close_loop` in favor for this option*

FasterSpeeding: Maybe this could be moved to a separate issue?

davfsa (approved):

PR #644, opened 2021-06-04
Add mention properties to GuildChannel and PartialRole

FasterSpeeding:
### Summary
Add mention properties to GuildChannel and PartialRole

### Checklist

### DIFF
--- a/hikari/channels.py
+++ b/hikari/channels.py
@@ -721,6 +721,10 @@ class GuildChannel(PartialChannel):
     def mention(self) -> str:
         """Return a raw mention string for the guild channel.
 
+        !!! note
+            As of writing, GuildCategory channels are a special case
+            for this and mentions of them will not resolve as clickable.
+
         Returns
         -------
         builtins.str
--- a/hikari/channels.py
+++ b/hikari/channels.py
@@ -718,6 +718,17 @@ class GuildChannel(PartialChannel):
     """
 
     @property
+    def mention(self) -> str:
+        """Return a raw mention string for the guild channel.
+
+        Returns
+        -------
+        builtins.str
+            The mention string to use.
+        """
+        return f"<#{self.id}>"
+
+    @property
     def shard_id(self) -> typing.Optional[int]:
         """Return the shard ID for the shard.
 
--- a/hikari/guilds.py
+++ b/hikari/guilds.py
@@ -584,6 +584,17 @@ class PartialRole(snowflakes.Unique):
     name: str = attr.field(eq=False, hash=False, repr=True)
     """The role's name."""
 
+    @property
+    def mention(self) -> str:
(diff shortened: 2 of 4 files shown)


FasterSpeeding:
Could add a mention property to PartialChannel instead but DM channel mentions aren't clickable on desktop (they do still resolve to the channel's name though) but are on android fwiw

hikari/channels.py
```diff
@@ -717,6 +717,17 @@ class GuildChannel(PartialChannel):
     If no parent category is set for the channel, this will be `builtins.None`.
     """
 
+    @property
+    def mention(self) -> str:
```
davfsa, line 721: I have a feeling this might be implemented in the future, so we can add it to PartialChannel. It not being clickable on desktop is probably a bug, since it resolves

FasterSpeeding: Eh I'd lean in the side of assuming that it not being clickable on desktop is intended, they previously had the same behaviour in place for voice channels until around when public stages were added.
You also get the same behaviour for other channel types like guild categories for more obvious reasons and if it was a bug all of this behaviour would've been "solved" years ago since its really well known behaviour.

FasterSpeeding: Also out of curiosity why do you think it could be implemented in the future

davfsa: They had the same situation with voice channels back in the day. They got rendered with the icon but weren't clickable.

Also, random thought, having it in this class would also break when trying to mention a category?

FasterSpeeding: > They had the same situation with voice channels back in the day. They got rendered with the icon but weren't clickable.

Bit of a difference between a guild channel type they didn't allow mentioning until they brought in a new sub type of it which encouraged advertising it and a channel type which they don't allow mentioning cause its not a guild channel though.

Also my guess for why voice channel mentions weren't allowed isn't that it was a bug but rather that they'd previously thought voice channel mentions would be more annoying than they are helpful since they'd jump you straight into a possibly live voice chat and weren't bothered to bring in a way to differentiate between voice and text channel mentions since there'd be no real return on investment there.

FasterSpeeding: If they start allowing for DM channel mentions in the future this can simply be moved, nothing will break if they do that so this isn't exactly a case of preserving future compatibility. But suggesting behaviour which isn't possibly (dm channel mentions) might end up just leading to confusion and issues being made about it.

(Avoiding putting this method on GuildCategory would also be better if we want to avoid suggesting behaviour which doesn't exist but this'd lead to a lot of duplication and risks making mentioning guild channels using this method while following static typing more annoying than it's worth since they'd have to narrow down the scope of the variable further from GuildChannel)

davfsa (approved):

PR #646, opened 2021-06-04
Remove ability to set the `rest_url` to connect to

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -282,9 +282,6 @@ class RESTApp(traits.ExecutorAware):
     proxy_settings : typing.Optional[hikari.config.ProxySettings]
         Proxy settings to use. If `builtins.None` then no proxy configuration
         will be used.
-    url : typing.Optional[builtins.str]
-        The base URL for the API. You can generally leave this as being
-        `builtins.None` and the correct default API base URL will be generated.
 
     !!! note
         This event loop will be bound to a connector when the first call
@@ -296,7 +293,6 @@ class RESTApp(traits.ExecutorAware):
         "_http_settings",
         "_max_rate_limit",
         "_proxy_settings",
-        "_url",
     )
 
     def __init__(
@@ -306,13 +302,11 @@ class RESTApp(traits.ExecutorAware):
         http_settings: typing.Optional[config.HTTPSettings] = None,
         max_rate_limit: float = 300,
         proxy_settings: typing.Optional[config.ProxySettings] = None,
-        url: typing.Optional[str] = None,
     ) -> None:
         self._http_settings = config.HTTPSettings() if http_settings is None else http_settings
         self._proxy_settings = config.ProxySettings() if proxy_settings is None else proxy_settings
         self._executor = executor
         self._max_rate_limit = max_rate_limit
-        self._url = url
 
     @property
     def executor(self) -> typing.Optional[concurrent.futures.Executor]:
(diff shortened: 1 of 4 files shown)


FasterSpeeding:
Should this be removed? There has been one case in the past where the canary/ptb API have had different behaviour to production and if Discord ever manages to fuck up the production API while leaving canary/ptb live this might be useful 

(example scenario where an endpoint was only broken on canary/ptb https://github.com/discord/discord-api-docs/issues/2880)

davfsa:
> Should this be removed? There has been one case in the past where the canary/ptb API have had different behaviour to production and if Discord ever manages to fuck up the production API while leaving canary/ptb live this might be useful
> 
> (example scenario where an endpoint was only broken on canary/ptb [discord/discord-api-docs#2880](https://github.com/discord/discord-api-docs/issues/2880))

I don't really see any moment where you should not be using stable. You can always change `hikari.urls.BASE_URL` to point to `"https://canary.discord.com"` if you really wanted to use canary. This should actually be preferred over letting people pass any url to the classes, at it makes sure that you know what you are doing and not just pasting code from the internet.

Furthermore, Discords' rollout system seems to be canary -> ptb -> stable. Meaning that if they break stable, they would have also broken canary + ptb, rendering them useless too.

FasterSpeeding:
>  You can always change hikari.urls.BASE_URL to point to "https://canary.discord.com" if you really wanted to use canary. This should actually be preferred over letting people pass any url to the classes, at it makes sure that you know what you are doing and not just pasting code from the internet.

Are you more inferring that randomly copied code off the internet can't also change a lib level constant or that changing a constant which is defined as being final over using an argument shows that you know what you're doing cause neither really makes sense?

> Furthermore, Discords' rollout system seems to be canary -> ptb -> stable. Meaning that if they break stable, they would have also broken canary + ptb, rendering them useless too.

Bold assumption that there won't ever be bugs which cause issues with scaling that don't effect PTB or canary due to the fact that they just operate on a lower scale than stable.

davfsa:
> Are you more inferring that randomly copied code off the internet can't also change a lib level constant or that changing a constant which is defined as being final over using an argument shows that you better know what you're doing cause neither really makes sense?

Makes it more obvious at least.

> Bold assumption that there won't ever be bugs which cause issues with scaling that don't effect PTB or canary as much if they even pop up due to the fact that they just operate on a lower scale than stable.

True, but again, I rather ensure (even a tiny bit) that people know what they are doing, when not using stable.

FasterSpeeding:
> True, but again, I rather ensure (even a tiny bit) that people know what they are doing, when not using stable.

So the new approach here is to make stuff more annoying to work with just so most people don't bother to use certain stuff?

hikari/impl/rest.py
```diff
@@ -563,7 +549,7 @@ async def _request(
 
         headers.put(_X_AUDIT_LOG_REASON_HEADER, reason)
 
-        url = compiled_route.create_url(self._rest_url)
+        url = compiled_route.create_url(urls.REST_API_URL)
```
FasterSpeeding, line 552: constantly accessing the top level constant instead of saving it to the instance and accessing what's saved to the instance has different implications about how top level patching of said constant (which apparently we encourage now?) would effect the behaviour of already active clients

FasterSpeeding:
> Makes it more obvious at least.

also what makes it more obvious?

PR #647, opened 2021-06-04
Do not allow `token_type` being `None`

davfsa:
### Breaking
This is marked as breaking because it slightly breaks `master`, not any release

### Checklist


hikari/impl/rest.py
```diff
@@ -416,7 +415,7 @@ def __init__(
         max_rate_limit: float,
         proxy_settings: config.ProxySettings,
         token: typing.Union[str, None, rest_api.TokenStrategy],
-        token_type: typing.Union[applications.TokenType, str, None],
+        token_type: typing.Union[applications.TokenType, str],
```
FasterSpeeding, line 418: What should be passed when the token is a Token Strategy or Nonr

davfsa: Good point, will fix

davfsa:
#627 fixes this

PR #648, opened 2021-06-04
Document new 20031 error code

davfsa:
### Checklist

### DIFF
--- a/hikari/errors.py
+++ b/hikari/errors.py
@@ -227,10 +227,7 @@ class HTTPError(HikariError):
 
 @typing.final
 class RESTErrorCode(int, enums.Enum):
-    """Error codes provided as further info on errors returned by the REST API."""
-
-    UNKNOWN_ERROR = -1
-    """Error is not known."""
+    """Non-exhaustive enum of error codes provided as further info on errors returned by the REST API."""
 
     GENERAL_ERROR = 0
     """A general error, no further info provided."""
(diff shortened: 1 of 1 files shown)


FasterSpeeding:
The docstring for this class should mention that it's non-exhaustive

davfsa:
> 
> 
> The docstring for this class should mention that it's non-exhaustive

Wdym?

FasterSpeeding:
> > The docstring for this class should mention that it's non-exhaustive
> 
> Wdym?

The RESTErrorCode enum is an non-exhaustive list (it doesn't cover all the error codes just like the list in the docs) and should be documented as such.

Also I don't think UNKNOWN_ERROR is relevant anymore since initialisation should just return a literal int if it doesn't find an enum member for the value so that should probably be removed at some point.

davfsa:
> The RESTErrorCode enum is an non-exhaustive list (it doesn't cover all the error codes just like the list in the docs) and should be documented as such.

Yep, will add that.

> Also I don't think UNKNOWN_ERROR is relevant anymore since initialisation should just return a literal int if it doesn't find an enum member for the value so that should probably be removed at some point.

UNKNOWN_ERROR is an actual code, not just the default

FasterSpeeding:
> UNKNOWN_ERROR is an actual code, not just the default

Are you thinking of the gateway close event error code? Json error codes never go negative as far as documented behaviour goes so not exactly sure `UNKNOWN_ERROR = -1` is an actual error code for rest json responses

davfsa:
> Are you thinking of the gateway close event error code? Json error codes never go negative as far as documented behaviour goes so not exactly sure `UNKNOWN_ERROR = -1` is an actual error code for rest json responses (would've thought stuff would always default to 0 - General error - for json error responses)

Oh, true. Will remove

FasterSpeeding (approved):

PR #649, opened 2021-06-05
Replace json.* with data_binding.json_* where appropriate

davfsa:
### Checklist


FasterSpeeding:
Isn't this change already present in #577

davfsa:
Oh true, it is. Should we fix it with this one or the other one?

FasterSpeeding:
> Oh true, it is. Should we fix it with this one or the other one?

Either works ig

FasterSpeeding (approved):

PR #651, opened 2021-06-08
Add thread permissions + all perms method

FasterSpeeding:
### Summary
Add thread permissions + all perms method
* Add note for permissions which can be gated behind 2fa

### Checklist

### DIFF
--- a/hikari/permissions.py
+++ b/hikari/permissions.py
@@ -26,6 +26,8 @@ from __future__ import annotations
 
 __all__: typing.List[str] = ["Permissions"]
 
+import functools
+import operator
 import typing
 
 from hikari.internal import enums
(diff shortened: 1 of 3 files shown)


hikari/permissions.py
```diff
@@ -198,3 +254,29 @@ class Permissions(enums.Flag):
...
+    USE_PRIVATE_THREADS = 1 << 36
+    """Allows for creating and participating in private threads."""
+
+    @classmethod
+    def all_permissions(cls) -> Permissions:
+        """Get an instance of `Permissions` with all the known permissions.
+
+        Returns
+        -------
+        Permissions
```
davfsa, line 279:
```suggestion
        hikari.permissions.Permissions
```

FasterSpeeding: why does that need to be an absolute link when it's in the current module? Seems to generate fine on my machine 
image

davfsa (approved):

PR #656, opened 2021-06-11
Add __init__.pyi files to git to help mypy when downloading from git

davfsa:
### Summary
Allow `mypy --strict` to work correctly when installing from Github

### Checklist


pipelines/mypy.nox.py
```diff
@@ -46,3 +46,13 @@ def generate_stubs(session: nox.Session) -> None:
...
+
+    for path in STUBGEN_GENERATE:
+        # We add "i" because we turn ".py" into ".pyi"
+        stub_path = path + "i"
+
+        with open(stub_path, "r") as fp:
+            content = fp.read()
+
+        with open(stub_path, "w") as fp:
+            fp.write("# This file was automatically generated by `nox -s generate-stubs`. Do not update!\n\n" + content)
```
FasterSpeeding, line 58: Probably better to say something like "do not manually update/modify" here

davfsa: Updated :+1:

FasterSpeeding: You forgot to re-run the nox task

FasterSpeeding (approved):

pipelines/mypy.nox.py
davfsa: Yep :sweat_smile:

PR #657, opened 2021-06-12
Remove a couple type ignores + cleanup CacheMappingView

A5rocks:
### Summary
I really wanted to remove a few of those pesky type ignores, you know? But I may have added more than I removed... Anyways, this PR also contains cache changes as when I asked around about whether I could change some interfaces, I got directed to remove some things instead. 

### Checklist

### Related issues
None.


hikari/files.py
```diff
@@ -955,10 +955,10 @@ async def _wrap_iter(self) -> typing.AsyncGenerator[typing.Any, bytes]:
             for i in range(0, len(self.data), _MAGIC):
                 yield self.data[i : i + _MAGIC]  # noqa: E203
 
-        elif aio.is_async_iterator(self.data) or inspect.isasyncgen(self.data):
+        elif isinstance(self.data, typing.AsyncIterator) or isinstance(self.data, typing.AsyncGenerator):
```
A5rocks, line 956: I'm not *quite* sure these `isinstance` checks are legal here, but I'll check soon:tm: if they actually are.

hikari/internal/cache.py
```diff
@@ -707,7 +724,8 @@ def _copy_embed(embed: embeds_.Embed) -> embeds_.Embed:
         author=copy.copy(embed.author) if embed.author else None,
         provider=copy.copy(embed.provider) if embed.provider else None,
         footer=copy.copy(embed.footer) if embed.footer else None,
-        fields=list(map(copy.copy, embed.fields)),  # type: ignore[arg-type]
+        # this lambda is a stupid workaround for mypy not being able to map(copy.copy, [5, 6, 7])
+        fields=list(map(lambda field: copy.copy(field), embed.fields)),
```
FasterSpeeding, line 728: might as well just do a generative list literal rather than using map here

hikari/internal/cache.py
```diff
@@ -132,28 +145,32 @@ def __getitem__(self, key: KeyT) -> ValueT:
...
-        for key, value in self._data.items():
-            if self._predicate is None or self._predicate(value):
-                index += 1
+    @typing.overload
+    def get_item_at(self, index: slice, /) -> typing.Sequence[ValueT]:
+        ...
 
-            if current_index == index:
-                return self[key]
+    def get_item_at(self, index: typing.Union[slice, int]) -> typing.Union[ValueT, typing.Sequence[ValueT]]:
```
FasterSpeeding, line 161: probably want to update the typing for this method on hikari.api.cache.CacheView and the few other classes which inherit from it (plus you forgot the `, /`)

hikari/internal/cache.py
```diff
@@ -99,30 +99,43 @@ class CacheMappingView(cache.CacheView[KeyT, ValueT], typing.Generic[KeyT, Value
 
     __slots__: typing.Sequence[str] = ("_builder", "_data", "_predicate")
 
+    @typing.overload
+    def __init__(
+        self,
+        items: typing.Mapping[KeyT, ValueT],
+        *,
+        builder: typing.Literal[None] = None,
```
FasterSpeeding, line 107: you can just not include builder in this overload

hikari/internal/cache.py
```diff
@@ -99,30 +99,43 @@ class CacheMappingView(cache.CacheView[KeyT, ValueT], typing.Generic[KeyT, Value
 
     __slots__: typing.Sequence[str] = ("_builder", "_data", "_predicate")
```
davfsa, line 100:
```suggestion
    __slots__: typing.Sequence[str] = ("_builder", "_data")
```

hikari/files.py
```diff
@@ -969,15 +967,15 @@ async def _wrap_iter(self) -> typing.AsyncGenerator[typing.Any, bytes]:
             except StopIteration:
                 pass
 
-        elif inspect.isgenerator(self.data):
+        elif isinstance(self.data, typing.Generator):
```
davfsa, line 970: How do these checks differentiate from `inspect.isgenerator` speed wise?

hikari/api/cache.py
```diff
@@ -54,8 +54,18 @@ class CacheView(typing.Mapping[_KeyT, _ValueT], abc.ABC):
     __slots__: typing.Sequence[str] = ()
 
     @abc.abstractmethod
-    def get_item_at(self, index: int) -> _ValueT:
-        """Get an entry in the view at position `index`."""
+    @typing.overload
+    def get_item_at(self, index: int, /) -> _ValueT:
+        ...
+
+    @abc.abstractmethod
```
A5rocks, line 61: Does anyone happen to know whether this is valid? I mean, it passes type checking and the tests don't error but... it just kind of feels wrong.

davfsa: We have been doing it already. I don't know what the correct way of doing it is, but we have using this way already. https://github.com/hikari-py/hikari/blob/master/hikari/api/rest.py#L496-L587

hikari/internal/aio.py
```diff
@@ -36,6 +36,10 @@
 import inspect
 import typing
 
+if typing.TYPE_CHECKING:
+    # typing_extensions is a dependency of mypy, and pyright vendors it.
```
davfsa, line 40: `mypy` is not installed as a requirement as hikari, so this will break if people try to analyze typehints without mypy installed

hikari/config.py
```diff
@@ -333,7 +333,7 @@ def _(self, _: attr.Attribute[typing.Optional[int]], value: typing.Optional[int]
     _ssl: typing.Union[bool, ssl_.SSLContext] = attr.field(
         default=True,
         converter=_ssl_factory,
-        validator=attr.validators.instance_of(ssl_.SSLContext),  # type: ignore[assignment,arg-type]
+        validator=attr.validators.instance_of((ssl_.SSLContext, bool)),
     )
```
davfsa, line 337: This is not correct, as we want to make sure that `_ssl` cannot hold a bool. Unfortunately, due to how attr works, the typehint we provide will be the same as the one used for both the `__init__` and the value hold. This type ignore is the only way we can fix this

A5rocks: Are you sure? Because `ssl` has the docstring:
https://github.com/hikari-py/hikari/blob/15624037e97d989c2e22b1bb1e1f8f6afd28f8ac/hikari/config.py#L341-L376

Which mentions it can be assigned either a bool or an SSLContext,

davfsa: Yes, we allow passing both bool and SSLContext. A bool will later be converted into a SSLContext in `_ssl_factory`. The validator just asserts that what was returned was an SSLContext

A5rocks: Shouldn't `_ssl` be `ssl_.SSLContext` then? mypy at the very least takes the type of the converter (`_ssl_factory`) for the type in `__init__`, not the type hint: https://github.com/python/mypy/blob/master/mypy/plugins/attrs.py#L113

hikari/config.py
```diff
@@ -330,10 +330,10 @@ def _(self, _: attr.Attribute[typing.Optional[int]], value: typing.Optional[int]
...
             raise ValueError("http_settings.max_redirects must be None or a POSITIVE integer")
 
-    _ssl: typing.Union[bool, ssl_.SSLContext] = attr.field(
-        default=True,
+    _ssl: ssl_.SSLContext = attr.field(
+        factory=lambda: _ssl_factory(True),
         converter=_ssl_factory,
-        validator=attr.validators.instance_of(ssl_.SSLContext),  # type: ignore[assignment,arg-type]
+        validator=attr.validators.instance_of(ssl_.SSLContext),
     )
```
davfsa, line 337: Since we are here, could you merge this with the `ssl` property. There is no need for it anymore :)

A5rocks: done

hikari/internal/aio.py
A5rocks: I think this is probably fine -- if anyone complains, it can easily be fixed then. Adding a runtime dependency for something that is supposed to be static just doesn't sit right with me...

Additionally, this will only error if forward annotations suddenly die, or if someone runs `typing.get_type_hints` on the functions in question: both are not very likely imo, especially this is internal to the library.

A5rocks:
This PR needs:

> First-time contributors need a maintainer to approve running workflows

A5rocks:
Gonna close and re-open to give github actions the little nudge it needs I guess?

hikari/internal/aio.py
davfsa: @FasterSpeeding ?

FasterSpeeding: Seems fine to me

A5rocks:
@FasterSpeeding you may want to review the latest commit, as it changes some of the cache things.

davfsa (approved):
Would like to get Snabs approval before merging :)

hikari/internal/cache.py
```diff
@@ -120,30 +116,23 @@ def __init__(
...
-        items: typing.Mapping[KeyT, typing.Union[ValueT, DataT]],
+        items: typing.Union[typing.Mapping[KeyT, ValueT], typing.Mapping[KeyT, DataT]],
         *,
         builder: typing.Optional[typing.Callable[[DataT], ValueT]] = None,
     ) -> None:
         self._builder = builder
         self._data = items
 
-    @classmethod
-    def _copy(cls, value: ValueT) -> ValueT:
```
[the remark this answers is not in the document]
davfsa: Ah, my bad. Rocks, can you please revert?

hikari/config.py
```diff
@@ -261,7 +261,7 @@ def _(self, attrib: attr.Attribute[typing.Optional[float]], value: typing.Option
 
 
 @attr_extensions.with_copy
-@attr.define(kw_only=True, weakref_slot=False)
+@attr.define(kw_only=True, weakref_slot=False, on_setattr=[attr.setters.convert, attr.setters.validate])
```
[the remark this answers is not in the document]
A5rocks: Pretty sure attrs doesn't add `attr.setters.convert` by default, I'll check locally.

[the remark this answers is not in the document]
A5rocks: Confirmed:
```py
In [1]: import attr

In [2]: @attr.define()
   ...: class Example:
   ...:     b: int = attr.ib(converter=int)
   ...: 

In [3]: n = Example(5)

In [4]: n.b = '42'

In [5]: n
Out[5]: Example(b='42')

In [6]: @attr.define(on_setattr=[attr.setters.convert, attr.setters.validate])
   ...: class Example:
   ...:     b: int = attr.ib(converter=int)
   ...: 

In [7]: nn = Example(5)

In [8]: nn.b = '42'

In [9]: nn
Out[9]: Example(b=42)
```

hikari/internal/cache.py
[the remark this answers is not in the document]
A5rocks: Done.

hikari/config.py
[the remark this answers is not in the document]
FasterSpeeding: Welp that's not the behaviour I'm seeing, on the latest ver of attr i'm getting
```
ok 123
<class 'int'>
```
from the following code
```py
import attr

def test_convert(value: str) -> int:
    print("ok", value)
    return int(value)

@attr.define()
class Foo:
    foo = attr.field(converter=test_convert)


print(type(Foo("123").foo))
```

Also under current behaviour the _ssl field is using a converter without setting the on_setattr and validation or the property's assertion statement would be constantly erroring if the converter wasn't running and it was being left as a boolean value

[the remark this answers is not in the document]
A5rocks: Yeah, converters always work in the initializer. But, the docstring mentions `If assigned to <boolean value>`, which has the connotation of `settings.ssl = False` after making the settings. I may be misreading though!

hikari/internal/cache.py
```diff
@@ -135,28 +146,32 @@ def __getitem__(self, key: KeyT) -> ValueT:
...
+        if isinstance(result, typing.Sequence):
+            if self._builder is not None:
+                return [self._builder(item) for item in result]  # type: ignore[arg-type]
+            else:
+                return [self._copy(item) for item in result]  # type: ignore[arg-type]
+        else:
+            if self._builder is not None:
+                return self._builder(result)  # type: ignore[arg-type]
+            else:
+                return self._copy(result)  # type: ignore[arg-type]
```
davfsa, line 174: You reverted this one, but I think it was better the other way

hikari/config.py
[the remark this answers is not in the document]
FasterSpeeding: Well typing wise you can't assigned a bool to it anymore cause as far as MYPY's concerned since it's typed as as SSLContext it therefore can only be set as an SSLContext attribute wise although currently this is an unsettable property so that whole "assigned" thing was probably just ambiguous phrasing or outdated docs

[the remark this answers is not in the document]
FasterSpeeding: Also it's probably preferable that people aren't encouraged to set those fields like that since there's no guarantee it'll update the behaviour of whatever client it's attached to so those docs should prob be updated ~~and there might actually be merit in freezing these classes~~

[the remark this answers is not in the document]
FasterSpeeding: > probably just ambiguous phrasing or outdated docs
It's the latter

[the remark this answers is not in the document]
A5rocks: Done

hikari/internal/cache.py
A5rocks: Done

davfsa (dismissed):

hikari/internal/cache.py
```diff
@@ -83,80 +83,78 @@
...
 
-        if self._predicate is not None and not self._predicate(entry):
-            raise KeyError(key)
-
-        if self._builder is not None:
-            entry = self._builder(entry)  # type: ignore[arg-type]
-
+        if self._builder:
+            return self._builder(entry)  # type: ignore[arg-type]
         else:
```
FasterSpeeding, line 139: this else statement is redundant

davfsa (approved):

FasterSpeeding (approved):

PR #660, opened 2021-06-12
introduce REST filled DM Channel ID cache resource

FasterSpeeding:
### Summary
introduce REST filled DM Channel ID cache resource
* Remove caching DM channel objects on user objects in favour of new cache resource

### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1832,7 +1832,6 @@ class RESTClientImpl(rest_api.RESTClient):
         await self._request(route)
 
     async def create_dm_channel(self, user: snowflakes.SnowflakeishOr[users.PartialUser], /) -> channels_.DMChannel:
-        user = snowflakes.Snowflake(user)
         route = routes.POST_MY_CHANNELS.compile()
         body = data_binding.JSONObjectBuilder()
         body.put_snowflake("recipient_id", user)
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -798,8 +798,7 @@ class RESTClientImpl(rest_api.RESTClient):
         assert isinstance(response, dict)
         result = self._entity_factory.deserialize_channel(response)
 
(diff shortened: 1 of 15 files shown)


hikari/impl/rest.py
```diff
@@ -791,7 +796,13 @@ async def fetch_channel(
         route = routes.GET_CHANNEL.compile(channel=channel)
         response = await self._request(route)
         assert isinstance(response, dict)
-        return self._entity_factory.deserialize_channel(response)
+        result = self._entity_factory.deserialize_channel(response)
+
+        if result.type is channels_.ChannelType.DM and self._cache:
+            assert isinstance(result, channels_.DMChannel)
+            self._cache.set_dm_channel_id(result.recipient.id, result.id)
```
davfsa, line 803:
```suggestion
        if self._cache and isinstance(result, channels_.DMChannel):
            self._cache.set_dm_channel_id(result.recipient.id, result.id)
```

davfsa (commented):
Looks good, I'm just not so sure if a REST based cache is the way to go. This leaves a lot of open questions, like, should we also be caching channels, emojis, guilds, etc

FasterSpeeding:
> Looks good, I'm just not so sure if a REST based cache is the way to go. This leaves a lot of open questions, like, should we also be caching channels, emojis, guilds, etc

Except A you're already caching EST data just in a kinda bad way which leads to stale data being returned and doesn't have any formalised interface for it.

B how the cache functions around removing entries is implementation detail and cacheing based on REST would only work for resource implemention which aren't reliant on delete events to remove resources from it (which only covers messages, users and dm channel ids to the standard implementation )

davfsa:
Sounds good then. just the little comment I made with the review

hikari/impl/rest.py
FasterSpeeding: Wouldn't that be slower for optimised runs for what is the exact same result?

davfsa (approved):

Issue #625, opened 2021-05-28
Usage of asyncio.Event in __init__ is problematic


FasterSpeeding:
asyncio.Event's initialiser calls asyncio.get_event_loop which will error when called in a thread without an active event loop under cases such as if the event loop has been set to None, if it hasn't been set yet and this isn't the main thread and if the current loop was closed. We shouldn't be relying on the event loop being available or implicitly settable when initiating clients such as the RestClient (due to RESTBucketManager) and BotApp, both of which currently initialise an Event object in their initialisers.

This issue may also apply to other asyncio objects so this probably needs further investigation.

FasterSpeeding:
#629 happens to touch on this so I'll be assigning this to myself for the time being

FasterSpeeding:
#629 seems to have solved any cases where this was an issue for GatewayBot and RESTApp/RESTClientImpl with RESTBot not introducing any new cases so this is being closed

PR #661, opened 2021-06-13
Delay initialisation of BotApp._closing_event and use it as ._closed

FasterSpeeding:
### Summary
Delay initialisation of BotApp._closing_event and use it as ._closed
* This also solves the potential issue where BotApp._closed was never re-set to `False` once a bot was re-started

### Checklist

### Related issues
This contributes to #625

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -355,12 +355,6 @@ class BotApp(traits.BotAware):
         if not self._is_alive:
             raise errors.ComponentStateConflictError("bot is not running so it cannot be interacted with")
 
-    def _get_closing_event(self) -> asyncio.Event:
-        if not self._closing_event:
-            self._closing_event = asyncio.Event()
-
-        return self._closing_event
-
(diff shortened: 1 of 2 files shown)


hikari/impl/bot.py
```diff
@@ -357,17 +355,22 @@ def _check_if_alive(self) -> None:
         if not self._is_alive:
             raise errors.ComponentStateConflictError("bot is not running so it cannot be interacted with")
 
+    def _get_closing_event(self) -> asyncio.Event:
+        if not self._closing_event:
+            self._closing_event = asyncio.Event()
+
+        return self._closing_event
```
davfsa, line 362: This might lead to some unexpected cases where we unset `_closing_event` but this call will set it again

davfsa (changes requested):

hikari/impl/bot.py
FasterSpeeding: That function is no more 🦀

davfsa (approved):

PR #665, opened 2021-06-16
Add sending multiple embeds on create and edit message

FasterSpeeding:
### Summary
Add sending multiple embeds on create and edit message
* Fix bug where execute_webhook(embeds=None) on the standard REST client implementation would lead to erroneous behaviour
* Fix bug where Message.respond(reply=False) would try to respond to a message with the ID 0

### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1151,32 +1151,28 @@ class RESTClientImpl(rest_api.RESTClient):
         assert isinstance(response, dict)
         return self._entity_factory.deserialize_message(response)
 
-    async def create_message(
+    async def _create_message(
         self,
-        channel: snowflakes.SnowflakeishOr[channels_.TextChannel],
-        content: undefined.UndefinedOr[typing.Any] = undefined.UNDEFINED,
+        route: routes.CompiledRoute,
+        body: data_binding.JSONObjectBuilder,
+        query: typing.Optional[data_binding.StringMapBuilder] = None,
         *,
-        embed: undefined.UndefinedOr[embeds_.Embed] = undefined.UNDEFINED,
-        embeds: undefined.UndefinedOr[typing.Sequence[embeds_.Embed]] = undefined.UNDEFINED,
-        attachment: undefined.UndefinedOr[files.Resourceish] = undefined.UNDEFINED,
-        attachments: undefined.UndefinedOr[typing.Sequence[files.Resourceish]] = undefined.UNDEFINED,
-        tts: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
-        nonce: undefined.UndefinedOr[str] = undefined.UNDEFINED,
-        reply: undefined.UndefinedOr[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = undefined.UNDEFINED,
-        mentions_everyone: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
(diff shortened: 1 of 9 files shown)


hikari/impl/rest.py
```diff
@@ -1256,6 +1278,7 @@ async def edit_message(
         content: undefined.UndefinedOr[typing.Any] = undefined.UNDEFINED,
         *,
         embed: undefined.UndefinedNoneOr[embeds_.Embed] = undefined.UNDEFINED,
+        embeds: undefined.UndefinedNoneOr[typing.Sequence[embeds_.Embed]] = undefined.UNDEFINED,
```
davfsa, line 1281: Missing check for both `embed` and `embeds` as well for the check of passing a non-collection to `embeds`.

hikari/impl/rest.py
```diff
@@ -1211,10 +1224,19 @@ async def create_message(
...
-            body.put("embed", embed_payload)
             final_attachments.extend(embed_attachments)
+            serialized_embeds.append(embed_payload)
+        elif embeds is not undefined.UNDEFINED:
+            for embed in embeds:
+                embed_payload, embed_attachments = self._entity_factory.serialize_embed(embed)
+                final_attachments.extend(embed_attachments)
+                serialized_embeds.append(embed_payload)
+
+        body.put("embeds", serialized_embeds)
```
davfsa, line 1239:
```suggestion
        if serialized_embeds:
            body.put("embeds", serialized_embeds)
```

FasterSpeeding: What's the point of that check

hikari/impl/rest.py
davfsa: Should have both imo. If not, people will start to see unexpected behavior. Also, would it be worth to move this logic to a super `_create_message` and `_edit_message` as it is now the same for both webhooks and bots?

FasterSpeeding: The nonce field isn't present on execute webhook and this isn't present yet but you can only provide flags on execute webhook, not create message

davfsa: Ah ok 👍

davfsa: Actually, we could just always send nonce as `None` in the webhook message, so then we can merge the logic

FasterSpeeding: Idk what you mean by that but nonce is a user defined value, also can't merge the edit message endpoints since webhooks can't edit flags

davfsa: Ik,  you can then default those depending on the call.

```py
async def _create_message(route = UNDEFINED, nonce = UNDEFINED, flag = UNDEFINED, avatar = UNDEFINED, username = UNDEFINED, ***):
    ...
    
async def create_message(nonce):
    await _create_message(nonce)
    
async def execute_webhook(flag, avatar, username):
    await _create_message(flag, avatar, username)
```

FasterSpeeding: I've already merged them but didn't use that style cause there's a better way to handle it in this case

hikari/impl/rest.py
davfsa: Not sending unnecessary information when not needed

FasterSpeeding: I still don't see what the issue is here. It works and it's what we've been doing the whole time. This isn't meant to be a refactoor; this is all just copy-pasted code.

davfsa (approved):

PR #669, opened 2021-06-17
Do not send "embeds" in message create payloads when not specified

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1202,31 +1202,31 @@ class RESTClientImpl(rest_api.RESTClient):
             attachment = content
             content = undefined.UNDEFINED
 
-        body.put(
-            "allowed_mentions",
-            self._generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
-        )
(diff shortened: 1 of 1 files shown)


FasterSpeeding:
Are you going to handle all the other cases where empty values could be sent in create endpoints as well?

davfsa:
> Are you going to handle all the other cases where empty values could be sent in create endpoints as well?

If any of those are created by us if the user didn't specify anything, yes. But I haven't seen any like that.

FasterSpeeding:
> > Are you going to handle all the other cases where empty values could be sent in create endpoints as well?
> 
> If any of those are created by us if the user didn't specify anything, yes. But I haven't seen any like that.

Why does this matter anyway

davfsa:
> Why does this matter anyway

Not sending unnecessary information when we don't have to

FasterSpeeding:
> > Why does this matter anyway
> 
> Not sending unnecessary information when we don't have to
> 

Then shouldn't you also add in checks for when the user specified an empty value to avoid sending unnecessary information to other endpoints

FasterSpeeding (approved):

PR #674, opened 2021-06-19
Default content to None when empty string

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -1739,6 +1739,10 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
         if "stickers" in payload:
             stickers = [self._deserialize_sticker(sticker) for sticker in payload["stickers"]]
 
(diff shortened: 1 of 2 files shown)


hikari/impl/entity_factory.py
```diff
@@ -1739,21 +1739,25 @@ def deserialize_partial_message(  # noqa CFQ001 - Function too long
         if "stickers" in payload:
             stickers = [self._deserialize_sticker(sticker) for sticker in payload["stickers"]]
 
+        content = payload.get("content", undefined.UNDEFINED)
+        if content is not undefined.UNDEFINED:
+            content = content or None  # Default to None if content is an empty string
```
FasterSpeeding, line 1744: Shouldn't an extra test be added to account for this case and lower the chance of any regressions in behaviour passing by unnoticed?

davfsa: Added :+1:

FasterSpeeding (approved):

PR #679, opened 2021-06-23
Add ci-done job

davfsa:
### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,10 +165,22 @@ jobs:
           retention-days: 2
           if-no-files-found: error
 
-  deploy:
-    # Only run on releases and if everything passes
-    if: github.event_name == 'release'
+  # This job is only used to mark the CI as being done
+  #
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -165,10 +165,22 @@ jobs:
...
           if-no-files-found: error
 
+  # This job is only used to mark the CI as being done
+  #
+  # Allows us to add this as a required check in Github branch rules, as all the
+  # other jobs are subject to change
+  ci-done:
+    needs: [upload-coverage, linting, twemoji, pages]
+
+    runs-on: ubuntu-20.04
```
FasterSpeeding, line 175: is there anything lighter this could run on?

davfsa: Lightest that is available. https://github.com/actions/virtual-environments#available-environments

This will also be almost instant to spin up, as it is for all the other jobs. Difference is we dont setup anything (like python), which is what usually slows it down

FasterSpeeding (approved):

PR #681, opened 2021-06-26
Use Signature's eval_str argument in resolve_signature (3.10+ specific)

FasterSpeeding:
### Summary
Use Signature's eval_str argument in resolve_signature (3.10+ specific)

### Checklist

### DIFF
--- a/hikari/internal/reflect.py
+++ b/hikari/internal/reflect.py
@@ -28,6 +28,7 @@ __all__: typing.List[str] = ["resolve_signature"]
 
 import functools
 import inspect
+import sys
 import typing
 
 if typing.TYPE_CHECKING:
@@ -56,6 +57,9 @@ def resolve_signature(func: typing.Callable[..., typing.Any]) -> inspect.Signatu
         A `inspect.Signature` object with all forward reference annotations
         resolved.
     """
+    if sys.version_info >= (3, 10):
+        return inspect.signature(func, eval_str=True)
+
     signature = inspect.signature(func)
     resolved_typehints = typing.get_type_hints(func)
     params = []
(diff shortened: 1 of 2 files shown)


davfsa:
How does this differentiate from us evaluating them?

FasterSpeeding:
> How does this differentiate from us evaluating them?

Same behaviour but faster (both when forward references are enabled and when they aren't) and less reliant on our code.

hikari/internal/reflect.py
```diff
@@ -56,6 +57,9 @@ def resolve_signature(func: typing.Callable[..., typing.Any]) -> inspect.Signatu
         A `inspect.Signature` object with all forward reference annotations
         resolved.
     """
+    if sys.version_info >= (3, 10):
```
davfsa, line 60: Would it be bette to have this check outside?

tests/hikari/internal/test_reflect.py
```diff
@@ -19,12 +19,18 @@
...
 import typing
 
+import mock
+import pytest
+
 from hikari.internal import reflect
 
 
-class TestResolveSignature:
+@pytest.mark.skipif(sys.version_info >= (3, 10), reason="This strategy is specific to 3.10 > versions")
```
davfsa, line 32: Random point, but it just came to mind. Do we want to allow for specific version tests or do we want to allow all tests pass in any version of Python by mocking?

FasterSpeeding: Since the tests are run against every supported python ver in the pipelines this approach still gives full test coverage while being easier to implement and less likely to lead to stuff going unnoticed due to excessive mocking

hikari/internal/reflect.py
FasterSpeeding: Either approach works, this approach just in this case this is just kinda easier since avoids stuff like having to duplicate the docstring between the two definitions

davfsa (approved):

PR #686, opened 2021-06-30
Improve typing/strictness of Emoji classes

FasterSpeeding:
### Summary
Improve typing of Emoji classes
* Make CustomEmoji (and therefore also Emoji) more strict
* Flatten down reaction event models and guild WelcomeChannel + document special case behaviour on them directly.
* Remove Emoji.is_mentionable as it is no longer relevant.

This also removes cases where CustomEmoji.filename, CustomEmoji.mention, CustomEmoji.url_name and CustomEmoji.url would return potentially invalid data if a field was None.

* For CustomEmoji.mention there was a partner CustomEmoji.is_mentionable field which was probably intended to indicate whether CustomEmoji.mention would return data which is guaranteed to be valid or not but this behaviour wasn't documented.
* When CustomEmoji.url_name would return `"None:{emoji.id}"` while this technically works under current behaviour (as the emoji name is currently ignored for custom emojis in the reaction endpoint paths) this behaviour isn't guaranteed and may break down the line (see https://github.com/discord/discord-api-docs/issues/2653#issuecomment-786276283 ).

This also fixes a bug where a WelcomeChannel with emoji_id set but emoji_name left as null would be treated as if it doesn't have any emoji set during deserialization

### Checklist

### DIFF
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -1573,8 +1573,8 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         ----------------
         emoji_id : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishOr[hikari.emojis.CustomEmoji]]
             ID of the custom emoji to react with.
-            This should only be provided when the custom emoji's name is passed
-            for `emoji` as a `builtins.str`.
+            This should only be provided when a custom emoji's name is passed
+            for `emoji`.
 
         Raises
         ------
@@ -1628,8 +1628,8 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         ----------------
         emoji_id : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishOr[hikari.emojis.CustomEmoji]]
             ID of the custom emoji to remove your reaction for.
-            This should only be provided when the custom emoji's name is passed
-            for `emoji` as a `builtins.str`.
+            This should only be provided when a custom emoji's name is passed
+            for `emoji`.
 
         Raises
         ------
@@ -1680,8 +1680,8 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         ----------------
         emoji_id : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishOr[hikari.emojis.CustomEmoji]]
             ID of the custom emoji to remove all the reactions for.
-            This should only be provided when the custom emoji's name is passed
-            for `emoji` as a `builtins.str`.
+            This should only be provided when a custom emoji's name is passed
+            for `emoji`.
 
         Raises
         ------
@@ -1740,8 +1740,8 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         ----------------
         emoji_id : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishOr[hikari.emojis.CustomEmoji]]
             ID of the custom emoji to react with.
-            This should only be provided when the custom emoji's name is passed
-            for `emoji` as a `builtins.str`.
+            This should only be provided when a custom emoji's name is passed
+            for `emoji`.
 
         Raises
         ------
@@ -1837,8 +1837,8 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         ----------------
         emoji_id : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishOr[hikari.emojis.CustomEmoji]]
             ID of the custom emoji to get the reactions for.
-            This should only be provided when the custom emoji's name is passed
-            for `emoji` as a `builtins.str`.
+            This should only be provided when a custom emoji's name is passed
+            for `emoji`.
 
         Returns
         -------
--- a/hikari/api/rest.py
+++ b/hikari/api/rest.py
@@ -1552,7 +1552,8 @@ class RESTClient(traits.NetworkSettingsAware, abc.ABC):
         self,
         channel: snowflakes.SnowflakeishOr[channels_.TextableChannel],
         message: snowflakes.SnowflakeishOr[messages_.PartialMessage],
-        emoji: emojis.Emojiish,
+        emoji: typing.Union[str, emojis.Emoji],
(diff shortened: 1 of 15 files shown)


hikari/impl/entity_factory.py
```diff
@@ -1084,23 +1084,19 @@ def deserialize_welcome_screen(self, payload: data_binding.JSONObject) -> guild_
...
-            if emoji_name is not None:
-                if emoji_id is not None:
-                    emoji = emoji_models.CustomEmoji(
-                        id=snowflakes.Snowflake(emoji_id), name=emoji_name, is_animated=None
-                    )
-                else:
-                    emoji = emoji_models.UnicodeEmoji(emoji_name)
+            emoji_name: typing.Union[None, emoji_models.UnicodeEmoji, str]
+            if (emoji_name := channel_payload["emoji_name"]) and not emoji_id:
+                emoji_name = emoji_models.UnicodeEmoji(emoji_name)
```
davfsa, line 1092: I dont really understand why this logic was changed. This change looks like it should affect (amongst more things) how we serialize the WelcomeChannel not deserialize it

FasterSpeeding: This is documented on the class definition; emoji_name isn't always provided when a custom emoji is set for a welcome channel (only emoji_id is guaranteed)

davfsa: For my tests, emoji_name seems to always we provided, with emoji_id being optional:

- `emoji_name` and `emoji_id`: CustomEmoji
- `emoji_name` and not `emoji_id`: UnicodeEmoji
- not `emoji_name` and not `emoji_id`: None

FasterSpeeding: it isn't

davfsa: image


I am unable to reproduce what you are saying, is there something I am missing?

FasterSpeeding: > it's returned as null for custom emojis if it wasn't specified in the patch request

If emoji_name is not specified for a channel where you did specify emoji_id when you PATCHed the resource, emoji_name will be returned as null for that channel

davfsa: Oh true:

image


But we already handle this with the current approach, as we allow CustomEmojis to not have a name, due to how weirdly Discord implement this.

FasterSpeeding: > EDIT: I just noticed that this should really have a builder behind it

Why? this just matches the pattern seen with stuff like PermissionOverwrites, embeds (but without the fluent interface since these objects still aren't considered mutable), and attachments since it isn't nested at all

FasterSpeeding: > But we already handle this with the current approach, as we allow CustomEmojis to not have a name, due to how weirdly Discord implement this.

No we don't? we just ignore the emoji all together if it doesn't have a name set here

davfsa: Well then wouldn't the better approach be to just document `CustomEmoji.name` with the warning you added and just fix it to parse it based on the emoji_id, not the name (like you did here), not split it into two different fields?

FasterSpeeding: > Well then wouldn't the better approach be to just document CustomEmoji.name with the warning you added and just fix it to parse it based on the emoji_id, not the name (like you did here), not split it into two different fields everywhere?

Except that doesn't deal with the fact that the interface of CustomEmoji is kinda useless without the use of assertions everywhere it's used since it only guarantees ID as to fit in with two specific edge cases (welcome channel and reaction events) (this is just kinda bad typing, don't exactly want to have to rely on the source of a model to know if it'll work in a specific case cause now any function which takes it will have to document the cases where it'll raise and when using it after getting it from the source you'll have to use assertions based on context which might not even be guaranteed to stay the same), also wouldn't solve the other issues about methods on it returning potentially invalid data when fields are None and also means that CustomEmoji will now have a specific case where CustomEmoji.name.

FasterSpeeding: That bug isn't the focus of this PR, that's just something which was picked up since it fell within scope

hikari/api/rest.py
```diff
@@ -1565,8 +1566,15 @@ async def add_reaction(
...
-            The emoji to react to the message with.
+        emoji : typing.Union[builtins.str, hikari.emojis.Emoji]
+            Object or name of the emoji to react with.
+
+        Other Parameters
+        ----------------
+        emoji_id : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishOr[hikari.emojis.CustomEmoji]]
+            ID of the custom emoji to react with.
+            This should only be provided when the custom emoji's name is passed
+            for `emoji` as a `builtins.str`.
```
davfsa, line 1577:
```suggestion
            This should only be provided when `emoji` is the name of a custom emoji.
```

hikari/guilds.py
```diff
@@ -1195,8 +1195,18 @@ class WelcomeChannel:
...
-    emoji: typing.Optional[emojis_.Emoji] = attr.field(default=None, kw_only=True, hash=False, repr=True)
-    """The emoji shown in the welcome screen channel if set else `builtins.None`."""
+    emoji_name: typing.Union[str, emojis_.UnicodeEmoji, None] = attr.field(
+        default=None, kw_only=True, hash=False, repr=True
+    )
+    """The emoji shown in the welcome screen channel if set to a unicode emoji.
+
+    !!! warning
+        While this may also be provided for custom emojis, it isn't guaranteed and
+        may also sometimes even be wrong for custom emojis.
```
davfsa, line 1205:
```suggestion
        may also sometimes even be wrong.
```

davfsa (approved):

davfsa (approved):

hikari/emojis.py
```diff
@@ -131,33 +124,27 @@ class UnicodeEmoji(Emoji):
         removed in a future release after a deprecation period.
     """
 
-    name: str = attr.field(repr=True, hash=True)
-    """The code points that form the emoji."""
-
-    def __str__(self) -> str:
-        return self.name
+    __slots__: typing.Sequence[str] = []
```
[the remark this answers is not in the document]
FasterSpeeding: Fixed that

PR #688, opened 2021-07-01
Split Webhook into type specific definitions

### LINKED ISSUE #608: Split up webhook into 3 different models
### Summary
This should bring in support for the new webhook type in https://github.com/discord/discord-api-docs/pull/2926 along with splitting the webhook class into 3 different types in a similar fashion to how channels are handled. In this new structure only IncomingWebhooks should be executable (only incoming webhooks and possibly following webhooks should be mutable rest wise).

This will likely be blocked until #577 is merged due to the changes that already makes to webhooks. 

### Problem
Interaction webhooks aren't supported and support should be added without muddying up the semantics of any existing class. 

### Ideal implementation
See summary

### Checklist

FasterSpeeding:
### Summary
Split Webhook into type specific definitions

### Checklist

### Related issues
Closes #608

### DIFF
--- a/hikari/webhooks.py
+++ b/hikari/webhooks.py
@@ -24,7 +24,14 @@
 
 from __future__ import annotations
 
-__all__: typing.List[str] = ["ExecutableWebhook", "WebhookType", "Webhook"]
+__all__: typing.List[str] = [
+    "ApplicationWebhook",
+    "ChannelFollowerWebhook",
+    "ExecutableWebhook",
+    "PartialWebhook",
+    "WebhookType",
+    "IncomingWebhook",
+]
 
 import abc
 import typing
@@ -58,6 +65,9 @@ class WebhookType(int, enums.Enum):
 
     CHANNEL_FOLLOWER = 2
     """Channel Follower webhook."""
+
+    APPLICATION = 3
+    """Application webhook (from the interactions flow)."""
 
 
 class ExecutableWebhook(abc.ABC):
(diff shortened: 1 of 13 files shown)


hikari/webhooks.py
```diff
@@ -761,71 +798,164 @@ async def fetch_self(self, *, use_token: undefined.UndefinedOr[bool] = undefined
...
 
-        return routes.CDN_USER_AVATAR.compile_to_file(
-            urls.CDN_URL,
-            user_id=self.id,
-            hash=self.avatar_hash,
-            size=size,
-            file_format=ext,
-        )
+
+@attr.define(hash=True, kw_only=True, weakref_slot=False)
```
davfsa, line 953:
```suggestion
@attr.define(hash=True, kw_only=True, weakref_slot=False, auto_attribs=False)
```

hikari/webhooks.py
```diff
@@ -59,6 +66,9 @@ class WebhookType(int, enums.Enum):
     CHANNEL_FOLLOWER = 2
     """Channel Follower webhook."""
 
+    APPLICATION = 3
+    """Application webhook (from the interactions flow)."""
```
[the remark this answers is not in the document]
FasterSpeeding: That still is and always has been an incoming webhook

hikari/webhooks.py
FasterSpeeding: Don't see why that'd need to be disabled, doesn't really change anything here

davfsa: Stops attr from adding attributes to it, doesnt hurt to add

FasterSpeeding: Gonna keep redefining the attr field since this ensures field.type is correct for this field

davfsa (approved):

Issue #623, opened 2021-05-27
Implement threads


davfsa:
https://discord.com/developers/docs/topics/threads
https://discord.com/developers/docs/change-log#api-v9

This should be done with the full switch to v9 when it is out

davfsa:
This can now be implemented if anybody would like to:

![image](https://user-images.githubusercontent.com/29100934/123813753-736c5a00-d8f5-11eb-84b1-2870b86c54a0.png)

FasterSpeeding:
Btw I do have code for this locally

PR #691, opened 2021-07-02
Add TextableGuildChannel mixin class + improve typing around guild text channel operations

FasterSpeeding:
### Summary
* Rename TextChannel to TextableChannel + add TextableGuildChannel
  This fixes some typing issues where GuildTextChannel was being asserted and typed in a few scenarios where GuildNewsChannel and in the future thread channels would also be expected while also trying to lay out a better structure for handling guild channels which handle messages in-order to lower the chance of this kind of issue recurring
* Add WebhookChannelT union and WebhookChannelTypes to channels.py 
  This is used in the type hint for create_webhook, fetch_channel_webhooks, and edit_webhook in place of TextableChannel in-order to account for the fact that webhooks cannot be directly attached to thread channels.
* Improve typing for edit voice state rest methods to use GuildVoiceChannel rather than PartialChannel

### Checklist

### Related issues
This is a part of #623

### DIFF
--- a/hikari/channels.py
+++ b/hikari/channels.py
@@ -31,7 +31,8 @@ __all__: typing.List[str] = [
     "PermissionOverwrite",
     "PermissionOverwriteType",
     "PartialChannel",
-    "TextChannel",
+    "TextableChannel",
+    "TextableGuildChannel",
     "PrivateChannel",
     "DMChannel",
     "GroupDMChannel",
@@ -44,7 +45,6 @@ __all__: typing.List[str] = [
     "GuildStageChannel",
 ]
 
-import abc
 import typing
 
 import attr
@@ -380,8 +380,8 @@ class PartialChannel(snowflakes.Unique):
         return await self.app.rest.delete_channel(self.id)
 
 
-class TextChannel(PartialChannel, abc.ABC):
-    """A channel that can have text messages in it."""
+class TextableChannel(PartialChannel):
+    """Mixin class for a channel which can have text messages in it."""
 
     # This is a mixin, do not add slotted fields.
     __slots__: typing.Sequence[str] = ()
@@ -825,7 +825,7 @@ class PrivateChannel(PartialChannel):
 
 
 @attr.define(hash=True, kw_only=True, weakref_slot=False)
-class DMChannel(PrivateChannel, TextChannel):
+class DMChannel(PrivateChannel, TextableChannel):
     """Represents a direct message text channel that is between you and another user."""
 
     recipient: users.User = attr.field(eq=False, hash=False, repr=False)
@@ -845,7 +845,7 @@ class GroupDMChannel(PrivateChannel):
     """Represents a group direct message channel.
 
     !!! note
-        This doesn't have the methods found on `TextChannel` as bots cannot
+        This doesn't have the methods found on `TextableChannel` as bots cannot
         interact with a group DM that they own by sending or seeing messages in
         it.
     """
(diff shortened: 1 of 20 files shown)


hikari/channels.py
```diff
@@ -1339,3 +1348,22 @@ class GuildStageChannel(GuildChannel):
...
+"""
+
+WebhookChannelTypes = (GuildTextChannel, GuildNewsChannel)
+"""Tuple of the channel types which are valid for `WebhookChannelT`.
+
+This includes:
+
+* `GuildTextChannel`
+* `GuildNewsChannel`
+"""
```
davfsa, line 1369: Are these really needed?

FasterSpeeding: Well the whole WebhookChannelT specific typing isn't necessary so WebhookChannelT doesn't need to exist (could just document all that stuff as GuildTextableChannel and ignore the fact that some of the types which will be covered by that won't exist in the future) and WebhookChannelTypes exists for cases where someone would want to use an instance check to narrow down a GuildTextableChannel to a WebhookChannelT so it's about as unnecessary as that.

davfsa: Also, aren't these both just `TextableGuildChannel`?

FasterSpeeding: No because one includes thread channels and the other doesn't, this is covered the pr's description

davfsa: Considering it is only two types, I would lend towards specifying both instead of having a random type which pollutes the root package

hikari/channels.py
```diff
@@ -1261,7 +1270,7 @@ class GuildTextChannel(GuildChannel, TextChannel):
 
 
 @attr.define(hash=True, kw_only=True, weakref_slot=False)
-class GuildNewsChannel(GuildChannel, TextChannel):
+class GuildNewsChannel(TextableGuildChannel):
```
davfsa, line 1273: Isnt a news channel also a GuildTextChannel?

hikari/channels.py
FasterSpeeding: which then leads to cases where the types on specific values come out of sync with the actual values as stuff is added but only some of the methods which refer to that specific union are updated

FasterSpeeding: Evened it out a bit by removing some top level types which were just `Union[ActualType, str/int]`

hikari/channels.py
```diff
@@ -42,9 +43,10 @@
     "GuildStoreChannel",
     "GuildVoiceChannel",
     "GuildStageChannel",
+    "WebhookChannelT",
```
davfsa, line 46: No need to export these

FasterSpeeding: There won't be any ref for what it refers to in the docs if it's not exported

davfsa: Then the `__init__.py` would have to be changed to not export these to the root, as we don't put typehints there

FasterSpeeding: This was never an issue in the past so I don't see what's changed and if you really want to remove all type variables from the top level the go ahead ig but they're probably more useful to than most of the classes which are exported on the top level so...

hikari/channels.py
davfsa: Could the be fetched by using the following to replace them? 

```py
import typing


assert isinstance(TextChannel(...), typing.get_args(WebhookChannelsT)) # True
# or (without typing module)
assert isinstance(TextChannel(...), WebhookChannelsT.__args__) # True, but __args__ is undocumented
```

FasterSpeeding: Since typing.get_args is typed as returning `Tuple[Any, ...]` that would do absolutely nothing to narrow the scope of a variable with type checkers

davfsa: That Tuple can be used for isinstances, which is what this variable is now used for

FasterSpeeding: which won't be compatible with type checkers

davfsa (approved):

PR #701, opened 2021-07-22
Move command models and rename interaction modules

FasterSpeeding:
### Summary
Move command models and rename interactions modules

This better follows the pattern seen in hikari.events although tbh I'm not sure how i fee about that pattern since it kinda goes against the whole flat instead of nested mentality. In another project i've been truing a pattern more like
```
src
  - events
    - __init__
    - _event_group_a
    - _event_group_b
```
where the enums and classes in the specific event group modules are all exported in `event.__init__` while the event group modules themselves are kept private to keep the public interface flatter and avoiding having too many different paths for importing the same classes and enums (and also keeping them shorter cause tbh `hikari.events.message_events.MessageCreateEvent` is both repeating information and kinda verbose).

This pattern would possibly apply to the interactions stuff as such 
```
src
  - interactions
    - __init__
    - _bases
    - _commands
    - _components
```

### Checklist


hikari/commands.py
```diff
@@ -0,0 +1,294 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Models and enums used for application commands on Discord."""
+from __future__ import annotations
+
+__all__: typing.List[str] = [
+    "Command",
+    "CommandChoice",
+    "CommandOption",
+    "OptionType",
```
FasterSpeeding, line 30: ,

FasterSpeeding:
bump

hikari/interactions/command_interactions.py
```diff
@@ -24,16 +24,12 @@
...
-    "CommandChoice",
     "CommandInteractionOption",
     "CommandInteraction",
-    "CommandOption",
     "COMMAND_RESPONSE_TYPES",
     "CommandResponseTypesT",
     "InteractionChannel",
     "ResolvedOptionData",
-    "OptionType",
 ]
```
davfsa, line 33: Why not move everything to commands.py?

FasterSpeeding: Like detailed before ths just more strictly follows the current pattern where dispatch classes are kept separate from general entity classes

davfsa (approved):

PR #703, opened 2021-07-22
Fix some bugs in voice impl

### LINKED ISSUE #693: Seperate voice disconnect
Something that just popped my mind, we dont actually have a disconnect a specific channels, we only have a full disconnect from all channels. Should we change disconnect to allow passing a channel to disconnect from and moving the closing all connections to `close`?

(This would be done in another pr, since it is a bit out of scope for this one)

_Originally posted by @davfsa in https://github.com/hikari-py/hikari/pull/629#discussion_r663154067_

davfsa:
- Allow disconnecting from voice channel
- ~~Allow moving voice channel in the same server~~

### Checklist

### Related issues
Closes #693

### DIFF
--- a/hikari/impl/voice.py
+++ b/hikari/impl/voice.py
@@ -77,22 +77,35 @@ class VoiceComponentImpl(voice.VoiceComponent):
         if self._is_closing:
             raise errors.ComponentStateConflictError("Component cannot be used while it's closing")
 
-    async def _disconnect(self) -> None:
-        if self._connections:
-            _LOGGER.info("shutting down %s voice connection(s)", len(self._connections))
-            # We rely on the assumption that _on_connection_close will be called here rather than explicitly
-            # emptying self._connections.
-            await asyncio.gather(*(c.disconnect() for c in self._connections.values()))
-
-    async def disconnect(self) -> None:
+    async def disconnect(self, guild: snowflakes.SnowflakeishOr[guilds.PartialGuild]) -> None:
         self._check_if_alive()
-        await self._disconnect()
+        guild_id = snowflakes.Snowflake(guild)
+
+        if guild_id not in self._connections:
+            raise errors.VoiceError("This application doesn't have any active voice connection in this server")
+
+        conn = self._connections[guild_id]
+        # We rely on the assumption that _on_connection_close will be called here rather than explicitly
+        # to remove the connection from self._connections.
+        await conn.disconnect()
+
+    async def _disconnect_all(self) -> None:
+        # We rely on the assumption that _on_connection_close will be called here rather than explicitly
+        # emptying self._connections.
+        await asyncio.gather(*(c.disconnect() for c in self._connections.values()))
+
+    async def disconnect_all(self) -> None:
+        await self._disconnect_all()
 
     async def close(self) -> None:
         self._check_if_alive()
         self._is_closing = True
         self._app.event_manager.unsubscribe(voice_events.VoiceEvent, self._on_voice_event)
-        await self._disconnect()
+
+        if self._connections:
+            _LOGGER.info("shutting down %s active voice connection(s)", len(self._connections))
+            await self._disconnect_all()
+
         self._is_alive = False
         self._is_closing = False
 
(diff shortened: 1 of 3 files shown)


FasterSpeeding:
> Allow disconnecting from voice channel

Isn't this already possible by just doing the following.
await `voice_client.connections[guildId].disconnect()`

> Allow moving voice channel in the same server

Shouldn't this be done by updating the voice state's channel for that guild rather than by nuking any previously tracked state for that guild's voice connection(?) and generating a new connection (both of which were already possible but now without any safeguard to prevent accidentally nuking a previous connection unless I'm missing something)?

davfsa:
I do agree with you in the connect_to function, will revert that, but I disagree with your opinion on the disconnect.

Having a disconnect function that kills all the active connections is not something useful (this is something you would expect from the close function), and having to check the dict manually yourself is not something we should promote (connections should actually be private), so I won't be reverting that.

FasterSpeeding:
> Having a disconnect function that kills all the active connections is not something useful (this is something you would expect from the close function)

Close also deactivates it though, are you saying that you'd expect to have to close and then restart it up again to disconnect all current connections while in the process possibly also missing new connections because it'd be racing with wherever new connections are being made from?

davfsa:
I cant see why you would need to close all active connections unless you are shutting it down.

FasterSpeeding:
broken tests

FasterSpeeding:
> I cant see why you would need to close all active connections unless you are shutting it down.

random thought, what if a bot is made for a specific community or event and during certain periods it streams audio into one or more voice channels and when that period is over it wants to leave all those channels at once. Or maybe a bot wants to leave all the voice channels at once because the feature which used voice is being temporarily disabled possibly due to maintenance reasons or to lower load on their backend

FasterSpeeding:
broken tests

davfsa:
> > I cant see why you would need to close all active connections unless you are shutting it down.
> 
> random thought, what if a bot is made for a specific community or event and during certain periods it streams audio into one or more voice channels and when that period is over it wants to leave all those channels at once. Or maybe a bot wants to leave all the voice channels at once because the feature which used voice is being temporarily disabled possibly due to maintenance reasons or to lower load on their backend

For both cases here, closing down the voice impl would be ideal, and this would have to have some forwards thinking from whatever dev team is doing it.

For case one, if the event is done, you just close down the voice impl,  no need for it anymore.

For case two, temporarily disabled would mean it cant be used by anyone meaning, again,  that it should be closed.

*For both cases they would have to handle the ComponentStateErrors to give back feedback of the feature being disabled*

FasterSpeeding:
> For case one, if the event is done, you just close down the voice impl, no need for it anymore.

> For case two, temporarily disabled would mean it cant be used by anyone meaning, again, that it should be closed.

> For both cases they would have to handle the ComponentStateErrors to give back feedback of the feature being disabled

You shouldn't really be closing internal parts of a bot client directly as this may lead to unexpected behaviour since it is well you know managed by the bot and other stuff will may be interacting with it. Especially when you have no way to consistently start them up again since when you're consuming them as the generic interfaces rather than impls since how they're started up or run is impl detail and these "start" methods you have on the interfaces in some places don't work well in environments such as structured async loops where you can't just off handedly spawn background tasks like with Asyncio and would likely just raise NotImplmenetedError for some environments

hikari/impl/voice.py
```diff
@@ -77,22 +77,36 @@ def _check_if_alive(self) -> None:
...
-    async def disconnect(self) -> None:
+    async def disconnect(self, guild: snowflakes.SnowflakeishOr[guilds.PartialGuild]) -> None:
         self._check_if_alive()
-        await self._disconnect()
+        guild_id = snowflakes.Snowflake(guild)
+
+        if guild_id not in self._connections:
+            raise errors.VoiceError("This application doesn't have any active voice connection in this server")
+
+        shard_id = snowflakes.calculate_shard_id(self._app, guild_id)
```
FasterSpeeding, line 87: Most of this stuff is done in _on_connection_close

davfsa: Removed duplicated logic :+1:

FasterSpeeding (approved):

hikari/impl/voice.py
```diff
@@ -77,22 +77,35 @@ def _check_if_alive(self) -> None:
...
+        # to remove the connection from self._connections.
+        await conn.disconnect()
+
+    async def _disconnect_all(self) -> None:
+        # We rely on the assumption that _on_connection_close will be called here rather than explicitly
+        # emptying self._connections.
+        await asyncio.gather(*(c.disconnect() for c in self._connections.values()))
+
+    async def disconnect_all(self) -> None:
+        await self._disconnect_all()
```
FasterSpeeding, line 98: check if alive

PR #706, opened 2021-07-26
Reformat generated stubs

davfsa:


FasterSpeeding:
Probably wanna also fix the fact that this just isn't don't what it's intended to on the top level

FasterSpeeding (approved):

pipelines/mypy.nox.py
```diff
@@ -47,10 +47,11 @@ def generate_stubs(session: nox.Session) -> None:
 def _generate_stubs(session: nox.Session) -> None:
     session.run("stubgen", *STUBGEN_GENERATE, "-o", ".", "--include-private", "--no-import")
 
-    for path in STUBGEN_GENERATE:
-        # We add "i" because we turn ".py" into ".pyi"
-        stub_path = path + "i"
+    stub_paths = [path + "i" for path in STUBGEN_GENERATE]
 
+    session.run("black", *stub_paths)
```
FasterSpeeding, line 52: Probably want to also be running isort over these since it'd change the styling

FasterSpeeding (changes requested):

FasterSpeeding (approved):

PR #708, opened 2021-07-28
Add some simplifications for external importers

FasterSpeeding:
### Summary
* Add api + impl to top level imports plus
* Flatten down impl and api in a similar style to events and interactions so that all exported items can be accessed via hikari.api/hikari.impl

### Checklist


hikari/__init__.py
```diff
@@ -72,11 +74,11 @@
...
-from hikari.impl.bot import BotApp
-from hikari.impl.bot import GatewayBot
-from hikari.impl.rest import ClientCredentialsStrategy
-from hikari.impl.rest import RESTApp
-from hikari.impl.rest_bot import RESTBot
+from hikari.impl import BotApp
+from hikari.impl import ClientCredentialsStrategy
+from hikari.impl import GatewayBot
+from hikari.impl import RESTApp
+from hikari.impl import RESTBot
```
davfsa, line 81: This import method is not something we have used until now. We always have the full qualified path to to object (how it was before)

hikari/__init__.py
```diff
@@ -30,6 +30,8 @@
 
 import os as _os
 
+from hikari import api
+from hikari import impl
```
davfsa, line 34: What does this achieve?

hikari/__init__.py
davfsa, line 81: This import method is not something we have used until now. We always have the full qualified path to to object (how it was before)

FasterSpeeding: Except you kinda don't in `__init__.py` files, in those you do stuff like `hikari.events import *` to import from hikari.events and stuff from in modules within hikari.events which is just an equivalent

hikari/__init__.py
FasterSpeeding: That behaviour is specific to MYPY, not all type checkers will see api or impl as members on the imported module `hikari` unless you explicitly export them, same applies to every other module and the current behaviour is that you have to directly import them if you're using something like pyright which as far as I could tell was what you'd intended anyways since well you just don't explicitly export modules within other modules at all. This is just a special case since the contents of those modules is mostly left not exposed on the top level so actually being able to access them from the hikari import should be common enough

hikari/__init__.py
FasterSpeeding: Also what you described there is kinda unnecessary and if anything possibly even slow
Doing the following

```py
from hikari.events import channel_events
from hikari.events import guild_events
from hikari.events import interaction_events
from hikari.events import lifetime_events
from hikari.events import member_events
from hikari.events import message_events
from hikari.events import reaction_events
from hikari.events import role_events
from hikari.events import shard_events
from hikari.events import typing_events
from hikari.events import user_events
from hikari.events import voice_events
```

when you can just do while accesing the event types on events instead of their specific modules

```py
from hikari.events
```

honestly seems like a lot of unnecessary import calls

FasterSpeeding:
bump

davfsa (approved):

PR #710, opened 2021-07-30
CommandInteractionOption typing fix

FasterSpeeding:
### Summary
Command options only have 1 value

### Checklist

### DIFF
--- a/hikari/interactions/commands.py
+++ b/hikari/interactions/commands.py
@@ -367,7 +367,7 @@ class CommandInteractionOption:
     type: typing.Union[OptionType, int] = attr.field(repr=True)
     """Type of this option."""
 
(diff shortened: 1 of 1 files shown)


hikari/interactions/commands.py
```diff
@@ -367,7 +367,7 @@ class CommandInteractionOption:
     type: typing.Union[OptionType, int] = attr.field(repr=True)
     """Type of this option."""
 
-    value: typing.Optional[typing.Sequence[typing.Union[str, int, bool]]] = attr.field(repr=True)
+    value: typing.Optional[typing.Union[str, int, bool]] = attr.field(repr=True)
```
davfsa, line 370:
```suggestion
    value: typing.Union[str, int, bool, None] = attr.field(repr=True)
```

davfsa (approved):

PR #728, opened 2021-08-15
Cleanup setup.py

davfsa:
### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,14 +19,11 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-
 import os
 import re
 import types
 
 import setuptools
-
-name = "hikari"
 
 
 def long_description():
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,14 +19,15 @@
...
 import os
 import re
+import sys
 import types
 
 import setuptools
 
-name = "hikari"
+if sys.version_info < (3, 8, 0):
+    raise RuntimeError("hikari requires Python 3.8.0 or greater")
```
FasterSpeeding, line 30: isn't this what python_requires is for

setup.py
```diff
@@ -68,22 +69,23 @@ def parse_requirements_file(path):
     maintainer_email=metadata.email,
     license=metadata.license,
     url=metadata.url,
+    python_requires=">=3.8.0",
```
FasterSpeeding, line 72: the project can't assume it's forward compatible though; it's already known that it definitely won't be compatible with the python ver which releases in just over 4 years from now and that's only accounting for one PEP

setup.py
davfsa: Unfortunately `python_requires` its only an indicator. Pip enforces that indicator but if you run `python setup.py install`, it completely ignores it. This makes sure its enforced

FasterSpeeding: Well one just because it cannot be installed into an environment doesn't mean it can't be built in that environment and that check seems to cover a wider scope than intended. 
Also that behaviour kinda sounds like normal expected behaviour? It's pip which is strict about this stuff; you don't get stuff like strict requirement cross-checks when you install using setup.py either and probably shouldn't be using setup.py directly if you want a safe install (wouldn't ever have setup.cfg raising because of that stuff either).

FasterSpeeding (approved):

PR #732, opened 2021-08-19
Clean up attr checks on `_EnumNamespace.__setitem__()`.

nxtlo:
### Summary

Cleans up some `or`'s attr checks in `_EnumNamespace.__setitem__(...)`

### Checklist


davfsa:
Thanks for the PR, but this seems to include a bit more complexity than before. I rather keep it like it was before due to it being both clearer (imo) and simplex logic wise

nxtlo:
Probably isn't plain simple and is more complex yeh. I agree with that. I will close the PR for now.

PR #733, opened 2021-08-19
Add has method to permissions

IkBenOlie5:
### Summary
Add a `has` method to the permissions enum. It's confusing for beginners to do:
```py
bool(permissions_1 & permissions_2)
```
and this should make that easier.

### Checklist

### DIFF
--- a/hikari/permissions.py
+++ b/hikari/permissions.py
@@ -270,6 +270,9 @@ class Permissions(enums.Flag):
     USE_PRIVATE_THREADS = 1 << 36
     """Allows for creating and participating in private threads."""
 
+    def has(self, permission: Permissions) -> bool:
+        return bool(self & permission)
+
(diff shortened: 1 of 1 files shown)


davfsa:
*Reason for closure was discussed in Discord, but just wanted to leave this here in case anybody stumbles upon this and is confused about it*

There is already an extensive documentation on how to do these actions in the [Permissions docstring](https://github.com/hikari-py/hikari/blob/master/hikari/permissions.py#L38-L90) as well as in the [IntFlag docstring](https://github.com/hikari-py/hikari/blob/master/hikari/internal/enums.py#L452-L592). Furthermore, IntFlag provides similar methods like `.all()` and `.any()`. If anybody is still confused and needs help, make sure to drop by the Discord and we will try to help you to the best of our abilities

PR #739, opened 2021-08-22
Install speedups in mypy pipeline to load stubs for them

davfsa:
### Checklist

### DIFF
--- a/pipelines/mypy.nox.py
+++ b/pipelines/mypy.nox.py
@@ -35,7 +35,7 @@ STUBGEN_GENERATE = [
 @nox.session(reuse_venv=True)
 def mypy(session: nox.Session) -> None:
     """Perform static type analysis on Python source code."""
-    session.install("-r", "requirements.txt", "-r", "dev-requirements.txt")
+    session.install("-r", "requirements.txt", "-r", "speedup-requirements.txt", "-r", "dev-requirements.txt")
 
     _generate_stubs(session)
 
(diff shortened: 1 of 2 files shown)


FasterSpeeding (dismissed):

pipelines/mypy.nox.py
```diff
@@ -46,7 +46,7 @@ def mypy(session: nox.Session) -> None:
 @nox.session(reuse_venv=True)
 def generate_stubs(session: nox.Session) -> None:
     """Generate the stubs for the package."""
-    session.install("-r", "requirements.txt", "-r", "dev-requirements.txt")
+    session.install("-r", "requirements.txt", "-r", "speedup-requirements.txt", "-r", "dev-requirements.txt")
```
FasterSpeeding, line 49: Does it actually need to install any of the standard requirements to generate the stubs?

davfsa: Yes, stubs are available for some of them, but it's best to install them than installing the stubfiles with mypy, as they are more up to date. An example of this can be seen with the new API for attrs, which we make use of

FasterSpeeding: Does that actually effect generating the `__init__.pyi` files though?

davfsa: Oh, didn't see this was stubgen. No, in that case we should really only need dev-requirement I think. Will test it remove the other two if we don't need them :+1:

davfsa: Fixed :+1:

FasterSpeeding (approved):

PR #742, opened 2021-08-23
Consistency fixes for `__all__` usages

davfsa:
### Checklist


davfsa:
Also, random thing I found out while doing this (just for fun), if we turn all `__all__` into tuples, we save about ~1MB of memory if the package is imported :)

FasterSpeeding (approved):

PR #759, opened 2021-08-30
Add missing audit log change keys

FasterSpeeding:
### Summary
Add missing audit log change keys
(Plus improve the duped entries warning on enums)

### Checklist

- [x] I have run `nox` and all the pipelines have passed.
- [x] I have made unittests according to the code I have added/modified/deleted.

PR #762, opened 2021-08-30
Add towncrier to generate CHANGELOG.md

davfsa:
Use towncrier to generate CHANGELOG.md

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 updates:
   # Github actions
-  - package-ecosystem: "github-actions"
+  - package-ecosystem: github-actions
     directory: /
     schedule:
       interval: daily
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,14 +184,6 @@ jobs:
 
     runs-on: ubuntu-20.04
 
-    env:
-      GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
-      GITHUB_TAG: ${{ github.event.release.tag_name }}
-      GITHUB_REPO_SLUG: ${{ github.repository }}
-      DEPLOY_WEBHOOK_URL: ${{ secrets.DEPLOY_WEBHOOK_URL }}
-      TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
-      TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
-
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2.3.4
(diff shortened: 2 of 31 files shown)


FasterSpeeding:
With all this fancy tech how would someone link to the current changelog for #master

davfsa:
That is unfortunately not possible. The only reason why I would like to use this is because having a shared file that everyone updates leads to conflict errors between PRS. I have not been able to find a better tool for this

FasterSpeeding:
> That is unfortunately not possible. The only reason why I would like to use this is because having a shared file that everyone updates leads to conflict errors between PRS. I have not been able to find a better tool for this

Also since you seem to be assuming that every change will be linked to a PR here even if that already introduced weird behaviour like linking for #759 for a change which isn't actually in 759 couldn't this just be generated from the recent PR titles on a regular basis or something

FasterSpeeding:
> Edit: A dirty way to generate the CHANGELOG when needed would be to run towncrier --draft

also that isn't overly helpful for anybody who's just going over the repo without having it clone which will likely be most ppl who pip install straight from the repo and nobody's going to want to clone a repo just to check the up to date readme

davfsa:
> Also since you seem to be assuming that every change will be linked to a PR here even if that already introduced weird behaviour like linking to #759 for a change which isn't actually in 759 couldn't this just be generated from the recent PR titles and tags on a regular basis or something 

All changes should have PR related to them, 759 is an exception due to a miss push to master, which should not happen again now that it's blocked from pushes even for administrators.

Furthermore, PR titles might not be the most descriptive and are very limited on what you can add, but the CHANGELOG can (and should) hold much more information if needed (multilines)

FasterSpeeding:
> Furthermore, PR titles might not be the most descriptive and are very limited on what you can add, but the CHANGELOG can (and should) hold much more information if needed (multilines)

PRs do also have descriptions, no reason you couldn't also include that

davfsa:
> PRs do also have descriptions, no reason you couldn't also include that

Our PR descriptions hold much more information than should be added to a CHANGELOG, and then tend to be more contributor friendly than maintainer friendly.

Adding to the generation based on the PR titles, some PRS target other aspects of the library that have nothing to do with what a user cares about (to give the most recent example, disabling and re-enabling the CI for CPython 3.10)

FasterSpeeding:
If you insist then we can go ahead with this IG but you'll have to add new CHANGELOG additions first

davfsa:
> If you insist then we can go ahead with this IG but you'll have to add new CHANGELOG additions first

I dont want to force this on the project, I just want to find the best possible solution, which in this case (based on the research I have and the alternatives available) is using `towncrier`, which is already being used by different python projects (including pip and aiohttp). I will have to change the PR to better clear some edge cases you mentioned, maybe even always enforce a changelog file to be added, but have the "irrelevant" ones under another sort of prefix to still have them if we ever needed them (changing the prefix will cause it to appear in the CHANGELOG).

I will have a look into other possible nicer ways of doing this

FasterSpeeding:
Also however the changelog's gonna work, it needs to be added to codespell

davfsa:
Still some things to finish, will do today

davfsa:
Should be good to for another review now :+1:

FasterSpeeding (approved):

PR #763, opened 2021-08-30
Feature/747 retry request on 5xx error

### LINKED ISSUE #747: Toggle to allow n amount of automatic-retries for internal server errors
### Summary
For this feature request I propose that an option be added to the rest client to allow for automatic retries on some if not all 5xx error codes either unlimited or up to a certain number of retries.

### Problem
Right now the behaviour of raising on 500s can lead to the code consuming the REST client having to put a ton of messy retry logic around requests to handle both 500s and ratelimited errors.

### Ideal implementation
Add an argument to the rest client and relevant bot clients which defaults to idk maybe 5 retires and allows for specify the amount of times the lib should automatically retry on a 5xx error code (with it likely using an exponential back off strategy between requests) which can be set to 0 to signify that it should always raise. Alternatively this argument could be a boolean  argument where true signifies unlimited retries on 5xx and False signifies always raise

### Checklist

- [ ] I have searched the issue tracker and have made sure it's not a duplicate.

ahnaf-zamil:
Added retry feature for HTTP requests. If a request fails with a `5xx` status, it will retry for `n` amount of times. By default, the amount of retries is 5. It is passed as an argument to the `RESTClient` constructor as `max_retries`.

### Checklist

### Related issues
Closes #747


hikari/impl/rest.py
```diff
@@ -549,6 +550,7 @@ def __init__(
         executor: typing.Optional[concurrent.futures.Executor],
         http_settings: config.HTTPSettings,
         max_rate_limit: float,
+        max_retries: int = 5,
```
davfsa, line 553: Wouldnt it be better to default at 3 with a top at 5?

hikari/impl/rest.py
```diff
@@ -696,6 +699,10 @@ async def _request(
         headers.put(_X_AUDIT_LOG_REASON_HEADER, reason)
 
         url = compiled_route.create_url(self._rest_url)
+
+        backoff = None
```
davfsa, line 703: To aid people reading the code and linters, might be best to document this and add typehints
```suggestion
        # This is initiated the first time we hit a 5xx error to save memory when nothing goes wrong
        backoff: typing.Optional[rate_limits.ExponentialBackoff] = None
```

hikari/impl/rest.py
```diff
@@ -759,6 +765,23 @@ async def _request(
...
+                # Handling 5xx errors
+                if 500 <= response.status < 600 and retries_done <= self._max_retries and self._max_retries != 0:
+                    if backoff is None:
+                        backoff = rate_limits.ExponentialBackOff()
+
+                    sleep_time = next(backoff)
+                    _LOGGER.warning(
+                        "Received status %s on request, backing off for %ss and retrying. Retries remaining: %s",
+                        response.status,
+                        round(sleep_time, 1),
```
davfsa, line 777: No need to round, being verbose is better here, as this is meant for debug purposes

```suggestion
                    _LOGGER.warning(
                        "Received status %s on request, backing off for %.2fs and retrying. Retries remaining: %s",
                        response.status,
                        sleep_time,
```

hikari/impl/rest.py
```diff
@@ -759,6 +765,23 @@ async def _request(
                     real_url = str(response.real_url)
                     raise errors.HTTPError(f"Expected JSON [{response.content_type=}, {real_url=}]")
 
+                # Handling 5xx errors
+                if 500 <= response.status < 600 and retries_done <= self._max_retries and self._max_retries != 0:
+                    if backoff is None:
+                        backoff = rate_limits.ExponentialBackOff()
```
davfsa, line 771: Having some default values here will help not have a huge backoff

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "lunr": "^2.3.7"
+  }
+}
```
davfsa, line 5: :thinking:

hikari/impl/rest.py
```diff
@@ -729,7 +736,6 @@ async def _request(
                         proxy=self._proxy_settings.url,
                         proxy_headers=self._proxy_settings.all_headers,
                     )
-
```
davfsa, line 732: Random line removal

hikari/impl/rest.py
```diff
@@ -759,6 +765,23 @@ async def _request(
                     real_url = str(response.real_url)
                     raise errors.HTTPError(f"Expected JSON [{response.content_type=}, {real_url=}]")
 
+                # Handling 5xx errors
+                if 500 <= response.status < 600 and retries_done <= self._max_retries and self._max_retries != 0:
```
davfsa, line 769:
```suggestion
                if 500 <= response.status < 600 and retries_done < self._max_retries:
```

davfsa (changes requested):

package.json
ahnaf-zamil: I don't know what changed here, happened on it's own

hikari/impl/rest.py
ahnaf-zamil: ¯\_(ツ)_/¯

hikari/impl/rest.py
ahnaf-zamil: I don't understand, could you please explain @davfsa ?

hikari/impl/rest.py
ahnaf-zamil: Alright, committing this

hikari/impl/rest.py
ahnaf-zamil: In the issue, @FasterSpeeding suggested to put 5 retries. But are you sure you want to limit it to 5? Since this feature was about "giving more control to users", they should be able to do as many retries as they want to be honest.

ahnaf-zamil:
Linting is failing, let me fix it.

ahnaf-zamil (commented):
Finished working on these requested changes

hikari/impl/rest.py
```diff
@@ -503,6 +504,11 @@ class RESTClientImpl(rest_api.RESTClient):
 
         This is provided since some endpoints may respond with non-sensible
         rate limits.
+    max_retries : typing.Optional[builtins.int]
+        Maximum number of times a request will be retried if
+        it fails with a `5xx` status. Defaults to 3 if set to `builtins.None`.
+
+        If you provide above 5, it will default to 5.
```
davfsa, line 511:
```suggestion
        If you provide a value above 5, it will default to 5.
```
Just a little wording issue

hikari/impl/rest.py
davfsa: It's back :eyes:

hikari/impl/rest.py
```diff
@@ -759,6 +772,23 @@ async def _request(
                     real_url = str(response.real_url)
                     raise errors.HTTPError(f"Expected JSON [{response.content_type=}, {real_url=}]")
 
+                # Handling 5xx errors
+                if response.status in _RETRY_ERROR_CODES and retries_done < self._max_retries:
+                    if backoff is None:
+                        backoff = rate_limits.ExponentialBackOff(maximum=16)
```
davfsa, line 778: Best not to hardcode this here. Create a global value ontop and use it here, similar to how shards use the ExponentialBackOff

davfsa (changes requested):
You are also missing being able to set the argument on GatewayBot, RESTBot and RESTApp like we do for `max_rate_limit`.

Everything should be good after than I will add the missing tests. Thanks for working on it :+1:

hikari/impl/rest.py
ahnaf-zamil: I really don't know why it is back

hikari/impl/rest.py
ahnaf-zamil: Understood

hikari/impl/bot.py
```diff
@@ -167,6 +167,11 @@ class GatewayBot(traits.GatewayBotAware):
         Note that this only applies to the REST API component that communicates
         with Discord, and will not affect sharding or third party HTTP endpoints
         that may be in use.
+    max_retries : typing.Optional[builtins.int]
+        Maximum number of times a request will be retried if
+        it fails with a `5xx` status. Defaults to 3 if set to `builtins.None`.
+
+        If you provide a value above 5, it will default to 5.
```
FasterSpeeding, line 174: If we're gonna have a cap for this it should really error if they try to go over the cap to be explicit instead of implicit

ahnaf-zamil: Instead of throwing an error, I think that it should rather log it onto the console like "A value higher than 5 has been provided for max_retries, going for 5 retries"

hikari/impl/rest.py
```diff
@@ -288,13 +290,17 @@ class RESTApp(traits.ExecutorAware):
...
+        Maximum number of times a request will be retried if
+        it fails with a `5xx` status. Defaults to 3 if set to `builtins.None`.
     proxy_settings : typing.Optional[hikari.config.ProxySettings]
         Proxy settings to use. If `builtins.None` then no proxy configuration
         will be used.
     url : typing.Optional[builtins.str]
         The base URL for the API. You can generally leave this as being
         `builtins.None` and the correct default API base URL will be generated.
 
+
```
davfsa, line 303: New line 👀

davfsa (changes requested):

FasterSpeeding (approved):

davfsa (approved):

PR #770, opened 2021-09-03
Add is_for_emoji methods to the relevant reaction events

FasterSpeeding:
### Summary
Add is_for_emoji methods to relevant reaction events, as suggested by @tandemdude 

### Checklist

### DIFF
--- a/hikari/events/reaction_events.py
+++ b/hikari/events/reaction_events.py
@@ -47,13 +47,13 @@ import typing
 
 import attr
 
+from hikari import emojis
 from hikari import intents
 from hikari.events import base_events
 from hikari.events import shard_events
 from hikari.internal import attr_extensions
 
 if typing.TYPE_CHECKING:
-    from hikari import emojis
     from hikari import guilds
     from hikari import snowflakes
     from hikari import traits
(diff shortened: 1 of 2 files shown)


hikari/events/reaction_events.py
```diff
@@ -258,6 +292,23 @@ def emoji_id(self) -> typing.Optional[snowflakes.Snowflake]:
...
+            The emoji to check.
+
+            Passing `builtins.str` here indicates a unicode emoji.
+
+        Returns
+        -------
+        builtins.bool
+            Whether the emoji is the one which was removed.
+        """
+        return emoji.id == self.emoji_id if isinstance(emoji, emojis.CustomEmoji) else emoji == self.emoji_name
```
davfsa, line 310: This could be moved to `ReactionEvent` to avoid code duplication

hikari/events/reaction_events.py
```diff
@@ -216,6 +233,23 @@ def emoji_id(self) -> typing.Optional[snowflakes.Snowflake]:
...
```
davfsa, line 251: This isnt exactly correct, as passing `None` as the emoji will return `True` if Discord dont send the emoji_name.

hikari/events/reaction_events.py
FasterSpeeding: reaction event doesn't have any emoji properties on it

hikari/events/reaction_events.py
FasterSpeeding: It doesn't document any support for passing None

davfsa: Fair enough

hikari/events/reaction_events.py
davfsa: Should it 🤔

FasterSpeeding: It can't because ReactionDeleteAllEvent event doesn't have a relevant emoji

davfsa (approved):

PR #772, opened 2021-09-03
Add information about versioning scheme and deprecation

davfsa:


CONTRIBUTING.md
```diff
@@ -10,6 +10,18 @@ of conduct can lead to a ban from the project and a report to GitHub.
...
+# Versioning scheme
+
+This project follows the versioning scheme stated [here](https://semver.org/), but with a
+slight modification to the pre-release versioning: we denote all pre-releases with `.dev{num}`.
+
+The development version number is increased automatically after each release in master.
+
+# Deprecation process
+
+The removal or rename of anything facing the public facing API must undergo the a removal process, which should oblide to
```
FasterSpeeding, line 22: need to add this file to spell checking, "oblige" not "oblide" and even then that phrasing is weird since oblige would generally donate a legal or moral obligation.

FasterSpeeding, line 22: "renaming" not rename

FasterSpeeding, line 22: "on the public facing API" seems better than "facing the public facing API"

FasterSpeeding, line 22: "go through" would be easier to say and interpret than "undergo" in this context. Also that "the a" seems like a typo

CONTRIBUTING.md
```diff
@@ -10,6 +10,18 @@ of conduct can lead to a ban from the project and a report to GitHub.
 
 You can read the code of conduct [here](https://github.com/hikari-py/hikari/blob/master/CODE_OF_CONDUCT.md).
 
+# Versioning scheme
+
+This project follows the versioning scheme stated [here](https://semver.org/), but with a
+slight modification to the pre-release versioning: we denote all pre-releases with `.dev{num}`.
+
+The development version number is increased automatically after each release in master.
```
FasterSpeeding, line 18: probably want to specify that by "master" you mean the master repo branch

CONTRIBUTING.md
```diff
@@ -10,6 +10,18 @@ of conduct can lead to a ban from the project and a report to GitHub.
 
 You can read the code of conduct [here](https://github.com/hikari-py/hikari/blob/master/CODE_OF_CONDUCT.md).
 
+# Versioning scheme
+
+This project follows the versioning scheme stated [here](https://semver.org/), but with a
+slight modification to the pre-release versioning: we denote all pre-releases with `.dev{num}`.
```
FasterSpeeding, line 16: Should probably link to the versioning format PEP this project follows since that's where ,dev comes from, it isn't a custom thing (might be PEP 440 but i'm not gonna check that rn).

FasterSpeeding (changes requested):

CONTRIBUTING.md
davfsa: Yep, I initially had PEP 440 linked, but the current link explains simple versioning better as well as what is accepted in each release. Will add both just to be safe :)

CONTRIBUTING.md
```diff
@@ -10,6 +10,21 @@ of conduct can lead to a ban from the project and a report to GitHub.
 
 You can read the code of conduct [here](https://github.com/hikari-py/hikari/blob/master/CODE_OF_CONDUCT.md).
 
+# Versioning scheme
+
+This project follows the versioning scheme stated by [PEP 440](https://www.python.org/dev/peps/pep-0440/), but with a
+slight modification to the pre-release versioning: we denote all pre-releases with the `.dev{num}` extension.
```
FasterSpeeding, line 16: this bit about a modification isn't a modification anymore since PEP 440 includes that

davfsa: PEP 440 specifies other types of [pre-releases](https://www.python.org/dev/peps/pep-0440/#pre-releases), which we don't allow. Thought this would clarify that, but I will make the `all` bold if that helps to get the point across

FasterSpeeding: So this project doesn't allow beta releases?

FasterSpeeding: pretty sure doing every pre-release as a dev release falls under PEP 440 without modifying it at all but ig that's just me

davfsa: Currently, from the strict implementation of HikariVersion (https://github.com/hikari-py/hikari/blob/master/hikari/internal/ux.py#L264-L315) it doesn't. Code there could be modified to support it

My idea of releases has been `MAJOR.MINOR.PATCH.dev{num}` for all beta, alpha and release candidate versions, with `MAJOR.MINOR.PATCH` for the latest versions. Im open to change it tho

FasterSpeeding: the lib still does alpha and beta release though it just doesn't explicitly version them and instead marks them all as dev releases which doesn't modify PEP 440, just ignores a part of it which isn't a modification

davfsa: Ok, will remove that part then :+1:

FasterSpeeding (approved):

PR #776, opened 2021-09-06
Add ApplicationCommand.version

FasterSpeeding:
### Checklist

### DIFF
--- a/hikari/commands.py
+++ b/hikari/commands.py
@@ -182,7 +182,7 @@ class Command(snowflakes.Unique):
     """
 
     version: snowflakes.Snowflake = attr.field(eq=False, hash=False, repr=True)
-    """Auto-incrementing version identifier updated during seubstantial record changes."""
+    """Auto-incrementing version identifier updated during substantial record changes."""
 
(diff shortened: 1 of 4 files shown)


hikari/commands.py
```diff
@@ -181,6 +181,9 @@ class Command(snowflakes.Unique):
     This will be `builtins.None` if this is a global command.
     """
 
+    version: snowflakes.Snowflake = attr.field(eq=False, hash=False, repr=True)
+    """Auto-incrementing version identifier updated during seubstantial record changes."""
```
davfsa, line 185: This might confuse people reading it, might be worth changing it up a little and making it easier to understand

FasterSpeeding: what about it would confuse ppl

davfsa: I find personally confusing that the type of the attribute is a snowflake but the docstring suggests it should be an int, since it is incremented sequencially every deployment, nothing to do with a snowflake

FasterSpeeding: It is incremented from the commands Id, it's a snowflake

davfsa (approved):

PR #779, opened 2021-09-10
Fix `embeds` type error in edit_message, fix `embeds` JSON serialize …

JellyWX:
…error in `create_interaction_response`

### Summary
Fixes 2 issues with the `embeds` parameter
* Serialization was failing due to the wrong variable being utilized in the `create_interaction_response` method
* Passing `embeds` to `edit_initial_response` was failing due to the list attempting to be checked if contained in a set, which raises an error remarking that lists are unhashable

nox passes ok on my system

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -258,7 +258,7 @@ class _RESTProvider(traits.RESTAware):
         return self._rest().proxy_settings
 
 
-_NONE_OR_UNDEFINED: typing.Final[typing.Set[undefined.UndefinedOr[None]]] = {None, undefined.UNDEFINED}
+_NONE_OR_UNDEFINED: typing.Final[typing.Tuple[None, undefined.UndefinedType]] = (None, undefined.UNDEFINED)
 
 
 class RESTApp(traits.ExecutorAware):
@@ -1239,7 +1239,7 @@ class RESTClientImpl(rest_api.RESTClient):
                 "meant to use 'collection' (singular) instead?"
             )
 
-        if embeds is not undefined.UNDEFINED and not isinstance(embeds, typing.Collection):
+        if embeds not in _NONE_OR_UNDEFINED and not isinstance(embeds, typing.Collection):
             raise TypeError(
                 "You passed a non collection to 'embeds', but this expects a collection. Maybe you meant to "
                 "use 'embed' (singular) instead?"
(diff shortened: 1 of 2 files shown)


hikari/impl/rest.py
```diff
@@ -1413,7 +1413,7 @@ async def _edit_message(  # noqa: C901 - Function too complex
                 "meant to use 'collection' (singular) instead?"
             )
 
-        if embeds not in _NONE_OR_UNDEFINED and not isinstance(embeds, typing.Collection):
+        if embeds is not undefined.UNDEFINED and not isinstance(embeds, typing.Collection):
```
davfsa, line 1416: This isn't correct, as passing None to embeds should not raise an error

hikari/impl/rest.py
```diff
@@ -3396,7 +3396,7 @@ async def create_interaction_response(
             embed_payloads: data_binding.JSONArray = []
             for embed in embeds:
                 serialized_embed, attachments = self._entity_factory.serialize_embed(embed)
-                embed_payloads.append(embed)
+                embed_payloads.append(serialized_embed)
```
davfsa, line 3399: Would you mind correcting the tests for this?

davfsa (changes requested):

hikari/impl/rest.py
davfsa: Nvm, seems like we don't have tests for this atm :sweat_smile:

JellyWX:
...that first comment was something that was used a few times; see if that solution is suitable. I'll look at tests now

hikari/impl/rest.py
```diff
@@ -258,7 +258,7 @@ def proxy_settings(self) -> config.ProxySettings:
         return self._rest().proxy_settings
 
 
-_NONE_OR_UNDEFINED: typing.Final[typing.Set[undefined.UndefinedOr[None]]] = {None, undefined.UNDEFINED}
+_NONE_OR_UNDEFINED: typing.Final[typing.Tuple[None, undefined.UndefinedType]] = (None, undefined.UNDEFINED)
```
davfsa, line 261: The reason this was a set initially is to be able to have fast checking, as that is what sets are used for. Not sure if tuple is ok will leave it up to @FasterSpeeding :P

JellyWX: The set requires the comparing item to be hashable, whereas the tuple doesn't. According to my IDE, this definition is unused other than the 3 usages here, which require a tuple rather than a set for hashability

davfsa:
While we wait for snabs comment, please have a read at https://github.com/hikari-py/hikari/blob/master/CONTRIBUTING.md#towncrier :)

hikari/impl/rest.py
FasterSpeeding: Looks fine to me

davfsa:
Added the changelog for you. @FasterSpeeding feel free to merge if you are happy with it :+1:

FasterSpeeding (approved):

PR #781, opened 2021-09-11
Add tests for some more modules

### LINKED ISSUE #750: Add missing tests to stickers
test coverage for all additions in this module?

_Originally posted by @FasterSpeeding in https://github.com/hikari-py/hikari/pull/736#discussion_r693385772_

davfsa:
### Checklist

### Related issues
Closes #750

### DIFF
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -496,9 +496,7 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
                         new_value = value_converter(new_value) if new_value is not None else None
                         old_value = value_converter(old_value) if old_value is not None else None
 
-                    elif _LOGGER.isEnabledFor(logging.DEBUG) and not isinstance(
-                        key, audit_log_models.AuditLogChangeKey
-                    ):
+                    elif not isinstance(key, audit_log_models.AuditLogChangeKey):
                         _LOGGER.debug("Unknown audit log change key found %r", key)
 
                     changes.append(audit_log_models.AuditLogChange(key=key, new_value=new_value, old_value=old_value))
@@ -2271,7 +2269,6 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
 
         if "reactions" in payload:
             reactions = [self._deserialize_message_reaction(reaction) for reaction in payload["reactions"]]
-
         else:
             reactions = []
 
@@ -2296,7 +2293,8 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
 
         if "sticker_items" in payload:
             stickers = [self.deserialize_partial_sticker(sticker) for sticker in payload["sticker_items"]]
-
+        elif "stickers" in payload:
+            stickers = [self.deserialize_partial_sticker(sticker) for sticker in payload["stickers"]]
         else:
             stickers = []
 
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -487,6 +487,11 @@ class _LiveAttributes:
         return self
 
 
+@attr.define(auto_exc=True, repr=False, weakref_slot=False)
+class _RetryRequest(RuntimeError):
+    ...
+
+
 class RESTClientImpl(rest_api.RESTClient):
     """Implementation of the V8-compatible Discord HTTP API.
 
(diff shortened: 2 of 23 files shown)


FasterSpeeding (dismissed):

hikari/impl/interaction_server.py
```diff
@@ -189,21 +189,23 @@ def is_alive(self) -> bool:
         return self._server is not None
 
     async def _fetch_public_key(self) -> ed25519.VerifierT:
+        if self._verify:
```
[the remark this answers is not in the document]
davfsa: Didn't notice, will revert :+1:

hikari/impl/entity_factory.py
```diff
@@ -496,9 +496,7 @@ def deserialize_audit_log(self, payload: data_binding.JSONObject) -> audit_log_m
                         new_value = value_converter(new_value) if new_value is not None else None
                         old_value = value_converter(old_value) if old_value is not None else None
 
-                    elif _LOGGER.isEnabledFor(logging.DEBUG) and not isinstance(
-                        key, audit_log_models.AuditLogChangeKey
-                    ):
+                    elif not isinstance(key, audit_log_models.AuditLogChangeKey):
```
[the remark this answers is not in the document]
davfsa: After testing, making the `isEnabledFor` checks are slower than just doing the isinstance check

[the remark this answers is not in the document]
FasterSpeeding: are you accounting for that fact that one isEnabledFor may be slow the the following will just be a dict lookup which is quicker than an instance check

tests/hikari/impl/test_interaction_server.py
```diff
@@ -123,6 +123,19 @@ def test_is_alive_property_when_active(self, mock_interaction_server):
 
         assert mock_interaction_server.is_alive is True
 
+    @pytest.mark.asyncio()
+    async def test___fetch_public_key_when_lock_is_None(self, mock_interaction_server, mock_rest_client):
+        mock_lock = mock.AsyncMock()
```
FasterSpeeding, line 128: Lock should already be None in the other tests, if you want to better test locking behaviour you might want to gather like 10 calls to an instance of interaction server which was built in that function's scope to ensure to ensure that the locking behaviour persists (where there'd be no need to explicitly set it to None )

davfsa: There are different test cases for this function and I don't think having them do the same call 10 times would be useful

FasterSpeeding: So testing that it doesn't overwrite the lock essentially making it useless isn't a useful test case? But testing that it sets the lock which if it didn't it'd either make a lock each time falling under what I just described but not under this test case or erroring on trying to use None as a lock.

davfsa: This is somehow tested by the use of different tests, but I get what you mean, will change it :+1:

hikari/impl/rest.py
```diff
@@ -789,25 +790,26 @@ async def _request(
...
-                can_re_auth = response.status == 401 and not (auth or no_auth or retried)
+                # Attempt to re-auth on UNAUTHORIZED if we are using a TokenStrategy
+                can_re_auth = response.status == 401 and not (auth or no_auth or re_authed)
                 if can_re_auth and isinstance(self._token, rest_api.TokenStrategy):
                     assert token is not None
                     self._token.invalidate(token)
                     token = await self._token.acquire(self)
                     headers[_AUTHORIZATION_HEADER] = token
-                    retried = True
+                    re_authed = True
```
FasterSpeeding, line 807: what does re_authed mean? this doesn't re-authenticate does it

hikari/impl/rest.py
```diff
@@ -2256,7 +2257,7 @@ async def fetch_available_sticker_packs(self) -> typing.Sequence[stickers.Sticke
         route = routes.GET_STICKER_PACKS.compile()
         response = await self._request(route, no_auth=True)
         assert isinstance(response, dict)
-        return [self._entity_factory.deserialize_sticker_pack(pack) for pack in response["sticker_packs"]]
+        return data_binding.cast_json_array(response["sticker_packs"], self._entity_factory.deserialize_sticker_pack)
```
FasterSpeeding, line 2260: isn't this just a tad bit unnecessary, kind of seems like this helper function introduces more lines of code than it saves

davfsa: Its for consistency with other functions that do this sort of this. Would it be best to remove `data_binding.cast_json_array` and just use a list cast?

davfsa: Removed :+1:

FasterSpeeding (approved):

PR #793, opened 2021-09-21
Add check for trailing whitespaces

davfsa:


pipelines/format.nox.py
```diff
@@ -97,6 +100,10 @@ def remove_trailing_whitespaces_for_file(file) -> bool:
         if lines == new_lines:
             return False
 
+        if check_only:
+            print("Trailing whitespaces found in", file)
+            return True
+
         print("Removing trailing whitespaces present in", file)
```
davfsa, line 107: Change these to `sessions.log` to prevent them not printing when expected (which can be seen in the github CI logs)

FasterSpeeding (approved):

PR #797, opened 2021-09-26
Document more newly documented rest codes

davfsa:


FasterSpeeding:
Legit question though what is the point of this enum? It doesn't document which endpoints return what codes and it doesn't cover anywhere near all of the error rest codes and what is the use case for matching error rest codes

davfsa:
> Legit question though what is the point of this enum? It doesn't document which endpoints return what codes and it doesn't cover anywhere near all of the error rest codes and what is the use case for matching error rest codes

It may very well be removed tbh, it is only used to provide a bit more info on rest errors, but ig we can just have the int and forget about it

FasterSpeeding:
> it is only used to provide a bit more info on rest errors

Here's the thing about that, the message for these codes is just in the error payload so if you want a more user friendly way to present the error then just do `error.message` for something to show to a user

davfsa:
Will play around with this and open a followup pr that removes the enum

PR #800, opened 2021-09-26
Add new channel_types field to CommandOption

davfsa:


hikari/commands.py
```diff
@@ -136,6 +137,11 @@ class CommandOption:
     options: typing.Optional[typing.Sequence[CommandOption]] = attr.field(default=None, repr=False)
     """Sequence of up to (and including) 25 of the options for this command option."""
 
+    channel_types: typing.Optional[typing.Sequence[typing.Union[channels.ChannelType, int]]] = attr.field(
+        default=None, repr=False
+    )
+    """If a channel type option, the valid channel types that this option will accept."""
```
FasterSpeeding, line 143: `valid channel types` you shouldn't be expecting to get invalid channel types here so thats redundant

davfsa:
```suggestion
    """If a channel type option, the channel types that this option will accept."""
```

hikari/commands.py
```diff
@@ -136,6 +137,11 @@ class CommandOption:
     options: typing.Optional[typing.Sequence[CommandOption]] = attr.field(default=None, repr=False)
     """Sequence of up to (and including) 25 of the options for this command option."""
 
+    channel_types: typing.Optional[typing.Sequence[typing.Union[channels.ChannelType, int]]] = attr.field(
+        default=None, repr=False
+    )
+    """If a channel type option, the channel types that this option will accept."""
```
FasterSpeeding, line 143: its also worth noting that if this is None for a channel type option then it accepts all option types

davfsa:
```suggestion
    """The channel types that this option will accept.
    
    If `builtins.None`, then all channel types will be accepted.
    """
```

FasterSpeeding (dismissed):

FasterSpeeding (approved):

PR #810, opened 2021-10-02
Add missing `old_message` to MessageDeleteEvent

### LINKED ISSUE #766: Add missing `old_x` fields in events
There is currently a bit of inconsistency in the `old_x` fields in the update and delete events for objects we cache

Examples that I have seen are (there might be more):
- *MessageDeleteEvent
- ~~*ChannelDeleteEvent (no need, discord send the info for this one)~~
- GuildLeaveEvent (solved by #806)

davfsa:
### Checklist

*Been a long day, will add tests and log fragment in further days. I opened to receive feedback :)*

### Related issues
Closes #766

### DIFF
--- a/hikari/events/message_events.py
+++ b/hikari/events/message_events.py
@@ -449,6 +449,17 @@ class MessageUpdateEvent(MessageEvent, abc.ABC):
 
     @property
     @abc.abstractmethod
+    def old_message(self) -> typing.Optional[messages.PartialMessage]:
+        """Object of the old message stored in cache.
+
+        Returns
+        -------
+        typing.Optional[hikari.snowflakes.Snowflake]
+            The old message, if in cache, else `builtins.None`.
+        """
+
+    @property
+    @abc.abstractmethod
     def message(self) -> messages.PartialMessage:
         """Partial message that was sent in the event.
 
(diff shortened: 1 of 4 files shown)


hikari/events/message_events.py
```diff
@@ -606,9 +617,40 @@ class MessageDeleteEvent(MessageEvent, abc.ABC):
...
+        event. For all other purposes, this is the same as running
+        `self.old_messages.get(next(iter(self.message_ids)))`.
+
+        Returns
+        -------
+        typing.Optional[hikari.snowflakes.Snowflake]
+            The first deleted message, if in cache, else `builtins.None`.
+        """
+        try:
+            return self.old_messages.get(next(iter(self.message_ids)))
```
FasterSpeeding, line 634: can't you just do `next(iter(self.old_messages.values()))` here?

davfsa: There is no guarantee that the first value will be the first message, as it might be missing from the cache due to Discord not sending the message create or some other reason

FasterSpeeding: In what situation when handling a bulk delete will someone only care about the first message which was deleted

PR #821, opened 2021-10-06
Add `allow` and `deny` to PermissionOverwrite's repr

davfsa:
### Checklist


FasterSpeeding:
Calling them allow and deny instead of white and black in order to avoid getting cancelled, nice thinking there

PR #829, opened 2021-10-07
Ensure hikari.File path points to an existing file

### LINKED ISSUE #826: Better handling of non-existent attachment attachment paths
### Summary
Consider checking if file attachments exist before passing them through to aiohttp.

### Problem
The current error for a file not being found is the following
```
aiohttp.client_exceptions.ClientOSError: [Errno 13] Can not write request body for
```
This is unintuitive and directs away from the actual problem of the file not existing and if we had an explicit check before the file was handed over to aiohttp we could improve the user experience when dealing with this issue.

### Ideal implementation
Check if file attachments exist before passing them through to aiohttp.

### Checklist

- [ ] I have searched the issue tracker and have made sure it's not a duplicate.

davfsa:
### Summary
Ensure hikari.File path points to an existing file

### Checklist

### Related issues
Closes #826

### DIFF
--- a/changes/829.bugfix.md
+++ b/changes/829.bugfix.md
@@ -1,0 +1,1 @@
+Ensure `hikari.File` paths point to a valid file
--- a/hikari/files.py
+++ b/hikari/files.py
@@ -48,6 +48,7 @@ import abc
 import asyncio
 import base64
 import concurrent.futures
+import errno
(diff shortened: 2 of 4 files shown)


davfsa:
Closing this, as the calls do seem to be blocking

Issue #808, opened 2021-10-01
Switch to `pdoc` and update theme


davfsa:
We should switch to [`pdoc`](https://github.com/mitmproxy/pdoc) with a new template to make the documentation easier to read and navigate. This should be done before #410 as the way pdoc outputs the HTML is a nicer and more compact format to allow for different version documentation.

Due to the huge scope of this, I will create a branch called `feature/pdoc` where progress will be made, which will later be merged into `master` once everything is ready.

### pdoc
- [x] Ensure everything renders correctly
  - `pdoc` uses a more strict numpy docstring style, so some things we are currently doing wont render nicecly anymore
    - `See also` section should only consist of links, no other text
- [x] Properties docstrings should not include a `Returns` section
- [x] Fix all the warnings printed (https://github.com/mitmproxy/pdoc/issues/307) [Python bug, so nothing we can do about it. Hopefully will be fixed in the future]
- [x] Ensure in-project linking works (relates to above)
- [x] `object.inv` output

### Theme
Still not decided. I have recently become a fan of rust projects documentation, as well as the nice usage of the sidebar in rustbook, so will probably take inspiration from there

#

*Both lists are not complete and can be edited to add more points as more issues are found or changes are to be made*

FasterSpeeding:
I'ma just drop this script here cause this is a pretty nice way to get an easy to consume search json index on v8 pdoc https://gist.github.com/FasterSpeeding/19a6d3f44cdd0a1f3b2437a8c5eef07a (the result is the same as v7's search.json)

davfsa:
Technically closed by #856, but will keep open till the branch is merged into master

PR #856, opened 2021-10-19
Switch to pdoc

davfsa:
### Summary
Switch to pdoc with new theme style and better mobile compatibility

This branch is not 100% functional to be merged into master, as it relies a bit on having multiple documentation versions. The idea is to have this merged into a separate branch and then it can be worked on from there, as this is already really big

### Related issues
Solves all points in #808

<details>
<summary>Screenshots</summary>

#### Desktop
image
image

#### Mobile
image
Screen Shot 2021-10-19 at 22 07 28
Screen Shot 2021-10-19 at 22 07 51
</details>


davfsa:
Fixed failing pipelines and added some screenshots to the initial comment, in case anybody is interested :)

hikari/impl/bot.py
```diff
@@ -213,6 +212,15 @@ class GatewayBot(traits.GatewayBotAware):
     ```
     """
 
+    shards: typing.Mapping[int, gateway_shard.GatewayShard]
+    """Mapping of shards in this application instance.
+
+    Each shard ID is mapped to the corresponding shard instance.
+
+    If the application has not started, it is acceptable to assume the
```
FasterSpeeding, line 220: Should this doc string not be being inherited from traits.ShardAware

hikari/impl/event_manager.py
```diff
@@ -88,7 +88,7 @@ def __init__(
         super().__init__(event_factory=event_factory, intents=intents)
 
     async def on_ready(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
-        """See https://discord.com/developers/docs/topics/gateway#ready for more info."""
+        """See <<https://discord.com/developers/docs/topics/gateway#ready>> for more info."""
```
FasterSpeeding, line 91: accidental double brackets

pipelines/utils.nox.py
```diff
@@ -25,29 +25,38 @@
...
+
+FILES_TO_DELETE = [
+    ".coverage",
     "package-lock.json",
 ]
 
 
 @nox.session(reuse_venv=False)
-def purge(_: nox.Session) -> None:
+def purge(session: nox.Session) -> None:
```
FasterSpeeding, line 46: this seems a bit out of scope

hikari/internal/deprecation.py
```diff
@@ -49,60 +47,25 @@ def warn_deprecated(
...
         The object that is deprecated.
+    reason: str
+        The reason for the deprecation.
 
     Other Parameters
     ----------------
-    version: typing.Optional[str]
-        If specified, the version it will be removed in.
-    alternative: typing.Optional[str]
-        If specified, the alternative to use.
```
FasterSpeeding, line 58: why were these args removed and what do these changes have to do with pdoc

hikari/internal/deprecation.py
```diff
@@ -49,60 +47,25 @@ def warn_deprecated(
...
+        action = ("Call to", "function/method")
 
-    if alternative is not None:
-        message += f" You can use '{alternative}' instead."
-
-    warnings.warn(message, category=DeprecationWarning, stacklevel=stack_level)
-
-
-def deprecated(
-    version: typing.Optional[str] = None, alternative: typing.Optional[str] = None
```
FasterSpeeding, line 75: why was this removed

hikari/impl/bot.py
davfsa: It doesnt seem to be, most probably since it becomes an attribute instead of a property

pipelines/utils.nox.py
davfsa: I know, but it was annoying me quite a bit when it didnt work correctly

hikari/internal/deprecation.py
davfsa: _This was explained on Discord, but I will put it here for context_

The decorator seemed to completely discard the docstring, and I couldn't get it to work, which is why I make it a function only instead.

Will try to get this to work when I have time

FasterSpeeding: What about the args though

pipelines/utils.nox.py
FasterSpeeding: wouldn't it be less annoying if you fixed it sooner by putting it in a pr which can be immediately pred to master instead of tagging it along with an unrelated massive change

hikari/impl/bot.py
FasterSpeeding: Well that shouldn't be an issue if you don't move where its being type hinting it for no reason but also cannot reproduce that behaviour

davfsa: I am still able to reproduce it unfortunately: 
image

I got absolutely no idea why this happens

FasterSpeeding: It shouldn't be an issue if its being type hinted in the `__init__` though since pdoc just doesn't introspect those

davfsa: Nope, even if I take the typehint out it still loses the docstring for some reason. I really havent been able to figure out

FasterSpeeding (approved):

pipelines/config.py
```diff
@@ -95,6 +95,5 @@
     *(f for f in _os.listdir(".") if _os.path.isfile(f) and f.endswith(REFORMATTING_FILE_EXTS)),
     ".github",
     "pages",
-    "docs",
```
davfsa, line 98: Re-add this and use `--ignore-words-list="nd"` when running codespell

FasterSpeeding: Btw --ignore-words-list doesn't exist rn, that's in this pr https://github.com/codespell-project/codespell/pull/2040. You'd probably want to use --ignore-words and make a file for it rn

FasterSpeeding:
was this never merged? I thought this was a new pr lol

hikari/api/rest.py
```diff
@@ -5097,21 +5052,18 @@ async def edit_my_nick(
     ) -> None:
         """Edit the associated token's member nick.
 
-        .. deprecated:: 2.0.0.dev104
-            Use `RESTClient.edit_my_member`'s `nick` argument instead.
```
FasterSpeeding, line 5101: Why was this removed?

hikari/internal/deprecation.py
```diff
@@ -35,72 +35,63 @@
...
-        )
-        obj.__doc__ = doc
+        old_doc = inspect.getdoc(obj)
+
+        # If the docstring is inherited we can assume that the deprecation warning was already added there
+        if old_doc:
+            first_line_end = old_doc.index("\n")
+            obj.__doc__ = (
+                old_doc[:first_line_end]
+                + f"\n\n.. deprecated:: {version}\n    {additional_information}"
```
FasterSpeeding, line 88: Wouldn't this lead to inconsistencies as to when the deprecation warnings are actually in the docs since it'll have to be done inline for cases where specific arguments were deprecated?

davfsa: This is aimed for fully deprecated functions/methods/classes. When deprecating only an argument we can add the warning to the docstring and then raise the warning with `warn_deprecated`

FasterSpeeding: Which makes the docs inconsistent for anybody reading through the code right?

davfsa: Not really, since the decorator would be there. The deprecation decorator could also be extended to include arguments and detect when a specific argument is passed

hikari/internal/deprecation.py
```diff
@@ -35,72 +35,63 @@
...
-    version: typing.Optional[str]
-        If specified, the version it will be removed in.
-    alternative: typing.Optional[str]
-        If specified, the alternative to use.
     stack_level: int
         The stack level for the warning. Defaults to `3`.
     """
-    if inspect.isclass(obj) or inspect.isfunction(obj):
+    if inspect.isclass(obj):
+        action = ("Instantiation of", "class")
```
FasterSpeeding, line 54: Also this isn't fit for classes as it'd break anything inheriting from said class so this seems a bit redundant

hikari/internal/deprecation.py
```diff
@@ -35,72 +35,63 @@
...
-        doc = inspect.getdoc(obj) or ""
-        doc += (
-            "\n"
-            "!!! warning\n"
-            f"    This {type_str} is deprecated and will be removed in {version_str}.\n"
-            f"    {alternative_str}\n"
-        )
-        obj.__doc__ = doc
+        old_doc = inspect.getdoc(obj)
+
```
FasterSpeeding, line 82: can we get a check here to error if this ever decorates a class

FasterSpeeding (approved):

davfsa:
Finally done with all points from #808, so it can be considered closed and solved for the most part. Will be actually closed once this is merged into master :)

PR #876, opened 2021-11-04
Allow for more relaxed arg type fallback in some overloaded signatures

### LINKED ISSUE #875: Invalid type signature for ActionRowBuilder.add_button
### Steps to reproduce
- Attempt to add a button to the builder passing in arguments with types ``ButtonStyle, str``
- Run mypy
- Mypy throws an error that no overload variant matches the types passed in

### Expected Result
Mypy should not throw an error as arguments of those types are valid.

### Actual Result
```
nox > python -m mypy lightbulb
lightbulb/utils/nav.py: note: In member "build" of class "ComponentButton":
lightbulb/utils/nav.py:129:15: error: No overload variant of "add_button" of "ActionRowBuilder" matches argument types "ButtonStyle", "str"  [call-overload]
            btn = container.add_button(self.style, self.custom_id)
                  ^
lightbulb/utils/nav.py:129:15: note: Possible overload variants:
lightbulb/utils/nav.py:129:15: note:     def add_button(self, Union[Literal[ButtonStyle.PRIMARY], Literal[1], Literal[ButtonStyle.SECONDARY], Literal[2], Literal[ButtonStyle.SUCCESS], Literal[3], Literal[ButtonStyle.DANGER], Literal[4]], str) -> InteractiveButtonBuilder[ActionRowBuilder]
lightbulb/utils/nav.py:129:15: note:     def add_button(self, Union[Literal[ButtonStyle.LINK], Literal[5]], str) -> LinkButtonBuilder[ActionRowBuilder]
Found 1 error in 1 file (checked 31 source files)
nox > Command python -m mypy lightbulb failed with exit code 1
nox > Session mypy failed.
```

### System info
```
hikari (2.0.0.dev103) [dc0f905a]
located at /Users/thommo/Development/hikari-lightbulb/venv/lib/python3.9/site-packages/hikari
CPython 3.9.6 Clang 12.0.5 (clang-1205.0.22.9)
Darwin Thommos-MacBook-Pro.local 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:24 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T8101 arm64 arm
```

### Checklist

FasterSpeeding:
### Summary
Allow for more relaxed arg type fallback in some overloaded signatures

### Checklist

### Related issues
Closes #875

### DIFF
--- a/hikari/impl/special_endpoints.py
+++ b/hikari/impl/special_endpoints.py
@@ -1270,12 +1270,18 @@ class ActionRowBuilder(special_endpoints.ActionRowBuilder):
     @typing.overload
     def add_button(
         self: _ActionRowBuilderT, style: typing.Union[int, messages.ButtonStyle], url_or_custom_id: str, /
-    ) -> special_endpoints.ButtonBuilder[_ActionRowBuilderT]:
+    ) -> typing.Union[
+        special_endpoints.LinkButtonBuilder[_ActionRowBuilderT],
+        special_endpoints.InteractiveButtonBuilder[_ActionRowBuilderT],
+    ]:
         ...
 
     def add_button(
         self: _ActionRowBuilderT, style: typing.Union[int, messages.ButtonStyle], url_or_custom_id: str, /
-    ) -> special_endpoints.ButtonBuilder[_ActionRowBuilderT]:
+    ) -> typing.Union[
+        special_endpoints.LinkButtonBuilder[_ActionRowBuilderT],
+        special_endpoints.InteractiveButtonBuilder[_ActionRowBuilderT],
+    ]:
         self._assert_can_add_type(messages.ComponentType.BUTTON)
         if style in messages.InteractiveButtonTypes:
             return InteractiveButtonBuilder(container=self, style=style, custom_id=url_or_custom_id)
--- a/hikari/impl/special_endpoints.py
+++ b/hikari/impl/special_endpoints.py
@@ -1261,10 +1261,16 @@ class ActionRowBuilder(special_endpoints.ActionRowBuilder):
     @typing.overload
     def add_button(
         self: _ActionRowBuilderT,
-        style: typing.Union[typing.Literal[messages.ButtonStyle.LINK], typing.Literal[5]],
+        style: typing.Literal[messages.ButtonStyle.LINK, 5],
         url: str,
         /,
     ) -> special_endpoints.LinkButtonBuilder[_ActionRowBuilderT]:
+        ...
+
+    @typing.overload
+    def add_button(
+        self: _ActionRowBuilderT, style: typing.Union[int, messages.ButtonStyle], url_or_custom_id: str, /
+    ) -> special_endpoints.ButtonBuilder[_ActionRowBuilderT]:
         ...
 
     def add_button(
--- a/changes/876.bugfix.md
+++ b/changes/876.bugfix.md
@@ -1,0 +1,1 @@
+Relaxed typing of methods with union entry specific specialisations through overloads.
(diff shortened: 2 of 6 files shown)


hikari/impl/special_endpoints.py
```diff
@@ -1261,12 +1261,18 @@ def add_button(
     @typing.overload
     def add_button(
         self: _ActionRowBuilderT,
-        style: typing.Union[typing.Literal[messages.ButtonStyle.LINK], typing.Literal[5]],
+        style: typing.Literal[messages.ButtonStyle.LINK, 5],
```
davfsa, line 1264: Do we really want to use Literals at all? It makes it for issues like what @tandemdude pointed out annoying to fix.

FasterSpeeding: what about this is annoying to fix

FasterSpeeding: Also do you really want add_button to be returning a union in cases where it can easily be inferred from the passed type that it should be returning a specific type from said union

davfsa (dismissed):

hikari/impl/special_endpoints.py
davfsa: For this specific `add_button` case we should probably have a separate function that sets the specific style to 5. Reason for this being the following example:
```py
import hikari
from hikari.impl import special_endpoints

style = hikari.ButtonStyle.LINK

reveal_type(style)

builder = special_endpoints.ActionRowBuilder().add_button(style, "test")

reveal_type(builder)
```

You will get these logs:
```
test.py:6: note: Revealed type is "hikari.messages.ButtonStyle"
test.py:10: note: Revealed type is "Union[hikari.api.special_endpoints.LinkButtonBuilder[hikari.impl.special_endpoints.ActionRowBuilder*], hikari.api.special_endpoints.InteractiveButtonBuilder[hikari.impl.special_endpoints.ActionRowBuilder*]]"

```
Which still wouldnt fix the issue in hand.

Maybe it would be worth revisiting how how we currently do component builder and switch to a more "embed like" pattern?

FasterSpeeding: > Which still wouldnt fix the issue in hand. 

How does this not fix the problem of you not being able to pass a more relaxed type to this function? Also what you just described wouldn't fix the issue as you'd have these functions

```py
    def add_interactive_button(
        self: _ActionRowBuilderT, style: messages.InteractiveButtonTypesT, custom_id: str, /
    ) -> special_endpoints.InteractiveButtonBuilder[_ActionRowBuilderT]:
        ...

    def add_url_button(
        self: _ActionRowBuilderT,
        style: typing.Literal[messages.ButtonStyle.LINK, 5],
        url: str,
        /,
    ) -> special_endpoints.LinkButtonBuilder[_ActionRowBuilderT]:
        ...

```

Neither of which you could pass something typed as `hikari.ButtonStyle` to, also what you just showed in your example is literally intended cause you passed something typed as hikari.ButtonStyle rather than a literal type.

FasterSpeeding: > switch to a more "embed like" pattern?

Ok so you know how you don't need an emoji object here, just an id or name, how would you deal with merging that with a data model class where you get a full emoji object without just making it less relaxed for no good reason.

And another question, isn't the embed approach just kinda bad but left there cause its too late to switch to a proper strict builder/dto pattern where's there's like actual separation. Like that pattern's already inferring that you can just safely relay received embeds which you can't cause there's like a ton of received fields you cannot send and it also breaks away from the lib standard that DTO models are considered immutable and probably also messes with the cache's reliability in the process because of that.

davfsa (approved):

PR #886, opened 2021-11-05
Add nox to dev requirements

### LINKED ISSUE #883: Add nox to dev-requirements
### Summary
It would make contributing to hikari a smoother experience if nox was actually installed as part of the dev reuqirements. It's a small change, but there's no reason it shouldn't be there (at least to my knowledge).

### Problem
People have to install nox separately. If they don't know what's going on when running nox, they may submit untested contributions.

### Ideal implementation
Something in the dev-requirements.txt. ~~Maybe also add a pytest patch in there too for being able to run tests on 3.11.~~

### Checklist

parafoxia:
### Summary
Add nox to dev requirements.

### Checklist
- [ ] ~~I have made unittests according to the code I have added/modified/deleted.~~ Not necessary

### Related issues
Closes #883.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,6 +2,9 @@
 # UNIT TESTING #
 ################
 
+# nox to run test sessions locally.
+nox==2021.10.1
+
 # Mocks (stdlib ones change between versions of Python, so are not consistent in behaviour like the backport is).
 mock==4.0.3
 
@@ -36,9 +39,9 @@ mypy==0.910
 
 safety~=1.10.3
 
-#############
+##############
 # FORMATTING #
-#############
+##############
 
 black==21.10b0
 isort==5.10.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,9 +1,8 @@
+nox==2021.10.1
+
 ################
 # UNIT TESTING #
 ################
-
-# nox to run test sessions locally.
-nox==2021.10.1
 
 # Mocks (stdlib ones change between versions of Python, so are not consistent in behaviour like the backport is).
 mock==4.0.3


dev-requirements.txt
```diff
@@ -2,6 +2,9 @@
 # UNIT TESTING #
 ################
 
+# nox to run test sessions locally.
+nox==2021.10.1
+
```
davfsa, line 7: This could be moved out of the unit tests section, since its used for so much more than that. Also, the comment is irrelevant, they were only used to comment on things like why we install `mock` even tho it is a standard lib

dev-requirements.txt
```diff
@@ -1,10 +1,13 @@
+###################
+# SESSION TESTING #
+###################
+
+nox==2021.10.1
```
davfsa, line 5:
```suggestion
nox==2021.10.1
```

Could just be this tbh :P

davfsa (approved):

FasterSpeeding:
Random thought, should this effect the pipelines at all, seems weird to be now hard linking a specific nox ver while having the pipelines just call `pip install nox`

davfsa:
> 
> 
> Random thought, should this effect the pipelines at all, seems weird to be now hard linking a specific nox ver while having the pipelines just call `pip install nox`

I was actually just thinking about this, which is why I removed it from the auto merge. Might be worth splitting the dev-requirements file into 2, one for the jobs and one for people to use generally. This is kinda the reason why it was never included in the file

*Might be a better way to handle this, but this is what first came to mind*

parafoxia:
In my projects I have the development dependencies (-dev, anything the user might use), and then a separate file for anything nox wants (-nox, stuff only used for nox. pytest is in there, for example). It works quite well for me, idk how efficient my specific use of it is but yeah may be something worth thinking about.

FasterSpeeding:
> and then a separate file for anything nox wants (-nox, stuff only used for nox. pytest is in there, for example)

I have something a bit different, for me my "nox-dependencies" are just requirements needed to run nox but doesn't include stuff nox itself would install in the sessions it runs which in this case would just be nox.

FasterSpeeding:
What's happening with this pr?

davfsa:
> 
> 
> What's happening with this pr?

I have been thinking on it for a while, but I dont know what the best course of action for this would be, sorry

Jonxslays:
Would having a set version of nox be an issue? It would require changes in the pipelines but i don't think there is really a downside to having it pinned. This same thing could be applicable to towncrier as well. I would assume that if i install the dev requirements on a project i'll be ready to start contributing with the necessary tools installed.

davfsa:
Closing in favor of #915

PR #887, opened 2021-11-05
Consistency between the usage of sequence and collection in typing

davfsa:
### Summary
* Fix checking against `typing.Collection` where the type was `typing.Sequence`
* Change uses of `typing.Collection` to `typing.Sequence`
   * Reason behind this change is that `typing.Sequence` reflects more what we are returning type wise (a read-only sequence)

### Checklist

### Related issues
None

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1238,19 +1238,19 @@ class RESTClientImpl(rest_api.RESTClient):
         if not undefined.any_undefined(embed, embeds):
             raise ValueError("You may only specify one of 'embed' or 'embeds', not both")
 
-        if attachments is not undefined.UNDEFINED and not isinstance(attachments, typing.Collection):
+        if attachments is not undefined.UNDEFINED and not isinstance(attachments, typing.Sequence):
             raise TypeError(
                 "You passed a non-collection to 'attachments', but this expects a collection. Maybe you meant to "
                 "use 'attachment' (singular) instead?"
             )
 
-        if components is not undefined.UNDEFINED and not isinstance(components, typing.Collection):
+        if components is not undefined.UNDEFINED and not isinstance(components, typing.Sequence):
             raise TypeError(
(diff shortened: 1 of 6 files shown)


hikari/api/event_manager.py
```diff
@@ -327,7 +327,7 @@ def get_listeners(
         /,
         *,
         polymorphic: bool = True,
-    ) -> typing.Collection[CallbackT[EventT_co]]:
+    ) -> typing.Sequence[CallbackT[EventT_co]]:
```
FasterSpeeding, line 330: There's no guarantee this will be sorted as an ordered dataset for all impls though

hikari/api/special_endpoints.py
```diff
@@ -392,7 +392,7 @@ def add_text_channel(
         rate_limit_per_user: undefined.UndefinedOr[time.Intervalish] = undefined.UNDEFINED,
         position: undefined.UndefinedOr[int] = undefined.UNDEFINED,
         permission_overwrites: undefined.UndefinedOr[
-            typing.Collection[channels.PermissionOverwrite]
+            typing.Sequence[channels.PermissionOverwrite]
```
FasterSpeeding, line 395: Why is order important here?

FasterSpeeding:
> Fix checking against typing.Collection where the type was typing.Sequence

How is this a fix? A sequence is a collection so I don't see what's wrong there.

hikari/impl/rest.py
```diff
@@ -1238,19 +1238,19 @@ async def _create_message(
         if not undefined.any_undefined(embed, embeds):
             raise ValueError("You may only specify one of 'embed' or 'embeds', not both")
 
-        if attachments is not undefined.UNDEFINED and not isinstance(attachments, typing.Collection):
+        if attachments is not undefined.UNDEFINED and not isinstance(attachments, typing.Sequence):
```
FasterSpeeding, line 1241: quick question but why does this matter

davfsa:
I guess it doesn't really matter

PR #891, opened 2021-11-06
Added VoiceEvent method get_guild

NobodyForYouCreator:
### Summary
In order to optimize all the events, it would be a good idea to add every event get_guild to make user's work as optimized as it can be

### Checklist

### DIFF
--- a/changes/891.feature.md
+++ b/changes/891.feature.md
@@ -1,0 +1,3 @@
+To make it possible to use events with decorators, where get_guild is required
+
+It's the class of event, where there's no feature as in others events like get_guild... so...
--- a/hikari/events/voice_events.py
+++ b/hikari/events/voice_events.py
@@ -44,6 +44,7 @@ if typing.TYPE_CHECKING:
     from hikari import snowflakes
     from hikari import traits
     from hikari import voices
+    from hikari import guilds
     from hikari.api import shard as gateway_shard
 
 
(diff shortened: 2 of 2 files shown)


NobodyForYouCreator:
Created file in changelog

hikari/events/voice_events.py
```diff
@@ -62,6 +63,21 @@ def guild_id(self) -> snowflakes.Snowflake:
         hikari.snowflakes.Snowflake
             The guild ID of the guild this event relates to.
         """
+    
+    def get_guild(self) -> typing.Optional[guilds.GatewayGuild]:
```
FasterSpeeding, line 67: Could we get two test cases for this? It should be pretty simple to do, you can pretty much just copy something like these two tests https://github.com/hikari-py/hikari/blob/fff9bca164271dad51b8a62aedafd64f52ed4990/tests/hikari/events/test_message_events.py#L207-L215

FasterSpeeding, line 67: If you want to, you could also add a similar fetch_guild method here

changes/891.feature.md
```diff
@@ -0,0 +1,3 @@
+To make it possible to use events with decorators, where get_guild is required
```
FasterSpeeding, line 1: could we just reword this to something like "Added a get_guild (and possibly fetch_guild) shortcut method(s) to the voice events"

hikari/events/voice_events.py
FasterSpeeding: I'll do the fetch_guild method in another pr

davfsa:
Closing due to inactivity and I dont think we will keep these methods around for long anyways

PR #892, opened 2021-11-07
Fix some cache bugs and ensure member is always returned on delete calls

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/cache.py
+++ b/hikari/impl/cache.py
@@ -799,7 +799,7 @@ class CacheImpl(cache.MutableCache):
         *,
         decrement: typing.Optional[int] = None,
         deleting: bool = False,
-    ) -> typing.Optional[cache_utility.RefCell[cache_utility.MemberData]]:
+    ) -> None:
         if deleting:
             member.object.has_been_deleted = True
 
@@ -808,10 +808,10 @@ class CacheImpl(cache.MutableCache):
 
         user_id = member.object.user.object.id
         if not guild_record.members or user_id not in guild_record.members:
-            return None
+            return
 
         if not self._can_remove_member(member):
-            return None
+            return
 
         del guild_record.members[user_id]
         self._garbage_collect_user(member.object.user, decrement=1)
@@ -819,8 +819,6 @@ class CacheImpl(cache.MutableCache):
         if not guild_record.members:
             guild_record.members = None
             self._remove_guild_record_if_empty(member.object.guild_id, guild_record)
-
-        return member
 
     def clear_members(
         self,
@@ -843,9 +841,10 @@ class CacheImpl(cache.MutableCache):
             return cache_utility.EmptyCacheView()
 
         cached_members = guild_record.members.freeze()
-        members_gen = (self._garbage_collect_member(guild_record, m, deleting=True) for m in cached_members.values())
-        # _garbage_collect_member will only return the member data object if they could be removed, else None.
-        cached_members = {member.object.user.object.id: member for member in members_gen if member}
+
+        for m in cached_members.values():
+            self._garbage_collect_member(guild_record, m, deleting=True)
+
         self._remove_guild_record_if_empty(guild_id, guild_record)
         return cache_utility.CacheMappingView(cached_members, builder=self._build_member)  # type: ignore[type-var]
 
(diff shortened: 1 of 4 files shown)


hikari/impl/cache.py
```diff
@@ -820,8 +820,6 @@ def _garbage_collect_member(
             guild_record.members = None
             self._remove_guild_record_if_empty(member.object.guild_id, guild_record)
 
-        return member
```
FasterSpeeding, line 823: You should just change this to return the object if it wasn't originally marked as has_been_deleted

hikari/impl/cache.py
```diff
@@ -1348,11 +1343,12 @@ def delete_voice_state(
         if not voice_state_data:
             return None
 
+        self._garbage_collect_member(guild_record, voice_state_data.member, decrement=1)
+
         if not guild_record.voice_states:
             guild_record.voice_states = None
+            self._remove_guild_record_if_empty(guild_id, guild_record)
```
FasterSpeeding, line 1351: Why was his moved

hikari/impl/cache.py
```diff
@@ -1295,12 +1289,13 @@ def clear_voice_states_for_channel(
         if not guild_record or not guild_record.voice_states:
             return cache_utility.EmptyCacheView()
 
-        cached_voice_states = {}
+        cached_voice_states = dict(
+            filter(lambda item: item[1].channel_id == channel_id, guild_record.voice_states.items())
+        )
```
FasterSpeeding, line 1294: This seems really unnecessary, you can just copy the guild's voice stats before iterating over them like everywhere else in the cache stores

hikari/impl/cache.py
```diff
@@ -1459,12 +1455,13 @@ def _garbage_collect_message(
         *,
         decrement: typing.Optional[int] = None,
         override_ref: bool = False,
-    ) -> typing.Optional[cache_utility.RefCell[cache_utility.MessageData]]:
+    ) -> bool:
```
FasterSpeeding, line 1458: This change seems unnecessary and breaks the standard pattern for these methods

hikari/impl/cache.py
```diff
@@ -1485,7 +1482,7 @@ def _garbage_collect_message(
         if message.object.id in self._referenced_messages:
```
FasterSpeeding, line 1482: This should return None/False/whatever you feel like if it was already marked as deleted

hikari/impl/cache.py
davfsa: The other methods currently only return `None`. I changed this since I dont think it makes sense to return the exact same thing that was sent in, considering how we are using this function atm

hikari/impl/cache.py
davfsa: Changed :+1:

hikari/impl/cache.py
davfsa: It removes an unnecessary call to `_remove_guild_record_if_empty` when not needed. Garbage collecting the member will have this exact check, so it just makes more sense to have it here instead, since only under this case it will make sense to call it

hikari/impl/cache.py
davfsa: The reason I changed this to not return anything is because there is no need, as the information we return in the same as the one we take in. The cases we use this for atm don't even need what we return

hikari/impl/cache.py
davfsa: The uses for this function stay the same, so either its a bug I didnt fix or I did in fact miss something. I dont get the point for an extra return tho

hikari/impl/cache.py
FasterSpeeding: Clear and delete methods should only be returning objects which were actually marked as deleted not objects which are being kept alive by references, the way you've changed the semantics still leads to erroneous behaviour and simply changing the behaviour of the garbage collect method will instruct both the delete and clear methods on what they should be saying was deleted and avoids a ton of other code changes

hikari/impl/cache.py
FasterSpeeding: The methods which were previously only returning None are doing this because they have no relevant delete and clear methods, they are not the same case

hikari/impl/cache.py
FasterSpeeding: Clear and delete methods should only be returning objects which were actually marked as deleted not objects which are being kept alive by references

hikari/impl/cache.py
FasterSpeeding: > think it makes sense to return the exact same thing that was sent in,

Look at how clear members was using it before

davfsa:
Will let @norinorin take care of this pr

davfsa:
Sorry, wrong pr :sweat_smile:

PR #903, opened 2021-11-12
Improve filtering by using per event bitmasks

davfsa:
### Checklist

### Related issues
Continuation of #636

### DIFF
--- a/hikari/api/entity_factory.py
+++ b/hikari/api/entity_factory.py
@@ -42,7 +42,6 @@ if typing.TYPE_CHECKING:
     from hikari import invites as invite_models
     from hikari import messages as message_models
     from hikari import presences as presence_models
-    from hikari import scheduled_events as scheduled_events_models
     from hikari import sessions as gateway_models
     from hikari import snowflakes
     from hikari import stickers as sticker_models
@@ -960,72 +959,6 @@ class EntityFactory(abc.ABC):
     ######################
 
     @abc.abstractmethod
-    def deserialize_slash_command(
-        self,
-        payload: data_binding.JSONObject,
-        *,
-        guild_id: undefined.UndefinedNoneOr[snowflakes.Snowflake] = undefined.UNDEFINED,
-    ) -> commands.SlashCommand:
-        """Parse a raw payload from Discord into a slash command object.
-
-        Parameters
-        ----------
-        payload : hikari.internal.data_binding.JSONObject
-            The JSON payload to deserialize.
-
-        Other Parameters
-        ----------------
-        guild_id : hikari.undefined.UndefinedNoneOr[hikari.snowflakes.Snowflake]
-            The ID of the guild this command belongs to. If this is specified
-            then this will be prioritised over `"guild_id"` in the payload.
-
-        Returns
-        -------
-        hikari.commands.SlashCommand
-            The deserialized slash command object.
-
-        Raises
-        ------
-        builtins.KeyError
-            If `guild_id` is left as `hikari.undefined.UNDEFINED` when
-            `"guild_id"` is not present in the passed payload for the payload of
-            the integration.
-        """
-
-    @abc.abstractmethod
-    def deserialize_context_menu_command(
-        self,
-        payload: data_binding.JSONObject,
-        *,
-        guild_id: undefined.UndefinedNoneOr[snowflakes.Snowflake] = undefined.UNDEFINED,
-    ) -> commands.ContextMenuCommand:
-        """Parse a raw payload from Discord into a context menu command object.
-
-        Parameters
-        ----------
-        payload : hikari.internal.data_binding.JSONObject
-            The JSON payload to deserialize.
-
-        Other Parameters
-        ----------------
-        guild_id : hikari.undefined.UndefinedNoneOr[hikari.snowflakes.Snowflake]
-            The ID of the guild this command belongs to. If this is specified
-            then this will be prioritised over `"guild_id"` in the payload.
-
-        Returns
-        -------
-        hikari.commands.ContextMenuCommand
-            The deserialized context menu command object.
-
-        Raises
-        ------
-        builtins.KeyError
-            If `guild_id` is left as `hikari.undefined.UNDEFINED` when
-            `"guild_id"` is not present in the passed payload for the payload of
-            the integration.
-        """
-
-    @abc.abstractmethod
     def deserialize_command(
         self,
         payload: data_binding.JSONObject,
@@ -1297,6 +1230,68 @@ class EntityFactory(abc.ABC):
     ##################
 
     @abc.abstractmethod
+    def deserialize_action_row(self, payload: data_binding.JSONObject) -> message_models.ActionRowComponent:
+        """Parse a raw payload from Discord into an action row component object.
+
+        Parameters
+        ----------
+        payload : hikari.internal.data_binding.JSONObject
+            The JSON payload to deserialize.
+
+        Returns
+        -------
+        hikari.messages.ActionRowComponent
+            The deserialized action row component.
+        """
+
+    @abc.abstractmethod
+    def deserialize_button(self, payload: data_binding.JSONObject) -> message_models.ButtonComponent:
+        """Parse a raw payload from Discord into a button component object.
+
+        Parameters
+        ----------
+        payload : hikari.internal.data_binding.JSONObject
+            The JSON payload to deserialize.
+
+        Returns
+        -------
+        hikari.messages.ButtonComponent
+            The deserialized button component.
+        """
+
+    @abc.abstractmethod
+    def deserialize_select_menu(self, payload: data_binding.JSONObject) -> message_models.SelectMenuComponent:
+        """Parse a raw payload from Discord into a select menu component object.
+
+        Parameters
+        ----------
+        payload : hikari.internal.data_binding.JSONObject
+            The JSON payload to deserialize.
+
+        Returns
+        -------
+        hikari.messages.SelectMenuComponent
+            The deserialized button component.
+        """
+
+    @abc.abstractmethod
+    def deserialize_component(self, payload: data_binding.JSONObject) -> message_models.PartialComponent:
+        """Parse a raw payload from Discord into a message component object.
+
+        Parameters
+        ----------
+        payload : hikari.internal.data_binding.JSONObject
+            The JSON payload to deserialize.
+
+        Returns
+        -------
+        hikari.messages.PartialComponent
+            The deserialized message component.
+        hikari.errors.UnrecognisedEntityError
+            If the message component type isn't recognised.
+        """
+
+    @abc.abstractmethod
     def deserialize_partial_message(self, payload: data_binding.JSONObject) -> message_models.PartialMessage:
         """Parse a raw payload from Discord into a partial message object.
 
@@ -1368,102 +1363,6 @@ class EntityFactory(abc.ABC):
             guild ID was passed for the `guild_id` parameter.
 
             If this is raised, no guild ID info was provided anywhere.
-        """
-
-    ##########################
-    # SCHEDULED EVENT MODELS #
-    ##########################
-
-    @abc.abstractmethod
-    def deserialize_scheduled_external_event(
-        self, payload: data_binding.JSONObject
-    ) -> scheduled_events_models.ScheduledExternalEvent:
-        """Parse a raw payload from Discord into a scheduled external event object.
-
-        Parameters
-        ----------
-        payload : hikari.internal.data_binding.JSONObject
-            The JSON payload to deserialize.
-
-        Returns
-        -------
-        hikari.scheduled_events.ScheduledExternalEvent
-            The deserialized scheduled external event object.
-        """
-
-    @abc.abstractmethod
-    def deserialize_scheduled_stage_event(
-        self, payload: data_binding.JSONObject
-    ) -> scheduled_events_models.ScheduledStageEvent:
-        """Parse a raw payload from Discord into a scheduled stage event object.
-
-        Parameters
-        ----------
-        payload : hikari.internal.data_binding.JSONObject
-            The JSON payload to deserialize.
-
-        Returns
-        -------
-        hikari.scheduled_events.ScheduledStageEvent
-            The deserialized scheduled stage event object.
-        """
-
-    @abc.abstractmethod
-    def deserialize_scheduled_voice_event(
-        self, payload: data_binding.JSONObject
-    ) -> scheduled_events_models.ScheduledVoiceEvent:
-        """Parse a raw payload from Discord into a scheduled voice event object.
-
-        Parameters
-        ----------
-        payload : hikari.internal.data_binding.JSONObject
-            The JSON payload to deserialize.
-
-        Returns
-        -------
-        hikari.scheduled_events.ScheduledVoiceEvent
-            The deserialized scheduled voice event object.
-        """
-
-    @abc.abstractmethod
-    def deserialize_scheduled_event(self, payload: data_binding.JSONObject) -> scheduled_events_models.ScheduledEvent:
-        """Parse a raw payload from Discord into a scheduled event object.
-
-        Parameters
-        ----------
-        payload : hikari.internal.data_binding.JSONObject
-            The JSON payload to deserialize.
-
-        Returns
-        -------
-        hikari.scheduled_events.ScheduledEvent
-            The deserialized scheduled event object.
-        """
-
-    @abc.abstractmethod
-    def deserialize_scheduled_event_user(
-        self,
-        payload: data_binding.JSONObject,
-        *,
-        guild_id: undefined.UndefinedOr[snowflakes.Snowflake] = undefined.UNDEFINED,
-    ) -> scheduled_events_models.ScheduledEventUser:
-        """Parse a raw payload from Discord into a scheduled event user object.
-
-        Parameters
-        ----------
-        payload : hikari.internal.data_binding.JSONObject
-            The JSON payload to deserialize.
-
-        Other Parameters
-        ----------------
-        guild_id : hikari.undefined.UndefinedOr[hikari.snowflakes.Snowflake]
-            The ID of the guild the user belongs to. If this is specified
-            then it is prioritised over `guild_id` in the payload.
-
-        Returns
-        -------
-        hikari.scheduled_events.ScheduledEventUser
-            The deserialized scheduled event user object.
         """
 
     ###################
--- a/hikari/api/entity_factory.py
+++ b/hikari/api/entity_factory.py
@@ -42,6 +42,7 @@ if typing.TYPE_CHECKING:
     from hikari import invites as invite_models
     from hikari import messages as message_models
     from hikari import presences as presence_models
+    from hikari import scheduled_events as scheduled_events_models
     from hikari import sessions as gateway_models
     from hikari import snowflakes
     from hikari import stickers as sticker_models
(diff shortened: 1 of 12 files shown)


hikari/impl/event_manager_base.py
```diff
@@ -289,26 +283,41 @@ def filtered(
     else:
         event_types = event_types.dispatches()
 
-    def decorator(method: UnboundMethodT[EventManagerBaseT], /) -> UnboundMethodT[EventManagerBaseT]:
+    bitmask = 0
+    for event_type in event_types:
+        bitmask |= event_type.bitmask()
+
+    # https://github.com/python/mypy/issues/2087
+    def decorator(method: UnboundMethodT[EventManagerBaseT], /) -> _FilteredMethodT:
```
davfsa, line 291:
```suggestion
    def decorator(method: typing.Any, /) -> _FilteredMethodT:
```

Would this work and remove all the ignores?

FasterSpeeding: Both of these changes kinda breaks the typing though; the methods being typed as _FilteredMethodT on the class is wrong since you'd also have to explicitly pass self and do you really want all the methods on the impl to be typed as _FilteredMethodT which includes internal detail which isn't meant to be exposed typing wise

hikari/impl/event_manager.py
```diff
@@ -241,16 +241,17 @@ async def on_guild_create(self, shard: gateway_shard.GatewayShard, payload: data
...
                     self._cache.set_voice_state(voice_state)
 
-        recv_chunks = self._enabled_for_event(shard_events.MemberChunkEvent) or self._cache_enabled_for(
-            config.CacheComponents.MEMBERS
-        )
         members_declared = self._intents & intents_.Intents.GUILD_MEMBERS
         presences_declared = self._intents & intents_.Intents.GUILD_PRESENCES
 
-        # When intents are enabled discord will only send other member objects on the guild create
+        # When intents are enabled Discord will only send the member objects on the guild create
```
FasterSpeeding, line 247: it is very much other member objects, not "the member objects" since you always get the current user's member; that's not a typo

hikari/impl/event_manager.py
```diff
@@ -162,50 +157,55 @@ async def on_channel_pins_update(self, shard: gateway_shard.GatewayShard, payloa
         # TODO: we need a method for this specifically
         await self.dispatch(self._event_factory.deserialize_channel_pins_update_event(shard, payload))
 
-    # Internal granularity is preferred for GUILD_CREATE over decorator based filtering due to its large cache scope.
-    async def on_guild_create(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
+    @event_manager_base.filtered(
+        (guild_events.GuildAvailableEvent, guild_events.GuildJoinEvent),
+        config.CacheComponents.GUILDS
+        | config.CacheComponents.GUILD_CHANNELS
```
FasterSpeeding, line 163: What's the point of checking it twice and since the comment was removed is the internal granularity also going to be removed?

hikari/impl/event_manager_base.py
```diff
@@ -448,9 +451,10 @@ def get_listeners(
     ) -> typing.Collection[event_manager_.CallbackT[event_manager_.EventT_co]]:
         if polymorphic:
             listeners: typing.List[event_manager_.CallbackT[event_manager_.EventT_co]] = []
-            for subscribed_event_type, subscribed_listeners in self._listeners.items():
-                if issubclass(subscribed_event_type, event_type):
-                    listeners += subscribed_listeners
+            for event in event_type.dispatches():
+                if subscribed_listeners := self._listeners.get(event):
+                    listeners.extend(subscribed_listeners)
```
FasterSpeeding, line 456: This changes the nature of the method and breaks from the documented behaviour for polymorphic

tests/hikari/impl/test_event_manager.py
```diff
@@ -106,7 +105,7 @@ def event_manager(self, entity_factory, event_factory):
         return obj
 
     @pytest.fixture()
-    def stateless_event_manager(self, event_factory, entity_factory):
+    def stateless_event_manager_impl(self, event_factory, entity_factory):
```
FasterSpeeding, line 108: Renaming these fixtures which are referenced here instead of the imports adds a ton of noise to the diff in this pr which means ima put off reviewing the changes in this module cause its just a big "fuck you, review every line"

hikari/impl/event_manager.py
```diff
@@ -241,16 +241,17 @@ async def on_guild_create(self, shard: gateway_shard.GatewayShard, payload: data
                 for voice_state in voice_states.values():
                     self._cache.set_voice_state(voice_state)
 
-        recv_chunks = self._enabled_for_event(shard_events.MemberChunkEvent) or self._cache_enabled_for(
```
FasterSpeeding, line 244: you know if nobody's listening to chunk events and we're not caching for chunk events its just a ton of wasted bandwidth and time so not sure why that was removed

tests/hikari/impl/test_event_manager.py
davfsa: Sorry, I had to rename `event_manager` to `event_manager_impl` since it shadowed the name. I also went ahead and did it for this one too for consistency sake.

hikari/impl/event_manager.py
davfsa: It was moved bellow. All the other checks are done first before the enabled for event, since its the most expensive

hikari/impl/event_manager.py
davfsa: Will revert 👍

hikari/impl/event_manager.py
FasterSpeeding: You're gonna have to dupe the enabled for event check for member chunk between here and the filtered decorator call so that seems like an arbitrary micro-optimisation

hikari/impl/event_manager.py
FasterSpeeding: You also don't account for chunking here and won't be able to if an explicit chunk setting is added so...

davfsa: The check is really fast, so I didn't see the point of having it dispatch every time, regardless of the cache settings/listeners/waiters

tests/hikari/impl/test_event_manager.py
FasterSpeeding: Welp if there's no other alternative and what I describe isn't doable then I'll just review this once i feel like reviewing a over 1k which gonna be a while tbh

hikari/impl/event_manager.py
```diff
@@ -62,9 +60,6 @@
     from hikari.internal import data_binding
 
 
-_LOGGER: typing.Final[logging.Logger] = logging.getLogger("hikari.event_manager")
```
FasterSpeeding, line 65: Would this debug logging not still be helpful for seeing how these events are being treated

davfsa: For now its not being used. When I pick this pr back (after exams), I will revert some of the changes and update the PR

hikari/impl/event_manager_base.py
davfsa: Updated docstring. Will add a `FIXME` to cause the pipelines to fail and make sure to add a breaking changelog fragment

davfsa:
Ready for another review

Flake8 failing is intentional btw

hikari/impl/event_manager.py
```diff
@@ -162,41 +162,50 @@ async def on_channel_pins_update(self, shard: gateway_shard.GatewayShard, payloa
...
         """See https://discord.com/developers/docs/topics/gateway#guild-create for more info."""
-        enabled_for_event = self._enabled_for_event(guild_events.GuildAvailableEvent)
-        if not enabled_for_event and self._cache:
+        event: typing.Union[guild_events.GuildAvailableEvent, guild_events.GuildJoinEvent, None]
+        members_cache_enabled = self._cache_enabled_for(config.CacheComponents.MEMBERS)
+
+        if "unavailable" in payload and self._enabled_for_event(guild_events.GuildAvailableEvent):
+            event = self._event_factory.deserialize_guild_available_event(shard, payload)
+        elif self._enabled_for_event(guild_events.GuildJoinEvent):
+            event = self._event_factory.deserialize_guild_join_event(shard, payload)
```
FasterSpeeding, line 176: This logic is wrong

hikari/impl/event_manager.py
```diff
@@ -162,41 +162,50 @@ async def on_channel_pins_update(self, shard: gateway_shard.GatewayShard, payloa
...
-    async def on_guild_create(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
+    # Internal granularity is preferred for GUILD_CREATE over decorator based filtering due to its large scope.
+    async def on_guild_create(  # noqa: C901 - Function too complex
+        self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject
+    ) -> None:
         """See https://discord.com/developers/docs/topics/gateway#guild-create for more info."""
-        enabled_for_event = self._enabled_for_event(guild_events.GuildAvailableEvent)
-        if not enabled_for_event and self._cache:
+        event: typing.Union[guild_events.GuildAvailableEvent, guild_events.GuildJoinEvent, None]
+        members_cache_enabled = self._cache_enabled_for(config.CacheComponents.MEMBERS)
```
FasterSpeeding, line 171: Moving this out to a variable seems kinda unnecessary

hikari/impl/event_manager.py
davfsa: How come?

FasterSpeeding: It'll treat available events as join events if there's listeners for join event but not available event

davfsa: Oh true, didnt think of that. Thanks for pointing it out!

hikari/impl/event_manager.py
```diff
@@ -162,13 +162,33 @@ async def on_channel_pins_update(self, shard: gateway_shard.GatewayShard, payloa
...
+    async def on_guild_create(  # noqa: C901 - Function too complex
+        self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject
+    ) -> None:
         """See https://discord.com/developers/docs/topics/gateway#guild-create for more info."""
-        enabled_for_event = self._enabled_for_event(guild_events.GuildAvailableEvent)
-        if not enabled_for_event and self._cache:
+        event: typing.Union[guild_events.GuildAvailableEvent, guild_events.GuildJoinEvent, None]
+
+        if "unavailable" in payload and self._enabled_for_event(guild_events.GuildAvailableEvent):
+            event = self._event_factory.deserialize_guild_available_event(shard, payload)
```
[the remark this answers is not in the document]
davfsa: Added :+1:

hikari/api/event_manager.py
```diff
@@ -356,6 +356,7 @@ async def on_message(event):
         Wait_for: `hikari.api.event_manager.EventManager.wait_for`
         """
 
+    # FIXME: Breaking change here, make sure to document
```
FasterSpeeding, line 359: This won't pass linting until this is fixed

davfsa: As stated in the comment of this PR, intentional.

Will remove and fix when opening a pr to merge into master

FasterSpeeding: Why not just add the changelog entry now and rename it when you open the pr? The pipelines do fail until you have changelog entries.

BTW you could've communicated that better, I usually don't read comments after they've been marked as resolved since you know its hidden and I'd just assumed you wanted to add the changelog entry before merging this since you know if pipelines fail my assumption is that a pr isn't read yet

davfsa: Sorry, should have specified better. Will add the fragment changelog later and, if it's good to merge, do so and open the big pr

hikari/api/entity_factory.py
```diff
@@ -42,7 +42,6 @@
     from hikari import invites as invite_models
```
davfsa, line 42: Changes in this file are a rebase problem. Need to revert

hikari/impl/entity_factory.py
```diff
@@ -77,7 +76,6 @@
     commands.OptionType.CHANNEL: snowflakes.Snowflake,
     commands.OptionType.ROLE: snowflakes.Snowflake,
     commands.OptionType.MENTIONABLE: snowflakes.Snowflake,
-    commands.OptionType.ATTACHMENT: snowflakes.Snowflake,
```
davfsa, line 80: Incorrect removal

hikari/impl/entity_factory.py
```diff
@@ -46,7 +46,6 @@
 from hikari import messages as message_models
 from hikari import permissions as permission_models
 from hikari import presences as presence_models
-from hikari import scheduled_events as scheduled_events_models
```
davfsa, line 49: Incorrect removal

hikari/impl/entity_factory.py
```diff
@@ -424,14 +405,14 @@ def __init__(self, app: traits.RESTAware) -> None:
...
+            commands.CommandType.USER: self._deserialize_context_menu_command,
+            commands.CommandType.MESSAGE: self._deserialize_context_menu_command,
         }
         self._component_type_mapping = {
-            message_models.ComponentType.ACTION_ROW: self._deserialize_action_row,
-            message_models.ComponentType.BUTTON: self._deserialize_button,
-            message_models.ComponentType.SELECT_MENU: self._deserialize_select_menu,
+            message_models.ComponentType.ACTION_ROW: self.deserialize_action_row,
+            message_models.ComponentType.BUTTON: self.deserialize_button,
+            message_models.ComponentType.SELECT_MENU: self.deserialize_select_menu,
```
davfsa, line 415: Incorrect change

hikari/impl/entity_factory.py
```diff
@@ -452,11 +433,6 @@ def __init__(self, app: traits.RESTAware) -> None:
             base_interactions.InteractionType.MESSAGE_COMPONENT: self.deserialize_component_interaction,
             base_interactions.InteractionType.AUTOCOMPLETE: self.deserialize_autocomplete_interaction,
         }
-        self._scheduled_event_type_mapping = {
-            scheduled_events_models.ScheduledEventType.STAGE_INSTANCE: self.deserialize_scheduled_stage_event,
-            scheduled_events_models.ScheduledEventType.VOICE: self.deserialize_scheduled_voice_event,
-            scheduled_events_models.ScheduledEventType.EXTERNAL: self.deserialize_scheduled_external_event,
```
davfsa, line 458: Incorrect removal

hikari/impl/entity_factory.py
```diff
@@ -1829,7 +1805,7 @@ def deserialize_slash_command(
             version=snowflakes.Snowflake(payload["version"]),
         )
 
-    def deserialize_context_menu_command(
+    def _deserialize_context_menu_command(
```
davfsa, line 1808: Incorrect change (includes bellow)

hikari/impl/entity_factory.py
```diff
@@ -2776,117 +2752,6 @@ def deserialize_member_presence(  # noqa: CFQ001 - Max function length
             client_status=client_status,
         )
 
-    ##########################
-    # SCHEDULED EVENT MODELS #
-    ##########################
-
-    def deserialize_scheduled_external_event(
```
davfsa, line 2783: Incorrect removal

hikari/impl/event_manager.py
```diff
@@ -654,33 +667,3 @@ async def on_webhooks_update(self, shard: gateway_shard.GatewayShard, payload: d
     async def on_interaction_create(self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject) -> None:
         """See https://discord.com/developers/docs/topics/gateway#interaction-create for more info."""
         await self.dispatch(self._event_factory.deserialize_interaction_create_event(shard, payload))
-
-    async def on_guild_scheduled_event_create(
```
davfsa, line 658: Incorrect removal

hikari/impl/event_manager_base.py
```diff
@@ -546,8 +550,7 @@ def dispatch(self, event: event_manager_.EventT_inv) -> asyncio.Future[typing.An
                 predicate, future = waiter
                 if not future.done():
                     try:
-                        result = predicate(event)
-                        if not result:
+                        if predicate is not None and not predicate(event):
```
davfsa, line 553:
```suggestion
                        if predicate and not predicate(event):
```

tests/hikari/impl/test_entity_factory.py
```diff
@@ -5240,296 +5243,6 @@ def test_deserialize_member_presence_with_unset_activity_sub_fields(self, entity
         assert activity.secrets.spectate is None
         assert activity.secrets.match is None
 
-    ##########################
```
davfsa, line 5243: Incorrect removal

tests/hikari/impl/test_event_manager.py
```diff
@@ -98,1399 +96,1152 @@ def event_factory(self):
...
         payload = {"id": "123"}
 
-        await event_manager.on_interaction_create(shard, payload)
+        await event_manager_impl.on_interaction_create(shard, payload)
 
         event_factory.deserialize_interaction_create_event.assert_called_once_with(shard, payload)
-        event_manager.dispatch.assert_awaited_once_with(event_factory.deserialize_interaction_create_event.return_value)
-
-    @pytest.mark.asyncio()
-    async def test_on_guild_scheduled_event_create(
```
davfsa, line 1419: Incorrect removal

hikari/impl/event_manager_base.py
```diff
@@ -362,15 +371,29 @@ def __init__(
...
-                        member, undefined.UNDEFINED, cache_components != cache_components.NONE
-                    )
+                    self._consumers[event_name] = _Consumer(member, -1, cache_components != cache_components.NONE)
 
-    def _clear_enabled_cache(self) -> None:
-        self._enabled_consumers_cache = {}
+    def _increment_listener_group_count(
+        self,
+        event_type: typing.Type[base_events.Event],
+        count: typing.Literal[-1, 1],
```
davfsa, line 383:
```suggestion
        self,
        event_type: typing.Type[base_events.Event],
        count: typing.Literal[-1, 1]
```

And then i would need to run black

hikari/impl/event_manager_base.py
```diff
@@ -306,26 +302,41 @@ def filtered(
...
-        method.__event_types__ = event_types  # type: ignore[attr-defined]
+        method.__events_bitmask__ = bitmask  # type: ignore[attr-defined]
         assert isinstance(method, _FilteredMethodT), "Incorrect attribute(s) set for a filtered method"
         return method  # type: ignore[unreachable]
 
     return decorator
 
 
-@attr.define(hash=True)
+@attr.define(weakref_slot=False, hash=True)
```
davfsa, line 319:
```suggestion
@attr.define(weakref_slot=False)
```
And then remove `hash=False` from attr.field below

hikari/impl/event_manager_base.py
```diff
@@ -306,26 +302,41 @@ def filtered(
...
+    """The registered events bitmask."""
 
     is_caching: bool = attr.ib(hash=False)
     """Cached value of whether or not this consumer is making cache calls in the current env."""
 
+    listener_group_count: int = attr.ib(init=False, hash=False, default=0)
+    """The number of listener groups registered to this consumer."""
+
+    waiter_group_count: int = attr.ib(init=False, hash=False, default=0)
+    """The number of waiters groups registered to this consumer."""
```
davfsa, line 334:
```suggestion
    events_bitmask: int = attr.field()
    """The registered events bitmask."""

    is_caching: bool = attr.field()
    """Cached value of whether or not this consumer is making cache calls in the current env."""

    listener_group_count: int = attr.field(init=False, default=0)
    """The number of listener groups registered to this consumer."""

    waiter_group_count: int = attr.field(init=False, default=0)
    """The number of waiters groups registered to this consumer."""
```

hikari/impl/event_manager_base.py
```diff
@@ -78,9 +74,13 @@
     ]
     _EventStreamT = typing.TypeVar("_EventStreamT", bound="EventStream[typing.Any]")
 
+_LOGGER: typing.Final[logging.Logger] = logging.getLogger("hikari.event_manager")
+
 
 @typing.runtime_checkable
 class _FilteredMethodT(fast_protocol.FastProtocolChecking, typing.Protocol):
```
davfsa, line 81: Could we avoid having to make it runtime checkable?

davfsa:
There are some new events missing the filter:
```
on_interaction_create
on_guild_scheduled_event_create
on_guild_scheduled_event_delete
on_guild_scheduled_event_update
on_guild_scheduled_event_user_add
on_guild_scheduled_event_user_remove
```

FasterSpeeding (approved):

PR #906, opened 2021-11-15
Switch to more relaxed requirements

davfsa:


requirements.txt
```diff
@@ -1,5 +1,5 @@
-aiohttp==3.8.1
-attrs==21.2.0
-colorlog==6.6.0
-multidict==5.2.0
-pure25519==0.0.1
+aiohttp~=3.8.0
+attrs~=21.2.0
```
FasterSpeeding, line 2: why still restrict them to one minor ver

davfsa: We dont. `~=` in this case will mean "Anything that matches `21.2.*` but not lower than `21.2.0`.

Left the minor version in case we ever want to limit it for some reason:

ie: A security vulnerability that was fixed from `21.2.4` onward, to what we would change that line to be `attrs~=21.2.4`

FasterSpeeding: > Anything that matches 21.2.* but not lower than 21.2.0.

This is exactly what I described, you limit it to one minor version even though minor bumps shouldn't introduce breaking changes for a stable project just like patch.

FasterSpeeding (approved):

PR #910, opened 2021-11-21
Fix reference messages not being fully deleted

### LINKED ISSUE #904: Referenced message entries exceed the max messages settings
### Steps to reproduce

~~There's no steps but to run the bot as usual.~~ Send a message, let's say it's message A. Then, reply to message A, this reply is message B. And finally, reply to message B.

### Expected result

~~`Cache.get_messages_view()` returns a bit more than n messages if not exact, where n is the max messages we put in the config.~~ Message B has the referenced_message attribute set to A.

### Actual result

~~It returns a lot more messages than it should. Turned out, `Cache._referenced_messages` seems to be caching messages that should've been GC'd as they're no longer referenced.~~ Message B has the referenced_message attribute set to None. This would lead to a memory leak if people were to reply a lot as it would leave some referenced messages untouched when garbage collecting.

### System info

```shell
hikari (2.0.0.dev104) [HEAD]
located at /home/nori/.local/lib/python3.9/site-packages/hikari
CPython 3.9.7 GCC 11.1.0
Linux arch 5.14.16-zen1-1-zen #1 ZEN SMP PREEMPT Tue, 02 Nov 2021 22:23:00 +0000 x86_64
```

### Further info

Might be somewhat relevant to #877 and perhaps #894 if we were to hide referenced messages?

### Checklist

- [X] I have made sure to remove ANY sensitive information (bot token, passwords, credentials, personal details, etc.).

davfsa:
Thanks to @norinorin for finding the bug and providing the patch!

*The patch was submitted with explicit permission from them*

### Checklist

### Related issues
Closes #904

### DIFF
--- a/hikari/impl/cache.py
+++ b/hikari/impl/cache.py
@@ -1473,8 +1473,13 @@ class CacheImpl(cache.MutableCache):
             if guild_record:
                 self._garbage_collect_member(guild_record, message.object.member, decrement=1)
 
-        if message.object.referenced_message:
-            self._garbage_collect_message(message.object.referenced_message, decrement=1)
+        referenced_message = message.object.referenced_message
+        if not referenced_message and message.object.message_reference and message.object.message_reference.id:
+            msg_id = message.object.message_reference.id
+            referenced_message = self._message_entries.get(msg_id) or self._referenced_messages.get(msg_id)
+
+        if referenced_message:
+            self._garbage_collect_message(referenced_message, decrement=1)
 
         if message.object.mentions.users:
             for user in message.object.mentions.users.values():
--- a/changes/910.bugfix.md
+++ b/changes/910.bugfix.md
@@ -1,0 +1,1 @@
+Fix reference messages not being fully deleted, which could cause a memory leak


hikari/impl/cache.py
```diff
@@ -1473,8 +1473,13 @@ def _garbage_collect_message(
             if guild_record:
                 self._garbage_collect_member(guild_record, message.object.member, decrement=1)
 
-        if message.object.referenced_message:
-            self._garbage_collect_message(message.object.referenced_message, decrement=1)
+        referenced_message = message.object.referenced_message
+        if not referenced_message and message.object.message_reference and message.object.message_reference.id:
+            msg_id = message.object.message_reference.id
+            referenced_message = self._message_entries.get(msg_id) or self._referenced_messages.get(msg_id)
```
FasterSpeeding, line 1479: Are you sure this actually fixes the issue, if a referenced message is cached when message.referenced_message is None then wouldn't that infer there's a case the update/set logic isn't accounting for (possibly to do with the idk lets say the referenced message being deleted and that not cascading to places where its being referenced).

hikari/impl/cache.py
```diff
@@ -1473,8 +1473,13 @@ def _garbage_collect_message(
...
+
```
FasterSpeeding, line 1480: Also you're not accounting for what kind of reference this is so for stuff like cross posts, channel follows, and pins adds where you'd have message_reference.id but not referenced_message this could lead to other edge case bugs.

hikari/impl/cache.py
norinorin: I'm not aware if there are any other cases, but another way that I could think of to handle this particular case is to ensure that we're not overriding referenced_message's referenced_message since Discord gives us None.

davfsa:
Closed to let @norinorin re-open with their changes

PR #920, opened 2021-11-26
Add min_value and max_value to CommandOption

thesadru:
### Summary
Add min_value and max_value to CommandOption. 

Reference: https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure

### Checklist

- [x] I have run `nox` and all the pipelines have passed. (it's taking so long 😭)

### DIFF
--- a/hikari/commands.py
+++ b/hikari/commands.py
@@ -146,10 +146,10 @@ class CommandOption:
     If `builtins.None`, then all channel types will be accepted.
     """
 
-    min_value: typing.Optional[float] = attr.field(default=None, repr=False)
+    min_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
     """The minimum value permitted (inclusive)"""
 
-    max_value: typing.Optional[float] = attr.field(default=None, repr=False)
+    max_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
     """The minimum value permitted (inclusive)"""
 
 
--- a/hikari/commands.py
+++ b/hikari/commands.py
@@ -147,10 +147,16 @@ class CommandOption:
     """
 
     min_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
-    """The minimum value permitted (inclusive)"""
+    """The minimum value permitted (inclusive)
+    
+    If the option type is `hikari.commands.OptionType.INTEGER` than this must be an `builtins.int`.
+    """
 
     max_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
-    """The minimum value permitted (inclusive)"""
+    """The minimum value permitted (inclusive)
+    
+    If the option type is `hikari.commands.OptionType.INTEGER` than this must be an `builtins.int`.
+    """
 
 
 @attr_extensions.with_copy
--- a/hikari/commands.py
+++ b/hikari/commands.py
@@ -148,13 +148,13 @@ class CommandOption:
 
     min_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
     """The minimum value permitted (inclusive)
-    
+
     If the option type is `hikari.commands.OptionType.INTEGER` than this must be an `builtins.int`.
     """
 
     max_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
     """The minimum value permitted (inclusive)
-    
+
     If the option type is `hikari.commands.OptionType.INTEGER` than this must be an `builtins.int`.
     """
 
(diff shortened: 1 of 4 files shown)


hikari/impl/entity_factory.py
```diff
@@ -1913,6 +1916,15 @@ def serialize_command_option(self, option: commands.CommandOption) -> data_bindi
         if option.options is not None:
             payload["options"] = [self.serialize_command_option(suboption) for suboption in option.options]
 
+        if option.min_value is not None:
+            payload["min_value"] = (
+                math.ceil(option.min_value) if option.type is commands.OptionType.INTEGER else option.min_value
+            )
+        if option.max_value is not None:
+            payload["max_value"] = (
+                math.floor(option.max_value) if option.type is commands.OptionType.INTEGER else option.max_value
```
davfsa, line 1925: Maybe its better to document the behavior when using `INTEGER` or `NUMBER` and just assume people read documentation. Discord will probably error if this happens, so it wont be as ominous as just it being casted upwards or downwards

hikari/commands.py
```diff
@@ -146,6 +146,12 @@ class CommandOption:
     If `builtins.None`, then all channel types will be accepted.
     """
 
+    min_value: typing.Optional[float] = attr.field(default=None, repr=False)
+    """The minimum value permitted (inclusive)"""
+
+    max_value: typing.Optional[float] = attr.field(default=None, repr=False)
```
davfsa, line 152:
```suggestion
    min_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
    """The minimum value permitted (inclusive)"""

    max_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
```
As well as documentation for the `int`/`float` behavior would be nice

davfsa (commented):
Looks really good, thanks a lot for the contribution! Just got some little points on explaining better some behavior

hikari/impl/entity_factory.py
thesadru: I'd say it's always better to gracefully handle floats since for the user the outcome would always be the same as if floats were allowed. Might be a good idea to recommend using integers in the docs though.

thesadru:
I think this is ready for a re-review.

hikari/commands.py
```diff
@@ -146,6 +146,18 @@ class CommandOption:
...
+    """The minimum value permitted (inclusive)
+
+    If the option type is `hikari.commands.OptionType.INTEGER` than this must be an `builtins.int`.
+    """
+
+    max_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
+    """The minimum value permitted (inclusive)
+
+    If the option type is `hikari.commands.OptionType.INTEGER` than this must be an `builtins.int`.
+    """
```
davfsa, line 159:
```suggestion
    min_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
    """The minimum value permitted (inclusive).

    This will be `builtins.int` if the type of the option is `hikari.commands.OptionType.INTEGER`
    and `builtins.float` if the type is `hikari.commands.OptionType.NUMBER`. 
    """

    max_value: typing.Union[int, float, None] = attr.field(default=None, repr=False)
    """The maximum value permitted (inclusive).

    This will be `builtins.int` if the type of the option is `hikari.commands.OptionType.INTEGER`
    and `builtins.float` if the type is `hikari.commands.OptionType.NUMBER`. 
    """
```

hikari/impl/entity_factory.py
```diff
@@ -1913,6 +1916,13 @@ def serialize_command_option(self, option: commands.CommandOption) -> data_bindi
         if option.options is not None:
             payload["options"] = [self.serialize_command_option(suboption) for suboption in option.options]
 
+        # because discord does not allow floating point min and max values with integer options we have to round
```
davfsa, line 1919: Still dont believe this is the way to go. We should just be adding it raw, since if not it could confuse some people.

An example of a case where this can be confusing is when a user specifies `INTEGER` by mistake and then starts receiving values under (lets say) `1.2` . This could take some time to debug, instead of just erroring as soon as the command is created, since (and I cant really test this until the weekend) Discord should error

davfsa (commented):
Sorry for taking so long to review, been caught up in a lot of other issues. Hopefully this will be the last review and we can merge it :)

davfsa (approved):
LGTM. Thanks a lot for taking the time to work on this!

hikari/impl/entity_factory.py
```diff
@@ -28,6 +28,7 @@
 
 import datetime
 import logging
+import math
```
davfsa, line 31:
```suggestion
```

davfsa (changes requested):
Just to make linting happy

davfsa (approved):

FasterSpeeding (approved):

PR #930, opened 2021-12-15
Fix PartialSticker.image_url not passing the hash as a string

davfsa:
### Checklist

### DIFF
--- a/hikari/internal/routes.py
+++ b/hikari/internal/routes.py
@@ -556,7 +556,7 @@ CDN_TEAM_ICON: typing.Final[CDNRoute] = CDNRoute("/team-icons/{team_id}/{hash}",
 # undocumented on the Discord docs.
 CDN_CHANNEL_ICON: typing.Final[CDNRoute] = CDNRoute("/channel-icons/{channel_id}/{hash}", {PNG, *JPEG_JPG, WEBP})
 
-CDN_STICKER: typing.Final[CDNRoute] = CDNRoute("/stickers/{hash}", {PNG, LOTTIE}, sizable=False)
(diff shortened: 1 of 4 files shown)


hikari/internal/routes.py
```diff
@@ -253,7 +253,7 @@ def compile(
                 + ", ".join(self.valid_formats)
             )
 
-        if "hash" in kwargs and not kwargs["hash"].startswith("a_") and file_format == GIF:
+        if "hash" in kwargs and not str(kwargs["hash"]).startswith("a_") and file_format == GIF:
```
FasterSpeeding, line 256: Ok so this isn't a fix to the issue as much as it just obfuscates it, really the cdn sticker routes are erroneously referring to pack_id and sticker_id as "hash" leading to this logic being triggered on them when it shouldn't so those should probably be renamed.
Also there's a point here for maybe moving "hash" to a separate optional key-word argument which is explicitly typed as `str` to avoid issues like this in the future

davfsa: I havent had time to fix this, but I did have a different way of fixing thus, but I have been unable to do it just yet. Will update the PR when I get to it.

hikari/stickers.py
```diff
@@ -157,7 +157,7 @@ def image_url(self) -> files.URL:
         """
         ext = "json" if self.format_type is StickerFormatType.LOTTIE else "png"
 
-        return routes.CDN_STICKER.compile_to_file(urls.CDN_URL, hash=self.id, file_format=ext)
+        return routes.CDN_STICKER.compile_to_file(urls.CDN_URL, hash=str(self.id), file_format=ext)
```
FasterSpeeding, line 160: the field should be renamed to sticker_id instead of casting to str here

davfsa: Changed :+1:

FasterSpeeding (approved):

PR #933, opened 2021-12-16
Fix some bugs in message deserialization

davfsa:
- Remove case for setting member to undefined.Undefined in full message deserialization
- Don't set member to undefined.UNDEFINED on partial message if message was sent by a webhook

### Checklist


tests/hikari/impl/test_entity_factory.py
```diff
@@ -4314,7 +4321,7 @@ def test_deserialize_message_with_other_unset_fields(self, entity_factory_impl,
         assert message.application.cover_image_hash is None
         assert message.application.icon_hash is None
         assert message.referenced_message is None
-        assert message.member is undefined.UNDEFINED
+        assert message.member is None
```
davfsa, line 4324: `referenced_message` should also be fixed, it should never be undefined for a full Message

davfsa: Looks like attrs handles re-assigning the attributes in a really nice way (mostly not editing them if it doesnt have to) and it makes mypy happy, so I think we will stick with it

FasterSpeeding (approved):

PR #936, opened 2021-12-18
pep8: replacing `value is True / False` with the ternary operator` if…

Animatea:
… value / if not value`.

### Summary
Small change: replacing `value is True / False` with the ternary operator` if value / if not value`.

### Checklist

### Related issues
-

### And some more information
**IN:**
```py
import abc
import typing


class AbstractOperator(metaclass=abc.ABCMeta):
    @property
    @abc.abstractmethod
    def benchmark_code(self) -> str:
        """ """

    @property
    @abc.abstractmethod
    def benchmarks(self) -> typing.List[float]:
        """ """


class TernaryOperator(AbstractOperator):
    ternary_benchmarks: typing.List[float] = []

    @property
    def benchmarks(self) -> typing.List[float]:
        return self.ternary_benchmarks

    @property
    def benchmark_code(self) -> str:
        return """
a = True
if a:
    pass
"""


class IsOperator(AbstractOperator):
    is_benchmarks: typing.List[float] = []

    @property
    def benchmarks(self) -> typing.List[float]:
        return self.is_benchmarks

    @property
    def benchmark_code(self) -> str:
        return """
a = True
if a is True:
    pass
"""


if __name__ == "__main__":
    import collections
    import timeit

    class Result(collections.namedtuple("Result", ("cls", "time"))):
        """For better IDE support (pycharm doesn't see namedtuple attrs)."""

    class TypeResultsDict(collections.UserDict):
        def get_pare_by_cls_name(self, cls_name: str, /) -> typing.Optional[Result]:
            for instance, time in self.items():
                if instance.__class__.__name__ == cls_name:
                    return Result(instance, time)
            return None  # Returning `None` explicitly (mypy).

    def _make_benchmark(operator: AbstractOperator) -> float:
        return sum(timeit.timeit(operator.benchmark_code) for _ in range(100))

    def main() -> None:
        results = TypeResultsDict({})
        for operator in {TernaryOperator(), IsOperator()}:
            results[operator] = _make_benchmark(operator)

        # Time comparison.
        print(f"{'Time comparison':*^50}")
        print(*(f"{o.__class__.__name__} -> {r}" for o, r in results.items()), sep="\n")

        # Displaying the time difference.
        is_operator = results.get_pare_by_cls_name(IsOperator.__name__)
        ternary_operator = results.get_pare_by_cls_name(TernaryOperator.__name__)

        assert is_operator is not None
        assert ternary_operator is not None

        print(f"{'Time difference':*^50}")
        print(
            "is.time - ternary.time =",
            (diff := is_operator.time - ternary_operator.time),
        )
        print(f"The `ternary operator` is faster than the `is operator` by {diff}!")

    main()
```
**OUT:**
```py
*****************Time comparison******************
TernaryOperator -> 1.5785474000000004
IsOperator -> 1.8276172999999982
*****************Time difference******************
is.time - ternary.time = 0.24906989999999785
The `ternary operator` is faster than the `is operator` by 0.24906989999999785!
```
Screenshot_68

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -899,7 +899,7 @@ class RESTClientImpl(rest_api.RESTClient):
         body = await response.json()
         body_retry_after = float(body["retry_after"])
 
-        if body.get("global", False) is True:
+        if body.get("global", False):
             _LOGGER.error(
                 "rate limited on the global bucket. You should consider lowering the number of requests you make or "
                 "contacting Discord to raise this limit. Backing off and retrying request..."
@@ -1019,10 +1019,10 @@ class RESTClientImpl(rest_api.RESTClient):
         if isinstance(request_to_speak, datetime.datetime):
             body.put("request_to_speak_timestamp", request_to_speak.isoformat())
 
-        elif request_to_speak is True:
+        elif request_to_speak:
             body.put("request_to_speak_timestamp", time.utc_datetime().isoformat())
 
-        elif request_to_speak is False:
+        elif not request_to_speak:
             body.put("request_to_speak_timestamp", None)
 
         await self._request(route, json=body)
--- a/hikari/internal/data_binding.py
+++ b/hikari/internal/data_binding.py
@@ -210,9 +210,9 @@ class StringMapBuilder(multidict.MultiDict[str]):
             if conversion is not None:
                 value = conversion(value)
 
-            if value is True:
+            if value:
                 value = "true"
-            elif value is False:
+            elif value is False:  # The most optimal option in this case.
                 value = "false"
             elif value is None:
                 value = "null"
--- a/hikari/internal/mentions.py
+++ b/hikari/internal/mentions.py
@@ -69,20 +69,20 @@ def generate_allowed_mentions(
     parsed_mentions: typing.List[str] = []
     allowed_mentions: typing.Dict[str, typing.Any] = {"parse": parsed_mentions}
 
-    if mentions_everyone is True:
+    if mentions_everyone:
         parsed_mentions.append("everyone")
 
-    if mentions_reply is True:
+    if mentions_reply:
         allowed_mentions["replied_user"] = True
 
-    if user_mentions is True:
+    if user_mentions:
         parsed_mentions.append("users")
     elif isinstance(user_mentions, typing.Collection):
         # Duplicates will cause discord to error.
         ids = {str(int(u)) for u in user_mentions}
         allowed_mentions["users"] = list(ids)
 
-    if role_mentions is True:
+    if role_mentions:
         parsed_mentions.append("roles")
     elif isinstance(role_mentions, typing.Collection):
         # Duplicates will cause discord to error.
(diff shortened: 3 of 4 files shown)


hikari/internal/mentions.py
```diff
@@ -69,20 +69,20 @@ def generate_allowed_mentions(
...
-    if mentions_everyone is True:
+    if mentions_everyone:
         parsed_mentions.append("everyone")
 
-    if mentions_reply is True:
+    if mentions_reply:
         allowed_mentions["replied_user"] = True
 
-    if user_mentions is True:
+    if user_mentions:
```
tandemdude, line 78: `is True` is definitely needed here as it has significance when compared to the variable just being truthy (see the elif clause)

hikari/impl/rest.py
```diff
@@ -1019,10 +1019,10 @@ async def edit_my_voice_state(
         if isinstance(request_to_speak, datetime.datetime):
             body.put("request_to_speak_timestamp", request_to_speak.isoformat())
 
-        elif request_to_speak is True:
+        elif request_to_speak:
             body.put("request_to_speak_timestamp", time.utc_datetime().isoformat())
 
-        elif request_to_speak is False:
+        elif not request_to_speak:
```
davfsa, line 1025: Here the `True` and `False` checks differentiate from `undefined.UNDEFINED`

hikari/internal/mentions.py
Animatea: - [x] Fixed

hikari/internal/data_binding.py
```diff
@@ -210,9 +210,9 @@ def put(
             if conversion is not None:
                 value = conversion(value)
 
-            if value is True:
+            if value:
```
davfsa, line 213: Here we really need it to be `True`, not just truthy.

hikari/internal/mentions.py
```diff
@@ -69,20 +69,20 @@ def generate_allowed_mentions(
...
         parsed_mentions.append("users")
     elif isinstance(user_mentions, typing.Collection):
         # Duplicates will cause discord to error.
         ids = {str(int(u)) for u in user_mentions}
         allowed_mentions["users"] = list(ids)
 
-    if role_mentions is True:
+    if role_mentions:
```
davfsa: Same as what @tandemdude mentioned above

hikari/messages.py
```diff
@@ -1267,10 +1267,10 @@ async def respond(
         builtins.TypeError
             If both `attachment` and `attachments` are specified.
         """  # noqa: E501 - Line too long
-        if reply is True:
+        if reply:
             reply = self
 
-        elif reply is False:
+        elif not reply:
```
davfsa, line 1273: These seperate `bool` from `undefined.UNDEFINED`, so this actually creates erroneous behaviour

davfsa (commented):
Thank you for the PR. All these changes will turn into a small 1 or 2 change, since a lot of the checks really do need to ensure its `True` or `False`. Thank you for taking the time to do it tho!

PR #940, opened 2021-12-22
Implement member timeout

tandemdude:
### Summary
Implements member `communication_disabled_until` and the new `MODERATE_MEMBERS` permission.

### Checklist

- [x] Dav has made unittests according to the code I have added/modified/deleted.

### DIFF
--- a/changes/940.feature.md
+++ b/changes/940.feature.md
@@ -1,4 +1,4 @@
 Implement member timeouts
- - Add `communication_disabled_until` and `is_timed_out` to `Member`
+ - Add `raw_communication_disabled_until` and `communication_disabled_until` to `Member`
  - Add `MODERATE_MEMBERS` to `Permission`
  - Add `communication_disabled_until` attribute to `edit_member`
--- a/changes/940.feature.md
+++ b/changes/940.feature.md
@@ -1,0 +1,4 @@
+Implement member timeouts
+ - Add `communication_disabled_until` and `is_timed_out` to `Member`
+ - Add `MODERATE_MEMBERS` to `Permission`
+ - Add `communication_disabled_until` attribute to `edit_member`
--- a/hikari/guilds.py
+++ b/hikari/guilds.py
@@ -68,6 +68,7 @@ from hikari import users
 from hikari.internal import attr_extensions
 from hikari.internal import enums
 from hikari.internal import routes
+from hikari.internal import time
 
 if typing.TYPE_CHECKING:
     import datetime
@@ -79,7 +80,6 @@ if typing.TYPE_CHECKING:
     from hikari import permissions as permissions_
     from hikari import presences as presences_
     from hikari import voices as voices_
-    from hikari.internal import time
 
 
 @typing.final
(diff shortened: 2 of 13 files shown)


davfsa (approved):

hikari/guilds.py
```diff
@@ -487,6 +498,15 @@ def mention(self) -> str:
         """
         return f"<@!{self.id}>" if self.nickname is not None else self.user.mention
 
+    def is_timed_out(self) -> bool:
```
FasterSpeeding, line 501: Why not have a "raw_communication_disabled_until" field with a communication_disabled_until property where the property only returns the datetime if its not in the past so that you don't have to use both fields or just do the check for whether its in the past or not yourself to work out when their timeout expires?

hikari/guilds.py
```diff
@@ -487,6 +498,15 @@ def mention(self) -> str:
...
+        """Whether the member is timed out.
+
+        !!! note
+            The output of this function can depend based on when
```
FasterSpeeding, line 505: This could probably just be `` Unlike `raw_communication_disabled_until`, this will never be in the past. `` if that other style was used.

hikari/guilds.py
davfsa: Thommo and I had a discussion about this, the problem with the property is that people may rely on it and during the execution of the code, it might return None instead of the object you want. It is unlikely to happen, but it can.

Ie:

```py
if member.communication_disabled_until is not None:
    # Do something with it, but once its accessed again it may be None
```

Having a function call clearly denotes that it might change the next time it's called as well as provide a more helpful function

hikari/guilds.py
davfsa: This returns a boolean tho

FasterSpeeding: Read the 2nd part of what I said

hikari/guilds.py
FasterSpeeding: Then why can't communication_disabled_until be a function

hikari/guilds.py
davfsa: Sorry, completely missed that

davfsa:
Just as a note, `Member.edit` is missing the `communication_disabled_until` argument

hikari/guilds.py
davfsa: The reason I didnt like it is because it breaks a bit the standard if suddenly the way of accessing that attribute is through a function call. 

Would also like to get @tandemdude's opinion on this so we can find the best way to approach this issue

tandemdude: Personally I think I prefer the `is_x` style as it is more consistent with the other functions

FasterSpeeding: > The reason I didnt like it is because it breaks a bit the standard if suddenly the way of accessing that attribute is through a function call.

is_timed_out also has the name (and doc string funnily enough) of an attribute so

tandemdude: Actually thinking about it more I think I'm with @FasterSpeeding here it would probably make more sense to be a function retuning a datetime or None given the that other `is_x` members are actually properties

davfsa: Ok, then I guess we can change it to a function for it :+1:

davfsa (approved):

changes/940.feature.md
```diff
@@ -0,0 +1,4 @@
+Implement member timeouts
+ - Add `communication_disabled_until` and `is_timed_out` to `Member`
```
davfsa, line 2: This is no longer correct

davfsa (approved):

FasterSpeeding (approved):

PR #945, opened 2021-12-23
Cleanup typehints for event_manager

davfsa:
### Checklist


hikari/impl/event_manager_base.py
```diff
@@ -282,12 +280,12 @@ def consume_raw_event(
...
         self,
-        event_type: typing.Type[event_manager_.EventT_co],
-        callback: event_manager_.CallbackT[event_manager_.EventT_co],
+        event_type: typing.Type[typing.Any],
+        callback: event_manager_.CallbackT[typing.Any],
         *,
         _nested: int = 0,
     ) -> None:
-        if not issubclass(event_type, base_events.Event):
+        if isinstance(event_type, type) and not issubclass(event_type, base_events.Event):
```
davfsa, line 288:
```suggestion
        if not isinstance(event_type, type) or not issubclass(event_type, base_events.Event):
```

davfsa:
There is unfortunately a bug I found that Nekokatt tried to mitigate in the past in `subscribe` and `unsubscribe` by allowing `typing.Any` instead of taking in the `Event` class, then enforcing it at runtime. Unfortunately, we cant do this with `listen`, `wait_for` and `stream`, since we do need to provide the return type correctly. I have tried my best to find the best way to fix this to no luck. We will have to wait on these stale almost 4 year old issues.

Relevant issues: https://github.com/python/mypy/issues/5374 and https://github.com/python/mypy/issues/4717

hikari/api/event_manager.py
```diff
@@ -113,10 +110,10 @@ def open(self) -> None:
 
     @abc.abstractmethod
     def filter(
-        self: _T,
+        self,
         *predicates: typing.Union[typing.Tuple[str, typing.Any], typing.Callable[[EventT], bool]],
         **attrs: typing.Any,
-    ) -> _T:
+    ) -> EventStream[EventT]:
```
FasterSpeeding, line 116: Why was this changed

hikari/impl/event_manager_base.py
```diff
@@ -351,8 +349,8 @@ def get_listeners(
 
     def unsubscribe(
         self,
-        event_type: typing.Type[event_manager_.EventT_co],
-        callback: event_manager_.CallbackT[event_manager_.EventT_co],
+        event_type: typing.Type[typing.Any],
```
FasterSpeeding, line 352: These shouldn't be typed as Any

davfsa: They have been for a while, just in bot and not in the event manager. Explanation here: https://github.com/hikari-py/hikari/pull/945#issuecomment-1000555490

hikari/api/event_manager.py
davfsa: Removes an unnecessary typevar

FasterSpeeding: So for no reason?

davfsa: As the pr says, this is just to cleanup typehints

hikari/impl/event_manager_base.py
FasterSpeeding: I am well aware of that, this just hasn't ever been an issue for anybody in this context and I don't think progressively spreading shit type hints to places where it hasn't mattered to anybody to support a bug in a specific type checker is really a good thing to be doing.

hikari/api/event_manager.py
FasterSpeeding: well if you want to go and change all the places which return Self and are typed with typevars rather than just the actual type (since this pattern works better with inheritance) because apparently that's messier(?) then go ahead ig

FasterSpeeding: also in the future this'll pattern will be replaced by typing.Self but that's way far off

FasterSpeeding: But yeah changing this could cause issues for anybody who's extending this interface for cases like
```py
class ExtendedEventStream(EventStream[EventT]):
    @abc.abstractmethod
    async def foo(self) -> int:
        raise NotImplementedError
 
def do_thing(stream: ExtendedEventStream[MemberCreateEvent]) -> int:
    return await stream.filter(lambda e: e.user.id in {...}).foo()
```

since you're removing the guarantee that the return type will always be Self

davfsa: Didnt realize at the time how dumb this was to remove. Fixed :+1:

hikari/api/event_manager.py
```diff
@@ -30,27 +30,24 @@
...
-    PredicateT = typing.Callable[[EventT_co], bool]
-    CallbackT = typing.Callable[[EventT_inv], typing.Coroutine[typing.Any, typing.Any, None]]
+    PredicateT = typing.Callable[[base_events.EventT], bool]
+    CallbackT = typing.Callable[[base_events.EventT], typing.Coroutine[typing.Any, typing.Any, None]]
     ConsumerT = typing.Callable[
         [gateway_shard.GatewayShard, data_binding.JSONObject], typing.Coroutine[typing.Any, typing.Any, None]
     ]
 
-EventT = typing.TypeVar("EventT", bound="base_events.Event")
+    EventStreamT = typing.TypeVar("EventStreamT")
```
FasterSpeeding, line 47: Switching from a private `_T` type var to a public typevar called `EventStreamT` which isn't actually bound by EventStream seems like an odd change (even if it doesn't have to be bound for its use in this module)

davfsa: Sorry, started playing around yesterday trying to improve typehints more but no luck. This is a left-over from that, as I was using _T for something else. Will change back

FasterSpeeding: I don't think the last push actually changed it back

davfsa: Should be done now

PR #953, opened 2022-01-01
Prevent referenced messages from replacing cached messages

### LINKED ISSUE #904: Referenced message entries exceed the max messages settings

norinorin:
### Checklist

### Related issues
Closes #904

### DIFF
--- a/hikari/impl/cache.py
+++ b/hikari/impl/cache.py
@@ -1560,7 +1560,15 @@ class CacheImpl(cache.MutableCache):
 
         referenced_message: typing.Optional[cache_utility.RefCell[cache_utility.MessageData]] = None
         if message.referenced_message:
-            referenced_message = self._set_message(message.referenced_message)
+            if referenced_message := self._message_entries.get(
+                message.referenced_message.id
+            ) or self._referenced_messages.get(message.referenced_message.id):
+                # treat referenced messages as partial
+                # so it shouldn't replace the full message if cached.
+                referenced_message.object.update(message.referenced_message)
+                message.referenced_message = self._build_message(referenced_message)
+            else:
+                referenced_message = self._set_message(message.referenced_message)
 
         # Only increment ref counts if this wasn't previously cached.
         if message.id not in self._referenced_messages and message.id not in self._message_entries:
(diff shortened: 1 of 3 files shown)


hikari/impl/cache.py
```diff
@@ -1560,7 +1560,15 @@ def _set_message(
 
         referenced_message: typing.Optional[cache_utility.RefCell[cache_utility.MessageData]] = None
         if message.referenced_message:
-            referenced_message = self._set_message(message.referenced_message)
+            if referenced_message := self._message_entries.get(
+                message.referenced_message.id
+            ) or self._referenced_messages.get(message.referenced_message.id):
+                # treat referenced messages as partial
+                # so it shouldn't replace the full message if cached.
+                referenced_message.object.update(message.referenced_message)
```
norinorin, line 1568: I'm unsure if updating it is really needed.

norinorin:
Not sure if I like it; will look for other possible solutions.

norinorin:
Closing as I'll leave it stale otherwise.

PR #958, opened 2022-01-09
Add new message content related application flags

FasterSpeeding:
### Checklist

### DIFF
--- a/hikari/applications.py
+++ b/hikari/applications.py
@@ -89,10 +89,10 @@ class ApplicationFlags(enums.Flag):
     EMBEDDED = 1 << 17
     """Denotes that the application has functionality that's specially embedded in Discord's client."""
 
-    MESSAGE_CONTENT = 1 << 18
+    MESSAGE_CONTENT_INTENT = 1 << 18
     """Denotes that the application has message content intent enabled in it's dashboard."""
 
-    MESSAGE_CONTENT_LIMITED = 1 << 19
+    MESSAGE_CONTENT_INTENT_LIMITED = 1 << 19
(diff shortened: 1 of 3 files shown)


hikari/applications.py
```diff
@@ -89,6 +89,12 @@ class ApplicationFlags(enums.Flag):
     EMBEDDED = 1 << 17
     """Denotes that the application has functionality that's specially embedded in Discord's client."""
 
+    GATEWAY_MESSAGE_CONTENT = 1 << 18
```
FasterSpeeding, line 92: Not sure whether these should be called GATEWAY_MESSAGE_CONTENT* or just MESSAGE_CONTENT* rn

davfsa: I suspect it should be `MESSAGE_CONTENT*`, as it wouldn't make sense for them to not send the information in the gateway but yes in the REST API.

hikari/applications.py
```diff
@@ -89,6 +89,12 @@ class ApplicationFlags(enums.Flag):
...
+    """Denotes that the application has the GATEWAY_MESSAGE_CONTENT intent enabled in it's dashboard."""
+
+    GATEWAY_MESSAGE_CONTENT_LIMITED = 1 << 19
+    """Denotes that the application has message content access while pending verification."""
```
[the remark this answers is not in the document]
FasterSpeeding: VERIFIED_FOR_GUILD_MEMBERS_INTENT has nothing to do with message content, it just gives you a slightly higher guild cap, it's not a blanket flag

davfsa (approved):

PR #967, opened 2022-01-13
feat: add `locale` and `guild_locale` properties to interactions

Victorsitou:
### Summary

Discord announced its release a few days ago.

1. `locale` will always be present (except PING, which is responded with pong).
2. `guild_locale` will only be present in guild context interactions, and defaults to `en-US` if the guild doesn't have `COMMUNITY` feature.

### Checklist


hikari/interactions/base_interactions.py
```diff
@@ -183,6 +183,15 @@ class PartialInteraction(snowflakes.Unique, webhooks.ExecutableWebhook):
     version: int = attr.field(eq=False, repr=True)
     """Version of the interaction system this interaction is under."""
 
+    locale: str = attr.field(eq=False, repr=True)
```
FasterSpeeding, line 186: While the lib doesn't use PartialInteraction like this in the standard impl right now, PartialInteraction is meant to be able to support Ping interactions so these should probably be added to each sub-class instead of the base class

Victorsitou: Should be resolved in this commit https://github.com/hikari-py/hikari/pull/967/commits/16dd451e99d1354d97be3245b4f454b2587a4894

hikari/interactions/command_interactions.py
```diff
@@ -145,6 +145,12 @@ class CommandInteraction(base_interactions.MessageResponseMixin[CommandResponseT
     This will be `builtins.None` for command interactions triggered in DMs.
     """
 
+    guild_locale: typing.Optional[str] = attr.field(eq=False, hash=False, repr=True)
+    """The preferred language of the guild this command interaction was triggered in.
```
FasterSpeeding, line 149: Can we document the fact that guild_locale defaults to en-us and can only be set by community guilds here as a note section, IMO this is pretty important info to communicate

FasterSpeeding:
Other than that, LGTM

FasterSpeeding (approved):

PR #971, opened 2022-01-19
Add ability to send attachments in an interaction initial response

tandemdude:
### Summary
Adds the ability to pass attachments with `create_initial_response`

### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1218,7 +1218,7 @@ class RESTClientImpl(rest_api.RESTClient):
         embed: undefined.UndefinedNoneOr[embeds_.Embed] = undefined.UNDEFINED,
         embeds: undefined.UndefinedNoneOr[typing.Sequence[embeds_.Embed]] = undefined.UNDEFINED,
         replace_attachments: bool = False,
-        flags: undefined.UndefinedOr[messages_.MessageFlag] = undefined.UNDEFINED,
+        flags: typing.Union[undefined.UndefinedType, int, messages_.MessageFlag] = undefined.UNDEFINED,
         tts: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
         mentions_everyone: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
         mentions_reply: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
@@ -1420,7 +1420,7 @@ class RESTClientImpl(rest_api.RESTClient):
         role_mentions: undefined.UndefinedOr[
             typing.Union[snowflakes.SnowflakeishSequence[guilds.PartialRole], bool]
         ] = undefined.UNDEFINED,
-        flags: undefined.UndefinedOr[messages_.MessageFlag] = undefined.UNDEFINED,
+        flags: typing.Union[undefined.UndefinedType, int, messages_.MessageFlag] = undefined.UNDEFINED,
     ) -> messages_.Message:
         route = routes.PATCH_CHANNEL_MESSAGE.compile(channel=channel, message=message)
         body, form = self._build_message_payload(
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1320,7 +1320,7 @@ class RESTClientImpl(rest_api.RESTClient):
         else:
             body.put("content", None)
 
-        if edit and replace_attachments:
+        if replace_attachments:
             body.put("attachments", None)
 
         body.put("tts", tts)
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1309,12 +1309,7 @@ class RESTClientImpl(rest_api.RESTClient):
                     final_attachments.extend(embed_attachments)
                     serialized_embeds.append(embed_payload)
 
-        if edit and not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
-            body.put(
-                "allowed_mentions",
-                mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
-            )
-        else:
+        if not edit or not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
             body.put(
                 "allowed_mentions",
                 mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1204,9 +1204,8 @@ class RESTClientImpl(rest_api.RESTClient):
         assert isinstance(response, dict)
         return self._entity_factory.deserialize_message(response)
 
-    def _build_message(
+    def _build_message_payload(  # noqa: C901
         self,
-        body: data_binding.JSONObjectBuilder,
         *,
         content: undefined.UndefinedOr[typing.Any] = undefined.UNDEFINED,
         attachment: undefined.UndefinedOr[files.Resourceish] = undefined.UNDEFINED,
@@ -1228,7 +1227,9 @@ class RESTClientImpl(rest_api.RESTClient):
         ] = undefined.UNDEFINED,
         edit: bool = False,
         replace_attachments: bool = False,
-    ) -> typing.Optional[data_binding.URLEncodedFormBuilder]:
+    ) -> typing.Tuple[data_binding.JSONObjectBuilder, typing.Optional[data_binding.URLEncodedFormBuilder]]:
+        body = data_binding.JSONObjectBuilder()
+
         if not undefined.any_undefined(attachment, attachments):
             raise ValueError("You may only specify one of 'attachment' or 'attachments', not both")
 
@@ -1308,29 +1309,24 @@ class RESTClientImpl(rest_api.RESTClient):
                     final_attachments.extend(embed_attachments)
                     serialized_embeds.append(embed_payload)
 
-        if edit:
-            # Special cases for message editing
-            if not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
-                # If specified we override the message's allowed mentions, else
-                # whatever the current value is will be used
-                body.put(
-                    "allowed_mentions",
-                    mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
-                )
-            if content is not None:
-                body.put("content", content, conversion=str)
-            else:
-                # None here means we want to remove the content from the message
-                body.put("content", None)
-            if replace_attachments:
-                # We are removing embeds and/or replacing them with new ones
-                body.put("attachments", None)
+        if edit and not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
+            body.put(
+                "allowed_mentions",
+                mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
+            )
         else:
             body.put(
                 "allowed_mentions",
                 mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
             )
+
+        if content is not None:
             body.put("content", content, conversion=str)
+        else:
+            body.put("content", None)
+
+        if edit and replace_attachments:
+            body.put("attachments", None)
 
         body.put("tts", tts)
 
@@ -1342,9 +1338,9 @@ class RESTClientImpl(rest_api.RESTClient):
 
             for i, attachment in enumerate(final_attachments):
                 form.add_resource(f"file{i}", attachment)
-            return form
+            return body, form
 
-        return None
+        return body, None
 
     async def create_message(
         self,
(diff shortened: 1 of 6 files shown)


FasterSpeeding:
I'ma see if i can work on the rest server equivalent logic for this today

FasterSpeeding:
Btw I do have a prototype impl of sending attachments on rest but i'm going to be waiting for discord to push a change they said they'd be pushing soon:tm: so if you remove the interaction server from this pr and switch out to handling de-duplicating the logic by moving out the payload building logic to a separate function instead of making the create_message logic work for create initial response this can be merged without the server logic

(and i'll do the server logic in another pr)

FasterSpeeding:
Sry for the delay

FasterSpeeding:
For a bit more specific detail on how OI think this should be handled: instead of modifying `_create_message` to also handle initial response specific logic I think we would be better off taking the logic for building the message response payload and doing something like
```py
def _build_message(self, message_body: JSONObjectBuilder, ...) -> Optional[URLEncodedFormBuilder]:
```
where _build_message should cover both edit and create semantics, and if it returns a `URLEncodedFormBuilder` then the calling logic should handle inserting the `"payload_json"` field before making the request

tandemdude:
Tests are still a TODO but a pre-emptive review would be welcomed @FasterSpeeding 😉

hikari/impl/rest.py
```diff
@@ -1273,46 +1278,73 @@ async def _create_message(
...
+            # Special cases for message editing
+            if not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
+                # If specified we override the message's allowed mentions, else
+                # whatever the current value is will be used
+                body.put(
+                    "allowed_mentions",
+                    mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
+                )
+            if content is not None:
+                body.put("content", content, conversion=str)
```
FasterSpeeding, line 1321: the None checks can just be run both bot edits and creates for simplicity

hikari/impl/rest.py
```diff
@@ -1273,46 +1278,73 @@ async def _create_message(
...
+            if embeds is not None:
+                serialized_embeds = []
+                for embed in embeds:
+                    embed_payload, embed_attachments = self._entity_factory.serialize_embed(embed)
+                    final_attachments.extend(embed_attachments)
+                    serialized_embeds.append(embed_payload)
+
+        if edit:
+            # Special cases for message editing
+            if not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
```
FasterSpeeding, line 1313: Since the None checks don't need to be in here, this could be simplified to

```py
if not edit:
    ....

elif not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
    ...
```

hikari/impl/rest.py
```diff
@@ -3512,24 +3547,34 @@ async def edit_interaction_response(
         ] = undefined.UNDEFINED,
     ) -> messages_.Message:
         route = routes.PATCH_INTERACTION_RESPONSE.compile(webhook=application, token=token)
-        return await self._edit_message(
-            route,
-            data_binding.JSONObjectBuilder(),
-            no_auth=True,
+
```
FasterSpeeding, line 3550: Some of this spacing seems a bit unnecessary (although the spaces around the `if/if else` statement is fine)

hikari/impl/rest.py
```diff
@@ -1204,26 +1204,31 @@ async def fetch_message(
         assert isinstance(response, dict)
         return self._entity_factory.deserialize_message(response)
 
-    async def _create_message(
+    def _build_message(
```
davfsa, line 1207:
```suggestion
    def _build_message_payload(
```

Just for a little more clarity

hikari/impl/rest.py
```diff
@@ -1204,26 +1204,31 @@ async def fetch_message(
...
+        mentions_reply: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
+        user_mentions: undefined.UndefinedOr[
+            typing.Union[snowflakes.SnowflakeishSequence[users.PartialUser], bool]
+        ] = undefined.UNDEFINED,
+        role_mentions: undefined.UndefinedOr[
+            typing.Union[snowflakes.SnowflakeishSequence[guilds.PartialRole], bool]
+        ] = undefined.UNDEFINED,
+        edit: bool = False,
+        replace_attachments: bool = False,
+    ) -> typing.Optional[data_binding.URLEncodedFormBuilder]:
```
davfsa, line 1231: Just because this might confuse people later
```suggestion
    ) -> typing.Optional[data_binding.URLEncodedFormBuilder]:
    # This function will build the payload on the passed `body` and then return a `URLEncodedFormBuilder`
    # if it should be used instead
```

*Only commit this if you dont agree with the other comment I made bellow*

hikari/impl/rest.py
```diff
@@ -1273,46 +1278,73 @@ async def _create_message(
...
                 form.add_resource(f"file{i}", attachment)
+            return form
 
-            response = await self._request(route, form_builder=form, query=query, no_auth=no_auth)
-        else:
-            response = await self._request(route, json=body, query=query, no_auth=no_auth)
-
-        assert isinstance(response, dict)
-        return self._entity_factory.deserialize_message(response)
+        return None
```
davfsa, line 1347: Instead of returning the form, could this be changed to actually do the request and then return whatever `_request` returns. The logic for deserializing the message can be moved outside. The interface will be the same as if you were doing `_request` but with it handling all the logic

FasterSpeeding: not all endpoints are putting the message body in the in the form builder under the key "payload_json" so that'd leak endpoint specific impl detail into this code which its meant to avoid

hikari/impl/rest.py
```diff
@@ -1527,8 +1568,8 @@ async def edit_message(
         route = routes.PATCH_CHANNEL_MESSAGE.compile(channel=channel, message=message)
         body = data_binding.JSONObjectBuilder()
         body.put("flags", flags)
-        return await self._edit_message(
```
FasterSpeeding, line 1530: You forgot to remove the _edit_message definition

hikari/impl/rest.py
thomm-o: Just what I was about to type lol 😅

davfsa: It looks like they all are tho

- https://github.com/hikari-py/hikari/pull/971/files#diff-9904e7f903a912bfe485f6f673fcd1e42da8b90786d58fb612077b2c42b68bfaR1394
- https://github.com/hikari-py/hikari/pull/971/files#diff-9904e7f903a912bfe485f6f673fcd1e42da8b90786d58fb612077b2c42b68bfaR1590
- https://github.com/hikari-py/hikari/pull/971/files#diff-9904e7f903a912bfe485f6f673fcd1e42da8b90786d58fb612077b2c42b68bfaR1921
- https://github.com/hikari-py/hikari/pull/971/files#diff-9904e7f903a912bfe485f6f673fcd1e42da8b90786d58fb612077b2c42b68bfaR1989
- https://github.com/hikari-py/hikari/pull/971/files#diff-9904e7f903a912bfe485f6f673fcd1e42da8b90786d58fb612077b2c42b68bfaR3521
- https://github.com/hikari-py/hikari/pull/971/files#diff-9904e7f903a912bfe485f6f673fcd1e42da8b90786d58fb612077b2c42b68bfaR3570

thomm-o: `create_initial_response` adds message to a nested `data` attribute iirc

thomm-o: I don't think its the form data that was the problem it was the actual json body

FasterSpeeding:
Meow

HyperGH:
Woof

hikari/impl/rest.py
```diff
@@ -1273,46 +1279,68 @@ async def _create_message(
...
+                mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
+            )
+
+        if content is not None:
+            body.put("content", content, conversion=str)
+        else:
+            body.put("content", None)
+
+        if edit and replace_attachments:
+            body.put("attachments", None)
```
davfsa, line 1324: Tbh, this could be left as
```suggestion
        if replace_attachments:
            body.put("attachments", None)
```

hikari/impl/rest.py
```diff
@@ -1273,46 +1279,68 @@ async def _create_message(
...
+        if edit and not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
+            body.put(
+                "allowed_mentions",
+                mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
+            )
+        else:
+            body.put(
+                "allowed_mentions",
+                mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
+            )
```
davfsa: This can be simplified to:
```suggestion
        if not edit or not undefined.all_undefined(mentions_everyone, mentions_reply, user_mentions, role_mentions):
            body.put(
                "allowed_mentions",
                mentions.generate_allowed_mentions(mentions_everyone, mentions_reply, user_mentions, role_mentions),
            )
```

hikari/impl/rest.py
```diff
@@ -1204,26 +1204,32 @@ async def fetch_message(
         assert isinstance(response, dict)
         return self._entity_factory.deserialize_message(response)
 
-    async def _create_message(
+    def _build_message_payload(  # noqa: C901
         self,
-        route: routes.CompiledRoute,
-        body: data_binding.JSONObjectBuilder,
-        query: typing.Optional[data_binding.StringMapBuilder] = None,
         *,
```
davfsa, line 1209: Just for the sake of it
```suggestion
        /,
        *,
```

hikari/impl/rest.py
```diff
@@ -1253,15 +1259,15 @@ async def _create_message(
 
         if undefined.all_undefined(embed, embeds) and isinstance(content, embeds_.Embed):
```
FasterSpeeding, line 1260: What about if embed or embeds is None

hikari/impl/rest.py
```diff
@@ -1253,15 +1259,15 @@ async def _create_message(
...
             # Syntactic sugar, common mistake to accidentally send an embed
-            # as the content, so lets detect this and fix it for the user.
+            # as the content, so let's detect this and fix it for the user.
             embed = content
             content = undefined.UNDEFINED
 
         elif undefined.all_undefined(attachment, attachments) and isinstance(
```
FasterSpeeding, line 1266: What about if attachment or attachments is None

hikari/impl/rest.py
```diff
@@ -1270,49 +1276,64 @@ async def _create_message(
         final_attachments: typing.List[files.Resource[files.AsyncReader]] = []
         if attachment is not undefined.UNDEFINED:
             final_attachments.append(files.ensure_resource(attachment))
-        if attachments is not undefined.UNDEFINED:
+        elif attachments is not undefined.UNDEFINED:
```
FasterSpeeding, line 1279: btw changing this to an elif is kinda unnecessary since an earlier check already asserts that only one is not undefined

hikari/impl/rest.py
davfsa: We dont allow attachments to be None atm, but this is syntactic sugar for passing it to the content. If they want to pass any of those as None, they can just pass it to the argument

FasterSpeeding: Sry i put this on the wrong resource

hikari/impl/rest.py
FasterSpeeding: Also I'm not inferring this is bad behaviour, i'm just asking cause it seems like this is something that previously slipped through the gap without any properconsideration

davfsa: This is also the same as the other comment bellow. This is the syntactic sugar. If they want to pass embed as None, they can do `embed=None` or `embeds=None`. I found while testing that some parts of the code (and typing) said that `component` and `components` could be `None`, but then the check didnt allow it, which I fixed:

https://github.com/hikari-py/hikari/pull/971/files#diff-9904e7f903a912bfe485f6f673fcd1e42da8b90786d58fb612077b2c42b68bfaR1248

FasterSpeeding: Yeah but with the current behaviour passing an embed to content then None to embed or embeds leads to the Nones being ignored which is why I asked about the semantics

davfsa: Honestly, I would go with "intended behavior" then. The documentation clearly specifies that the syntactic sugar only triggers if the other arguments are not passed:
```
        content : hikari.undefined.UndefinedOr[typing.Any]
            If provided, the message contents. If
            `hikari.undefined.UNDEFINED`, then nothing will be sent
            in the content. Any other value here will be cast to a
            `builtins.str`.

            If this is a `hikari.embeds.Embed` and no `embed` nor
            no `embeds` kwarg is provided, then this will instead
            update the embed. This allows for simpler syntax when
            sending an embed alone.

            Likewise, if this is a `hikari.files.Resource`, then the
            content is instead treated as an attachment if no `attachment` and
            no `attachments` kwargs are provided.
 ```

**Review by FasterSpeeding — APPROVED** [review]

File: hikari/impl/rest.py [hunk 14]
```diff
@@ -1253,15 +1259,15 @@ async def _create_message(
...
+            # as the content, so let's detect this and fix it for the user.
             embed = content
             content = undefined.UNDEFINED
 
         elif undefined.all_undefined(attachment, attachments) and isinstance(
             content, (files.Resource, files.RAWISH_TYPES, os.PathLike)
         ):
             # Syntactic sugar, common mistake to accidentally send an attachment
-            # as the content, so lets detect this and fix it for the user. This
+            # as the content, so let's detect this and fix it for the user. This
```
FasterSpeeding, line 1270: also its lets not let's, let's would be let is

tandemdude: Here it is actually a contraction of "let us" so is correct

FasterSpeeding: ah ok

PR #974, opened 2022-01-22
Fix error: cannot be converted to 'PyString'

HazemMeqdad:
### Summary
Fix the TypeError: 'int' object cannot be converted to 'PyString'

### Checklist

### DIFF
--- a/examples/simple_dashboard/simple_dashboard.py
+++ b/examples/simple_dashboard/simple_dashboard.py
@@ -58,7 +58,7 @@ rillrate.install()
 bot = Bot(token=os.environ["BOT_TOKEN"])
 
 
-values = list(range(0, 256 + 1, 32))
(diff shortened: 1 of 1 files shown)


examples/simple_dashboard/simple_dashboard.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, **kwargs: Any) -> None:
 bot = Bot(token=os.environ["BOT_TOKEN"])
 
 
-values = list(range(0, 256 + 1, 32))
+values = [str(i) for i in list(range(0, 256 + 1, 32))]
```
davfsa, line 61:
```suggestion
values = [str(i) for i in range(0, 256 + 1, 32)]
```

davfsa (approved):
Ty!

PR #924, opened 2021-11-30
Add context menu commands and autocomplete

thesadru:
```
pip install git+https://github.com/thesadru/hikari@feature/interactions
```

### Summary

- Add user commands and message commands
- Allow for application command autocomplete

Please merge without keeping every single commit, I spammed them a lot ^^.

### Checklist

- [x] I have run `nox` and all the pipelines have passed.
- [x] I have made unittests according to the code I have added/modified/deleted.
- [x] Break backwards compatibility.

### Related issues
closes #879
closes #794

PR #976, opened 2022-01-23
make `is_required` optional

Lunarmagpie:
### Checklist

### DIFF
--- a/hikari/commands.py
+++ b/hikari/commands.py
@@ -124,8 +124,10 @@ class CommandOption:
         This will be inclusively between 1-100 characters in length.
     """
 
-    is_required: bool = attr.field(repr=False)
-    """Whether this command """
+    is_required: typing.Optional[bool] = attr.field(repr=False)
+    """Whether this command is required. This will be `builtins.None` if the option type
+    is `SUB_COMMAND` or `SUB_COMMAND_GROUP`.
+    """
 
     choices: typing.Optional[typing.Sequence[CommandChoice]] = attr.field(default=None, repr=False)
     """A sequence of up to (and including) 25 choices for this command.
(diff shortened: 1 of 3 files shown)


hikari/commands.py
```diff
@@ -124,8 +124,10 @@ class CommandOption:
         This will be inclusively between 1-100 characters in length.
     """
 
-    is_required: bool = attr.field(repr=False)
-    """Whether this command """
+    is_required: typing.Optional[bool] = attr.field(repr=False)
+    """Whether this command is required. This will be `builtins.None` if the option type
+    is `SUB_COMMAND` or `SUB_COMMAND_GROUP`.
+    """
```
davfsa, line 130:
```suggestion
    is_required: typing.Optional[bool] = attr.field(repr=False)
    """Whether this command is required.
    
    This will be `builtins.None` if the option type is `SUB_COMMAND`
    or `SUB_COMMAND_GROUP`.
    """
```

davfsa (commented):
This documentation change is a bit contradictory, as when deserializing the object, this will never be correct, as it will only deserialize into a bool. Probably should be changed there too to account for this, but I am not sure if I am too happy making a special case for those

Lunarmagpie:
I see what you mean about the documentation changes. After looking over the discord docs again I think its best to keep it deserializing to false.

thesadru:
Pretty sure #924 already does this.

Lunarmagpie:
The description is changed but the `Optional` type hint isn't added

davfsa:
> The description is changed but the `Optional` type hint isn't added

Not too sure I like having it be Optional without correctly documenting why. Will wait on @FasterSpeeding's opinion

hikari/commands.py
FasterSpeeding: That isn't how the API object works, not present means false for all command option types as per the docs and it'll  never be returned as null. this should just be defaulting to False as it is in #924

FasterSpeeding:
Closed in favour of the changes in #924

PR #986, opened 2022-01-29
Merge ed25519 dependencies to an optional server feature flag

FasterSpeeding:
### Summary
This removes the old pure ed25519 and ed25519 dependencies and replaces them with one optional pynacl dependency which is installed with the "server" feature flag.

The goal being to stream-line ed25519 handling, remove dependency on less-well maintained projects (as ed25519 is currently blocking py 3.11 support) and remove dependency on pure ed25519 which from a technical cryptographic perspective hasn't been "validated" like libsodium

### Checklist


hikari/impl/interaction_server.py
```diff
@@ -189,7 +206,10 @@ def is_alive(self) -> bool:
         """
         return self._server is not None
 
-    async def _fetch_public_key(self) -> ed25519.VerifierT:
+    async def _fetch_public_key(self) -> signing.VerifyKey:
+        # This is kept inline as pynacl is an optional dependency.
+        from nacl import signing
```
davfsa, line 211: This will import it every time we want to use it. Instead, should we try and import it when we load the file and then set a flag to know if it has been imported or not, which we can later use:

```py
try:
    from nacl import signing
except ModuleNotFoundError:
    signing = None
```

And then some checks can be set before its used

hikari/impl/interaction_server.py
```diff
@@ -176,7 +189,11 @@ def __init__(
         self._loads = loads
         self._rest_client = rest_client
         self._server: typing.Optional[aiohttp.web_runner.AppRunner] = None
-        self._verify = ed25519.build_ed25519_verifier(public_key) if public_key is not None else None
+
+        # if not isinstance(public_key, bytes) or len(public_key) != 32:
+        #     raise ValueError("public_key must be a 32-byte sequence.")
```
davfsa, line 194: Any reason for this being commented out?

hikari/impl/interaction_server.py
```diff
@@ -165,6 +169,15 @@ def __init__(
...
         public_key: typing.Optional[bytes] = None,
     ) -> None:
+        # This is kept inline as pynacl is an optional dependency.
+        try:
+            from nacl import signing
+
+        except ModuleNotFoundError as exc:
+            raise RuntimeError(
+                "You must install the optional `hikari[server]` dependencies to use the default interaction server."
+            ) from exc
```
davfsa, line 179: Same as the other comment

tests/hikari/impl/test_interaction_server.py
```diff
@@ -21,21 +21,61 @@
 # SOFTWARE.
 import asyncio
 import contextlib
+import typing
```
davfsa, line 24: Redundant import?

tests/hikari/impl/test_interaction_server.py
```diff
@@ -115,6 +155,19 @@ def test___init___without_public_key(self, mock_verifier, mock_rest_client, mock
...
+        [
+            "okokokokokokokokokokokokokokokok",
+            b"NO",
+        ],
+    )
+    def test___init___with_invalid_public_key(self, key: typing.Union[str, bytes]):
+        with mock.patch.object(aiohttp.web, "Application"):
+            interaction_server_impl.InteractionServer(
+                dumps=object(), entity_factory=object(), loads=object(), rest_client=object(), public_key=key
+            )
```
davfsa, line 169: There are no checks here. Is this test incomplete?

hikari/impl/interaction_server.py
FasterSpeeding: Because i haven't poked around how pynacl handles those cases itself yet

tests/hikari/impl/test_interaction_server.py
FasterSpeeding: the tests haven't been done yet

tests/hikari/impl/test_interaction_server.py
FasterSpeeding: typing isn't being imported anywhere else there, any new tests I'm writing will be typed btw

hikari/impl/interaction_server.py
FasterSpeeding: Mypy wouldn't allow that.  also this isn't hot code, with the current approach signing is imported 1-2 times per bot that's initiated and then run 1+ times

davfsa: Fair enough

hikari/impl/interaction_server.py
```diff
@@ -189,7 +205,10 @@ def is_alive(self) -> bool:
         """
         return self._server is not None
 
-    async def _fetch_public_key(self) -> ed25519.VerifierT:
+    async def _fetch_public_key(self) -> signing.VerifyKey:
+        # This is kept inline as pynacl is an optional dependency.
+        from nacl import signing
```
FasterSpeeding, line 210: instead of re-importing this i'd be safer to just assign signing to an attribute in the init

hikari/impl/interaction_server.py
FasterSpeeding: Oh yeah another reason why I prefer this style is that we're not erasing the source error since silently throwing away the raised ModuleNotFoundError instead of using it as the cause would make it take more difficult/effort to debug any cases where it wasn't importing despite being supposedly installed (e.g. one of its own dependencies is no-longer installed).

hikari/impl/interaction_server.py
```diff
@@ -206,25 +203,23 @@ def is_alive(self) -> bool:
...
 
             if self._rest_client.token_type == applications.TokenType.BOT:
                 application = await self._rest_client.fetch_application()
 
             else:
                 application = (await self._rest_client.fetch_authorization()).application
 
-            self._verify = signing.VerifyKey(application.public_key)
-            return self._verify
+            print(application.public_key)
```
davfsa: Print left here

hikari/impl/interaction_server.py
```diff
@@ -186,13 +186,10 @@ def __init__(
         self._is_closing = False
         self._listeners: typing.Dict[typing.Type[base_interactions.PartialInteraction], typing.Any] = {}
         self._loads = loads
+        self._nacl = nacl
```
davfsa, line 189: Idk if I like this tbh. I am going to have a look at an idea I had and see if mypy would be happy about it and come back to this

hikari/impl/interaction_server.py
```diff
@@ -45,8 +45,7 @@
 
     import aiohttp.typedefs
 
-    # This is only imported here for internal type-hints and otherwise needs
-    # to be inline imported as it's an optional dependency.
+    # This is kept inline as pynacl is an optional dependency.
     from nacl import signing
```
davfsa, line 49: Wouldnt this make typechecking fail if they dont have the dependency installed?

FasterSpeeding: Wdym make type checking fail? Also it's only used internally

hikari/impl/interaction_server.py
FasterSpeeding: That print isn't there any more, you reviewed an old commit

hikari/impl/interaction_server.py
```diff
@@ -165,6 +167,16 @@ def __init__(
...
     ) -> None:
+        # This is kept inline as pynacl is an optional dependency.
+        try:
+            import nacl.exceptions
+            import nacl.signing
+
+        except ModuleNotFoundError as exc:
+            raise RuntimeError(
+                "You must install the optional `hikari[server]` dependencies to use the default interaction server."
+            ) from exc
```
davfsa, line 178: Coverage is missing here

davfsa (commented):
Just moving over comments from Discord so I dont forget. Everything else looks alright, just the small comment

tests/hikari/impl/test_interaction_server.py
```diff
@@ -61,6 +153,18 @@ def test_when_content_type_provided(self):
         assert response.status_code == 201
 
 
+@pytest.mark.skipif(nacl_present, reason="PyNacl is present")
+def test_interaction_server_init_when_no_pynacl():
+    with pytest.raises(
+        RuntimeError,
+        match=re.escape(
+            "You must install the optional `hikari[server]` dependencies to use the default interaction server."
+        ),
```
davfsa, line 162:
```suggestion
        match=r"You must install the optional `hikari[server]` dependencies to use the default interaction server."
```
This could also be simplified to this.

tests/hikari/impl/test_interaction_server.py
```diff
@@ -21,21 +21,113 @@
 # SOFTWARE.
 import asyncio
 import contextlib
+import re
```
davfsa, line 24:
```suggestion
```
Complementary to other comment

davfsa (approved):
Since automerge is disabled, left 2 little comments that I dont mind having, just small things.

Ty for the pr tho :+1:

PR #987, opened 2022-01-29
Remove pages and safety from the default nox sessions

FasterSpeeding:
### Summary
Mostly just a suggestion since this doesn't seem helpful to most ppl running it locally and under the rare case that their changes break doc gen this'll still be flagged up by pipelines when they open a PR

### Checklist

### DIFF
--- a/pipelines/nox.py
+++ b/pipelines/nox.py
@@ -32,7 +32,7 @@ from nox.sessions import Session
 from pipelines import config
 
 # Default sessions should be defined here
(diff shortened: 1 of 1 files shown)


FasterSpeeding:
Just a thought, is it really important that safety is run by default locally for ppl?

Most people aren't going to be adding dependencies and safety can error on environment specific factors like the installed pip version which we won't necessarily want to be accounting for and this could even just cause unnecessary confusion.
(while we currently have a work around for avoiding it erroring on old pip installs, that's only temporary and should probably be removed at some point)

davfsa (approved):

PR #996, opened 2022-02-02
Improve deprecation notices for rest cmd methods

FasterSpeeding:
### Summary
This should better clarify that the deprecated methods still only create slash command (builders) and only point towards the relevant new slash command method.

### Checklist


davfsa (commented):
The decorators should be edited too. Ty tho!

davfsa (approved):

PR #999, opened 2022-02-03
Add startup and shutdown callbacks to the RESTBot interface

FasterSpeeding:
### Summary
Add startup and shutdown callbacks to the RESTBot interface

### Checklist

### DIFF
--- a/hikari/errors.py
+++ b/hikari/errors.py
@@ -28,7 +28,6 @@ __all__: typing.Sequence[str] = (
     "HikariError",
     "HikariWarning",
     "HikariInterrupt",
-    "MultiError",
     "ComponentStateConflictError",
     "UnrecognisedEntityError",
     "NotFoundError",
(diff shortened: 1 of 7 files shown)


hikari/errors.py
```diff
@@ -90,6 +91,44 @@ class HikariWarning(RuntimeWarning):
     """
 
 
+@attr.define(auto_attribs=False, auto_exc=True, repr=False, weakref_slot=False)
+class MultiError(HikariError):
```
FasterSpeeding, line 95: @davfsa not sure if you'll be a fan of this but I'd rather keep the error more generic rn rather than case specific

davfsa (commented):
I just realized I forgot to submit the review for this pr, so my comment has just been sitting in a limbo state for a bit too long :P

hikari/errors.py
FasterSpeeding: I mean it does show the error raised? They're all in the repr

FasterSpeeding: And what if they catch and handle the error? You've now just spammed their console with a ton of rubbish theyd handled

FasterSpeeding: You already do this on delete messages btw

FasterSpeeding: The gateway behaviour is just an artifact of the fact that its done through an event dispatch system, its not the best way to handle starting + stopping callbacks and if a starting callback fails there's no guarantee the server will have setup all the resources it needs to function and the assumption should be that it hasnt

This is modelled after asgi which would call a set of startup failed callbacks after a failed startup instead of the shutdown callbacks as the shutdown callbacks would only be called if server had been well successfully started

FasterSpeeding: The approach of just throwing the error to asyncio would make it a lot harder to actually handle said errors while associating them as having killed the application

FasterSpeeding: As the example I showed before goes it is possibly to concatenate traceback but I'm not sure you'd like that either lol https://github.com/python-trio/trio/blob/7b86d20549bba2629eb929a72022a19b999c91d5/trio/_core/_multierror.py

davfsa: Bump to this :)

I would rather just logging the errors instead of actually exiting with an error for consistency with GatewayBot

davfsa (approved):

PR #1013, opened 2022-02-08
Remove redundant CommandBuilder fields

FasterSpeeding:
### Checklist


changes/1013.change.md
```diff
@@ -0,0 +1 @@
+Removed some redundant/unused fields from CommandBuilder.
```
FasterSpeeding, line 1: yeah idk what category to put this under cause its not really breaking (CommandBuilder was already broken for ppl who were directly using its not useful by itself so ppl will have moved away from using it directly as its a base class) and its not really a feature either

davfsa: Can be marked as a bugfix then

hikari/impl/special_endpoints.py
```diff
@@ -1005,20 +998,23 @@ def options(self) -> typing.Sequence[commands.CommandOption]:
...
 class ContextMenuCommandBuilder(CommandBuilder, special_endpoints.ContextMenuCommandBuilder):
     """Builder class for context menu commands."""
 
-    def build(self, entity_factory: entity_factory_.EntityFactory, /) -> data_binding.JSONObjectBuilder:
-        data = super().build(entity_factory)
-        data["type"] = self._type
-        return data
+    _type: commands.CommandType = attr.field()
+    # name is re-declared here to ensure type is before it in the initializer's args.
+    _name: str = attr.field()
```
davfsa, line 1013:
```suggestion
    _type: commands.CommandType = attr.field(kw_only=True)
    # name is re-declared here to ensure type is before it in the initializer's args.
    _name: str = attr.field(kw_only=True)
```

This is a bit from memory, since github mobile doesnt let me investigate nicely, but I believe the others were marked as kwargs only

FasterSpeeding: These weren't since they're required, this just matches the pattern found in other parts of the library for signatures

davfsa (approved):

PR #1022, opened 2022-02-17
Allow for custom $-substitutions in `hikari.internal.ux.print_banner`

### LINKED ISSUE #1006: Allow custom args in `hikari.internal.ux.print_banner`
### Summary

Allow any `args` dict to be used in `hikari.internal.ux.print_banner`

### Why is this needed?

It is difficult to add extra information that changes to the banner. For example, a version number. Allowing custom args to be passed in would make this extremely easy,

### Ideal implementation

I think the a kwarg should be added.
```python
async def print_banner(package, allow_color, force_color, args=None):
    if args is None:
       # Set args to the current version
```
Updating args may also work
```python
async def print_banner(package, allow_color, force_color, custom_args=None):
    args = {...}    
    if custom_args:
        args.update(custom_args)
```

### Checklist

Lunarmagpie:
### Summary

closes #1006 
Allows custom $-substitutions to be used when printing a banner.

### Checklist


hikari/internal/ux.py
```diff
@@ -198,6 +205,9 @@ def print_banner(package: typing.Optional[str], allow_color: bool, force_color:
         "system_description": " ".join(filtered_system_bits),
     }
 
+    if extra_args:
+        args.update(extra_args)
```
davfsa, line 209: Do we really want to allow then to update the values of the existing args?

Lunarmagpie: I could switch it out for something like this.
```python
for key in extra_args:
    if key in args:
        raise ValueError("Default $-substitutions cannot be modified")
args.update(extra_args)
```
imo using update is just much cleaner code and fields like `system_description` should be modifiable anyway.

davfsa: I was thinking more on using `.setdefault`, but that approach looks good too.

I am not a fan of allowing people to modify it easily as it could lead to conflics as well as over writing values that third party libs should not be modifying

Lunarmagpie: Thats a good point, and also people could accidentally mess something up. I made it so arguments can't be overwritten.

davfsa:
https://github.com/hikari-py/hikari/blob/master/hikari/impl/bot.py#L652
https://github.com/hikari-py/hikari/blob/master/hikari/impl/rest_bot.py#L319

Signature needs to be updated here too. But rest looks ok 👍

Lunarmagpie:
i've updated the signatures

hikari/internal/ux.py
```diff
@@ -198,6 +206,12 @@ def print_banner(package: typing.Optional[str], allow_color: bool, force_color:
         "system_description": " ".join(filtered_system_bits),
     }
 
+    if extra_args:
+        for key in extra_args:
+            if key in args:
+                raise ValueError(f"Cannot overwrite $-substitution `{key}`. Please use a different key.")
```
davfsa, line 212: One last thing I just realized I forgot to comment on, we should also document that it raises a value error:

```
Raises
------
builtins.ValueError
    <Hi>
```

davfsa (commented):
LGTM, just the one little documentation thing and then can be merged :)

hikari/internal/ux.py
Lunarmagpie: Should this also be added to `RestBot.print_banner` and `GatewayBot.print_banner`?

davfsa: Yep :)

Lunarmagpie: added :)

hikari/internal/ux.py
```diff
@@ -95,6 +95,11 @@ def init_logging(
         If `builtins.True`, return `builtins.True` always, otherwise only
         return `builtins.True` if the device supports colour output and the
         `allow_color` flag is not `builtins.False`.
+
+    Raises
+    ------
+    builtins.ValueError
+        Default $-substitution was overridden.
```
davfsa, line 102: Think this is meant to go bellow instead of here :)

hikari/impl/bot.py
```diff
@@ -674,8 +679,16 @@ def print_banner(banner: typing.Optional[str], allow_color: bool, force_color: b
...
         !!! note
             `force_color` will always take precedence over `allow_color`.
+        extra_args : typing.Optional[typing.Dict[builtins.str, builtins.str]]
+            If provided, extra $-substitutions to use when printing the banner.
+            Default substitutions can not be overwritten.
+
+        Raises
+        ------
+        builtins.ValueError
+            Default $-substitution was overridden.
```
davfsa, line 689:
```suggestion
            If `extra_args` contains a default $-substitution.
```

Lunarmagpie: done

hikari/internal/ux.py
Lunarmagpie: Can you clarify what I should do?

Lunarmagpie: oh...

Lunarmagpie: it has been fixed :pensive:

davfsa (approved):

FasterSpeeding (approved):

PR #1023, opened 2022-02-17
Properly close when not fully started

davfsa:
### Summary
Properly close when not started correctly.

Cleans up errors with "Unclosed client session" and anything else the event loop might want to throw
image

### Checklist

### DIFF
--- a/hikari/impl/bot.py
+++ b/hikari/impl/bot.py
@@ -439,7 +439,6 @@ class GatewayBot(traits.GatewayBotAware):
         return self._cache.get_me()
 
     async def close(self) -> None:
-        """Kill the application by shutting all components down."""
         self._check_if_alive()
         await self._close()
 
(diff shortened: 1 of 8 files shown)


hikari/impl/bot.py
```diff
@@ -968,19 +971,20 @@ async def start(
 
         _validate_activity(activity)
 
-        # Dispatch the update checker, the sharding requirements checker, and dispatch
-        # the starting event together to save a little time on startup.
         start_time = time.monotonic()
+        self._rest.start()
+        self._voice.start()
```
FasterSpeeding, line 976: What're the implications of starting up voice (a competent which should be relying on gateway) before the actual shards are started?

davfsa: This is already done currently. https://github.com/hikari-py/hikari/blob/master/hikari/impl/bot.py#L1017-L1018

I am unsure why it is like this and it has a comment to specify that it must, as there is no code that leads me to believe it should.

Instead of this fix, I can also have checks to only call close on the components that are actually open. That or we could allow closing even if its not open. Having that call just return normally

PR #1024, opened 2022-02-17
Add a slash command example

thesadru:
### Summary
Added a slash command example since Hikari was lacking one.

### Checklist

### DIFF
--- a/changes/1024.doc.md
+++ b/changes/1024.doc.md
@@ -1,1 +1,0 @@
-Add a slash command example.
--- a/changes/1024.doc.md
+++ b/changes/1024.doc.md
@@ -1,0 +1,1 @@
+Add a slash command example.
--- a/examples/slash.py
+++ b/examples/slash.py
@@ -17,8 +17,8 @@ bot = hikari.GatewayBot(token=os.environ["BOT_TOKEN"])
 
 
 @bot.listen()
-async def on_interaction(event: hikari.InteractionCreateEvent) -> None:
-    """Listen for messages being created."""
+async def handle_interactions(event: hikari.InteractionCreateEvent) -> None:
+    """Listen for slash commands being executed."""
     if not isinstance(event.interaction, hikari.CommandInteraction):
         # only listen to command interactions, no others!
         return
@@ -34,6 +34,28 @@ async def on_interaction(event: hikari.InteractionCreateEvent) -> None:
             hikari.ResponseType.MESSAGE_CREATE,
             "Hello, this is an example bot written in hikari!",
         )
+
+
+@bot.listen()
+async def register_commands(event: hikari.StartedEvent) -> None:
+    """Register ping and info commands."""
+    # replace with the guild id you want the commands to be in
+    # by default commands will be global, but may take up to an hour to register
+    guild_id = hikari.UNDEFINED
+
+    me = bot.get_me()
+    assert me is not None, "Bot hasn't been started yet"
+
+    commands = [
+        bot.rest.slash_command_builder("ping", "Get the bot's latency."),
+        bot.rest.slash_command_builder("info", "Learn something about the bot."),
+    ]
+
+    await bot.rest.set_application_commands(
+        application=me.id,
+        commands=commands,
+        guild=guild_id,
+    )
 
 
 bot.run()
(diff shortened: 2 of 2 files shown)


changes/1024.doc.md
```diff
@@ -0,0 +1 @@
+Add a slash command example.
```
davfsa, line 1: No need for a changelog file :)

examples/slash.py
```diff
@@ -0,0 +1,39 @@
...
+# You should have received a copy of the CC0 Public Domain Dedication along with this software.
+# If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
+"""A simple bot with some simple commands."""
+
+import os
+
+import hikari
+
+bot = hikari.GatewayBot(token=os.environ["BOT_TOKEN"])
+
```
davfsa, line 17:
```suggestion
@bot.listen()
async def starting_event(_: hikari.StartingEvent) -> None:
    # Create commands
```

davfsa (commented):
You sort of snatched this one from under my belt, had it down on my TODO list as something I need to add. But thank you regardless :)

examples/slash.py
```diff
@@ -0,0 +1,39 @@
...
+
+@bot.listen()
+async def on_interaction(event: hikari.InteractionCreateEvent) -> None:
+    """Listen for messages being created."""
```
davfsa, line 21:
```suggestion
async def handle_interactions(event: hikari.InteractionCreateEvent) -> None:
    """Listen for slash commands being executed."""
```

examples/slash.py
```diff
@@ -0,0 +1,61 @@
...
+#
+# You should have received a copy of the CC0 Public Domain Dedication along with this software.
+# If not, see <https://creativecommons.org/publicdomain/zero/1.0/>.
+"""A simple bot with some simple commands."""
+
+import os
+
+import hikari
+
+bot = hikari.GatewayBot(token=os.environ["BOT_TOKEN"])
```
davfsa, line 16:
```suggestion
# Set this to the guild to create the slash commands in.
# To create global commands, set to `hikari.UNDEFINED`.
#
# It is important to note that global commands can take up to an hour to show in the client
# and should only be used when "publishing" commands after they are built. For testing
# use guild based commands as they are instant. 
COMMAND_GUILD_ID = 0

bot = hikari.GatewayBot(token=os.environ["BOT_TOKEN"])
```

examples/slash.py
```diff
@@ -0,0 +1,61 @@
...
+    me = bot.get_me()
+    assert me is not None, "Bot hasn't been started yet"
+
+    commands = [
+        bot.rest.slash_command_builder("ping", "Get the bot's latency."),
+        bot.rest.slash_command_builder("info", "Learn something about the bot."),
+    ]
+
+    await bot.rest.set_application_commands(
+        application=me.id,
```
davfsa, line 55:
```suggestion
    application = await bot.rest.fetch_application()

    commands = [
        bot.rest.slash_command_builder("ping", "Get the bot's latency."),
        bot.rest.slash_command_builder("info", "Learn something about the bot."),
    ]

    await bot.rest.set_application_commands(
        application=application.id,
```

examples/slash.py
```diff
@@ -0,0 +1,61 @@
...
+            hikari.ResponseType.MESSAGE_CREATE,
+            f"Pong! {bot.heartbeat_latency * 1_000:.0f}ms",
+        )
+
+    elif event.interaction.command_name == "info":
+        await event.interaction.create_initial_response(
+            hikari.ResponseType.MESSAGE_CREATE,
+            "Hello, this is an example bot written in hikari!",
+        )
+
```
davfsa, line 37:
```suggestion
    elif event.interaction.command_name == "ephemeral":
        await event.interaction.create_initial_response(
            hikari.ResponseType.MESSAGE_CREATE,
            "Only you can see this, keep it a secret :)",
            flags=hikari.MessageFlag.EPHEMERAL,
        )
```

examples/slash.py
thesadru: Most of the time it just works to keep it as undefined. Idk if we should completely force users to make them guild commands.

davfsa: I have lately seen a lot of people wondering why their commands dont show up as soon as they run. At least like this they will get a nice error that they will have to fix by deciding to either place an ID or set it to `hikari.UNDEFINED`

thesadru: Fair enough

FasterSpeeding (approved):

PR #1034, opened 2022-02-26
Add slots check to the CI

FasterSpeeding:
### Summary
This verifies that all classes which can have slots do have slots, makes sure there aren't repeated slots and also makes sure that slots aren't redundantly declared on classes which can't have slots due to their parent class not having slots (e.g. Exceptions).

For reference this is used by aiohttp as a part of their CI.

### Checklist


pipelines/slotscheck.nox.py
```diff
@@ -0,0 +1,32 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Check for common slotting mistakes."""
+
+from pipelines import config
+from pipelines import nox
+
+
+@nox.session()
+def slotscheck(session: nox.Session) -> None:
```
davfsa, line 29: Might want to add it to the default sessions too

davfsa (approved):

PR #1058, opened 2022-03-06
Optimize 'receive_json' for the happy path

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -177,53 +177,60 @@ class _GatewayTransport(aiohttp.ClientWebSocketResponse):
         await self.send_str(pl, compress)
 
     async def _receive_and_check(self, timeout: typing.Optional[float], /) -> str:
-        buff = bytearray()
+        message = await self.receive(timeout)
+
+        if message.type == aiohttp.WSMsgType.TEXT:
+            assert isinstance(message.data, str)
+            return message.data
+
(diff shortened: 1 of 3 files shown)


hikari/impl/shard.py
```diff
@@ -177,53 +177,60 @@ async def send_json(
         await self.send_str(pl, compress)
 
     async def _receive_and_check(self, timeout: typing.Optional[float], /) -> str:
-        buff = bytearray()
+        message = await self.receive(timeout)
 
-        while True:
-            message = await self.receive(timeout)
+        if message.type == aiohttp.WSMsgType.TEXT:
+            return message.data  # type: ignore
```
FasterSpeeding, line 183: why not assert the type?

FasterSpeeding:
LGMT though

hikari/impl/shard.py
davfsa: This was already like this, I only moved code around. Can be changed tho

FasterSpeeding: Yeah I'd appreciate changing it even if its not really the purpose of this pr 👍

davfsa: Added

FasterSpeeding (dismissed):

FasterSpeeding (approved):

PR #1060, opened 2022-03-06
Added timeout methods to `hikari.guilds.Member`

Shinobou:
Added `timeout` and `untimeout` methods to `hikari.guilds.Member`

### Summary

Since ban and kick have their own methods, I thought adding timeout methods to the Member class would be useful. 

### Checklist

### DIFF
--- a/hikari/guilds.py
+++ b/hikari/guilds.py
@@ -1014,6 +1014,109 @@ class Member(users.User):
             reason=reason,
         )
 
+    async def timeout(
+        self,
+        communication_disabled_until: datetime.datetime,
+        *,
+        reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
+    ) -> Member:
+        """Puts the Member in timeout.
+
+        Parameters
+        ----------
+        communication_disabled_until : datetime.datetime
+            the datetime when the timeout (disable communication)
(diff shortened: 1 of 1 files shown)


davfsa:
This is just an alias of `Member.edit`, which I see as unnecessary and just obfuscates things.

I -1 this

FasterSpeeding:
I can see how this'd make sense from a UX perspective, my assumption is that most of the time you're not going to be editing 
any of the other fields on a member when you're timing them out and having a named moderation action method for it should make it generally easier to find how to time someone out from a doc search or from glossing over code.

But i'm honestly mostly indifferent to this idea tbh ~~also "untimeout" just sounds weird~~

HyperGH:
I think, if this were to be implemented at all, it is sufficient to pass `None` to `communication_disabled_until` in `timeout`. There is no need for `untimeout`. With that said I think `Member.edit()` is already more than enough for this.

davfsa:
I really appreciate the PR, but I dont think it is needed.

I agree it does make sense in a UX perspective, which is the idea behind this PR. But such a small alias of something that already exists in this object I don't see as necessarily helpful and just adds more code to maintain.

Again, thanks for the PR, but will be closing it

PR #1061, opened 2022-03-07
Fix handling of when the first zlib message is complete

FasterSpeeding:
### Checklist

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -112,6 +112,17 @@ if typing.TYPE_CHECKING:
     _ZlibDecompressor = zlib._Decompress
 
 
+_RECONNECTABLE_CLOSE_CODES = frozenset(
+    (
+        errors.ShardCloseCode.UNKNOWN_ERROR,
+        errors.ShardCloseCode.DECODE_ERROR,
+        errors.ShardCloseCode.INVALID_SEQ,
+        errors.ShardCloseCode.SESSION_TIMEOUT,
+        errors.ShardCloseCode.RATE_LIMITED,
+    )
+)
+
+
 # aiohttp.ClientWebSocketResponse isn't slotted
 @typing.final
 class _GatewayTransport(aiohttp.ClientWebSocketResponse):
@@ -184,21 +195,16 @@ class _GatewayTransport(aiohttp.ClientWebSocketResponse):
             return message.data
 
         if message.type == aiohttp.WSMsgType.BINARY:
+            if message.data.endswith(b"\x00\x00\xff\xff"):
+                return self.zlib.decompress(message.data).decode("utf-8")
+
             return await self._receive_and_check_complete_zlib_package(message.data, timeout)
 
         if message.type == aiohttp.WSMsgType.CLOSE:
             close_code = int(message.data)
             reason = message.extra
             self.logger.error("connection closed with code %s (%s)", close_code, reason)
-
-            can_reconnect = close_code < 4000 or close_code in (
-                errors.ShardCloseCode.UNKNOWN_ERROR,
-                errors.ShardCloseCode.DECODE_ERROR,
-                errors.ShardCloseCode.INVALID_SEQ,
-                errors.ShardCloseCode.SESSION_TIMEOUT,
-                errors.ShardCloseCode.RATE_LIMITED,
-            )
-
+            can_reconnect = close_code < 4000 or close_code in _RECONNECTABLE_CLOSE_CODES
             raise errors.GatewayServerClosedConnectionError(reason, close_code, can_reconnect)
 
         if message.type == aiohttp.WSMsgType.CLOSING or message.type == aiohttp.WSMsgType.CLOSED:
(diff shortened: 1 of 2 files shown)


hikari/impl/shard.py
```diff
@@ -221,16 +227,18 @@ async def _receive_and_check_complete_zlib_package(
     ) -> str:
         buff = bytearray(initial_data)
 
-        while True:
+        while not buff.endswith(b"\x00\x00\xff\xff"):
             message = await self.receive(timeout)
```
FasterSpeeding, line 232: just a note, the old logic would've handled the CLOSE, CLOSED and CLOSING message types being received between zlib payloads but this current code treats that as a general error

davfsa: That is true, didn't even think of that, will rethink the logic.

Thanks for pointing it out

FasterSpeeding: If you want to handle those cases (that i'm not even sure could happen tbh, just wanted to point out theoretical changes in behaviour) I can just move the error code handling logic over to a function and call it in both places on error but from some cursory testing seems like payloads are rarely split between messages when using zlib transport compression

davfsa: Tbh, uppon more thinking on this, we can treat it as an error. The connection was closed but we expected a BINARY package, so error sounds good enough for me. Will test a bit and see what would actually happen if we close mid buffer.

hikari/impl/shard.py
```diff
@@ -221,16 +228,18 @@ async def _receive_and_check_complete_zlib_package(
...
 
-        while True:
+        while not buff.endswith(_ZLIB_SUFFIX):
             message = await self.receive(timeout)
 
-            if message.type != aiohttp.WSMsgType.BINARY:
-                raise errors.GatewayError(f"Unexpected message type received {message.type.name}, expected BINARY")
+            if message.type == aiohttp.WSMsgType.BINARY:
+                buff.extend(message.data)
+                continue
```
davfsa, line 236: Having it in this order seems to be more innefficient, this way is prefered

```py
            if message.type != aiohttp.WSMsgType.BINARY:
                # Raise error

            buff.extend(message.data)
```

My (not researched, just tested) guess is that the continue adds one more instruction that affects how fast this runs.

*Not a strict suggestion per say, just a comment :)*

hikari/impl/shard.py
```diff
@@ -184,21 +196,16 @@ async def _receive_and_check(self, timeout: typing.Optional[float], /) -> str:
             return message.data
 
         if message.type == aiohttp.WSMsgType.BINARY:
+            if message.data.endswith(_ZLIB_SUFFIX):
+                return self.zlib.decompress(message.data).decode("utf-8")
```
davfsa, line 200: Since we already have the while loop in the `_receive_and_check_complete_zlib_package`, wouldnt this just be duplicated code?

*If anything, `_receive_and_check_complete_zlib_package` could just be merged into here*

hikari/impl/shard.py
FasterSpeeding: > that i'm not even sure could happen tbh

Cause discord, as of present, basically just doesn't send split zlib payloads so the chances of getting a close in a case that doesn't happen are like next to impossible under current behaviour but that doesn't matter since this is handling a theoretical case

hikari/impl/shard.py
davfsa: I was just pointing out that, for some reason, in cpython 3.8 its faster the other way around. Just a little comment I wanted to leave. Idm :)

hikari/impl/shard.py
davfsa: > From testing i couldn't find a case where discord splits zlib payload between payloads and have to assume its either a rare case or just doesn't ever happen. This heavily hampers the chances they'd ever be splitup by a close.

From the example in the docs, it looks like it is possible. https://github.com/discord/discord-api-docs/blob/master/docs/topics/Gateway.md#transport-compression.

> The current code would just kill the shard since GatewayError bubbles up from the shard run logic to whatever ran it

Yep, thanks for pointing it out, I didnt notice when I did it. But idk if we should change it to actually handle the close message types.

hikari/impl/shard.py
```diff
@@ -221,16 +228,18 @@ async def _receive_and_check_complete_zlib_package(
     ) -> str:
         buff = bytearray(initial_data)
 
-        while True:
+        while not buff.endswith(_ZLIB_SUFFIX):
```
davfsa, line 231: Since we have the fast path above, is this really needed anymore?

hikari/impl/shard.py
davfsa: > If it isn't reconnects wouldn't work under these cases

Ok. Will try and have a look because I have a slight hunch (random idea, probably not correct) on why https://github.com/hikari-py/hikari/blob/19a90ea879c4dcb0e61a71d7070d9f2c83df39fc/hikari/impl/shard.py#L215 is a thing. Will make sure tho, and maybe we can remove that line all together.

For now this looks good. Thanks a lot for catching it!

hikari/impl/shard.py
FasterSpeeding: > since _receive_and_check_complete_zlib_package is a rare case it was left with the change of using while not like that to make it just more resilient and less likely to be quietly broken by changes to the code calling it (it also just looks nicer that way)

(looks nicer meaning it just is nicer code)
(and changes to the code calling it includes the removal of a fast path)
(also this code is rarely called and even under cases where discord split up large payloads it'd still be rarely called so its shouldn't really a performance concern)

hikari/impl/shard.py
FasterSpeeding: I've changed the code a bit to handle closes and what not between zlib portions btw

davfsa (approved):

PR #1066, opened 2022-03-07
Fix incorrect type for ATTACHMENT option values

tandemdude:
- Fixes the value for options with ATTACHMENT type being of `str` instead of `Snowflake`

### Checklist


thesadru (approved):
mb 😅

davfsa (commented):
Missing changelog

davfsa (approved):

Issue #748, opened 2021-08-28
Better distinguish impl detail in configuration classes 


FasterSpeeding:
### Summary
The configuration classes should better distinguish between standard implementation detail and configuration which should apply to all implementations.

### Problem
Right now implementation detail and fields which should be enforced for all implementations are both present on the config classes with no distinction, this can be an issue when clients are expected to expose these configuration classes through properties are some of the fields may not even apply to them and this would break introspection logic.

### Ideal implementation
~~Where applicable configuration classes which mix implementation detail and standard information should be split into bases classes which just have detail which is always present (and are exposed by the abstract clients through properties) and extended by another class to add fields which are specific to the standard implementations; with the standard implementations documenting their `__init__`s as taking the extended impl specific config classes).~~

Alternatively the config module could be split between `hikari.api.config` and `hikari.impl.config` to better follow our impl vs api strategy which now that I think about it is probably a better idea.

### Checklist

- [x] I have searched the issue tracker and have made sure it's not a duplicate.
      If it is a follow up of another issue, I have specified it.

PR #1067, opened 2022-03-09
Split config module between api.config and impl.config

FasterSpeeding:
### Summary
This is just an attempt at #748

Most of the http and proxy settings were not included in the abstract interface for this attempt as these mostly rely on impl detail around aiohttp and its configuration and aren't overly helpful for introspection.

### Checklist

### Related issues
Close: #748


pessiuff (dismissed):

hikari/impl/bot.py
```diff
@@ -195,9 +195,9 @@ class GatewayBot(traits.GatewayBotAware):
         will __force__ colour to be used in console-based output. Specifying a
         `"CLICOLOR_FORCE"` environment variable with a non-`"0"` value will
         override this setting.
-    cache_settings : typing.Optional[hikari.config.CacheSettings]
+    cache_settings : typing.Optional[hikari.impl.config.CacheSettings]
```
davfsa, line 198: Shouldnt this be `api`?

davfsa (approved):

hikari/impl/bot.py
FasterSpeeding: Nah, the idea here is that `hikari.impl.config` is the configuration objects the standard implementations take in and `hikari.api.config` are the fields that are guaranteed to be exposed on config properties between impls.

For this feature I took the liberty of breaking the normal standard of only ever exposing the api classes on return values as it is helpful for fields like `hikari.impl.GatewayBot.http_settings` to expose the config impl detail since this impl detail has a lot 
 more info and is guaranteed to stay the same for that class until a breaking change is made like switching to a different rest client.

FasterSpeeding: Although i do need to change `from hikari.impl import config` to `from hikari.impl import config as config_impl` in this module to make it more clear lol

FasterSpeeding: Fixed those inconsistent imports btw

davfsa (approved):

PR #1069, opened 2022-03-09
Disable ws autoclose

davfsa:
### Summary
From testing, this doesnt seem to affect anything at all and just sends payloads that Discord will ignore as well as close unexpectedly on us. Further testing (running constantly for at least 48h) is needed, but I dont see why we should have these enabled.

Would greatly appreciate if someone could try out this PR in their bot and lmk how it goes

### Checklist

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -279,6 +279,8 @@ class _GatewayTransport(aiohttp.ClientWebSocketResponse):
                     proxy=proxy_settings.url,
                     proxy_headers=proxy_settings.headers,
                     url=url,
+                    # We manage these ourselves
+                    autoclose=False,
                 )
             )
             assert isinstance(web_socket, cls)
--- a/tests/hikari/impl/test_shard.py
+++ b/tests/hikari/impl/test_shard.py
@@ -354,6 +354,7 @@ class TestGatewayTransport:
             proxy=proxy_settings.url,
             proxy_headers=proxy_settings.headers,
             url="https://some.url",
+            autoclose=False,
         )
         mock_client_session.assert_used_once()
         mock_websocket.assert_used_once()


CircuitSacul:
I tested this by installing directly from your branch. I didn't notice anything different when running the bot (I also double checked the version, which was dev108). I ran it for about 10 minutes, occasionally using commands.

davfsa:
> I tested this by installing directly from your branch. I didn't notice anything different when running the bot (I also double checked the version, which was dev108). I ran it for about 10 minutes, occasionally using commands.

Glad to hear, but a more real-word test would be having the bot run for 48h, to make sure it handles all cases correctly (reconnects, unexpected closes, etc).

Appreciate the testing tho!

Jonxslays:
Starr is now running on this PR. I'll get back to you.

davfsa:
Ran by @Jonxslays for a while now and there are no changes in reconnect/random disconnects behavior (unfortunately the annoying bug I wanted to fix is still there, but no can do ig)

Marking as ready for review and leaving it up to @FasterSpeeding chose if useful or not

hikari/impl/shard.py
```diff
@@ -279,6 +279,9 @@ async def connect(
                     proxy=proxy_settings.url,
                     proxy_headers=proxy_settings.headers,
                     url=url,
+                    # We manage these ourselves
+                    autoping=False,
```
FasterSpeeding, line 283: This is wrong, none of this logic handles PING or PONG events and without autoping this logic could break if discord started sending PINGs

davfsa: Discord's way of ping poing is by using heartbeats, but I do see and understand the concern in the case where they start sending ping pongs. Will close this then

FasterSpeeding: You can still disable autoclose though, that will improve semantics since the current logic does handle close codes.

Also ping/pong is a part of the spec and should be handled regardless of discord's current behaviour

FasterSpeeding (approved):

PR #1092, opened 2022-03-25
Remove store sku related application fields and store channels

FasterSpeeding:
### Checklist


davfsa:
Are these removed completely from the API? As far as I am aware there are still some lindering around (cant check atm, but check one of a store channel I have access to later today)

davfsa (dismissed):

davfsa (approved):

FasterSpeeding:
> Are these removed completely from the API? As far as I am aware there are still some lindering around (cant check atm, but check one of a store channel I have access to later today)

The assumption here is that all of the store (channel) specific application fields should be treated as likely to be removed in the future since they serve no purpose without the store and primary_sku_id + slug were never useful to bots as far as the documented API goes to begin with anyways.

PR #1094, opened 2022-03-25
Filter event dispatch to only deserialize events being listened to

FasterSpeeding:
### Checklist


hikari/impl/entity_factory.py
```diff
@@ -1357,7 +1514,7 @@ def deserialize_guild_preview(self, payload: data_binding.JSONObject) -> guild_m
             description=payload["description"],
         )
 
-    def _set_guild_attributes(self, payload: data_binding.JSONObject) -> _GuildFields:
+    def set_guild_attributes(self, payload: data_binding.JSONObject) -> _GuildFields:
```
FasterSpeeding, line 1517: Could this just pass the private method to the gold builder instead of making it public?

davfsa: There really isnt a need for this to be public, so yes, it can be made private again

davfsa (approved):

davfsa (approved):

PR #915, opened 2021-11-23
Add hikari-dev package to install development dependencies

davfsa:
### Checklist

- [x] I have run `nox` and all the pipelines have passed.
- [x] I have made unittests according to the code I have added/modified/deleted.

### Related issues
Closes #883
Closes #886

PR #1098, opened 2022-03-25
Temporarily freeze jinja2 to 3.0.3

davfsa:
### Summary
Flake8-html already has a fix for this on master, so we just need to wait for a release (https://github.com/lordmauve/flake8-html/commit/edef580c6d5e0ccb3ab54aa31b42b4359f7b9235)

### Checklist


FasterSpeeding:
Using pip-compile in a similar style to how some of my other projects use it for dev deps would avoid this btw

FasterSpeeding (approved):

davfsa:
> Using pip-compile for dev deps in a similar fashion to how some of my other projects use it would avoid this for the most part btw

Was thinking about changing #915 to use that format, but havent gotten around to do it yet

PR #1090, opened 2022-03-23
add hikari.locales.Locale

Le0Developer:
### Summary

Add `hikari.locales.Locale` enum. (possibly breaking change?)

Not sure where else to put it, tell me if there's a better place for it. Open for feedback :)

### Checklist

- [x] I have run `nox` and all the pipelines have passed.
- [x] I have made unittests according to the code I have added/modified/deleted.

PR #1100, opened 2022-03-25
fix: locale is optional in deserialize_my_user

Le0Developer:
### Summary

Fixes an error introduced in #1090.

### Checklist


hikari/impl/entity_factory.py
```diff
@@ -2921,7 +2921,7 @@ def deserialize_my_user(self, payload: data_binding.JSONObject) -> user_models.O
             is_bot=user_fields.is_bot,
             is_system=user_fields.is_system,
             is_mfa_enabled=payload["mfa_enabled"],
-            locale=locales.Locale(payload["locale"]),
+            locale=locales.Locale(payload.get("locale")),
```
davfsa, line 2924: There is actually more regressions than this one. Additionally, this wont really fix the error as you will just be casting None to the enum. Will open a PR fixing all of them :)

PR #1103, opened 2022-03-25
Allow passing multiple events to the @listen decorator

Jonxslays:
### Summary

This pull requests allows passing multiple events to the listen decorator.
Alternatively, a Union can be used in the type hint to achieve the same thing.

```py
@bot.listen(hikari.GuildAvailableEvent, hikari.GuildJoinEvent)
async def on_guild_available(event: ...) -> None:
    ...

@bot.listen()
async def on_start(event: hikari.StartingEvent | hikari.StartedEvent) -> None:
    ...
```

### Checklist

Flake8 session is failing - but it was on master before I touched anything.

### DIFF
--- a/hikari/api/event_manager.py
+++ b/hikari/api/event_manager.py
@@ -408,7 +408,6 @@ class EventManager(abc.ABC):
             instead from the type hints on the function signature.
 
             `T` must be a subclass of `hikari.events.base_events.Event`.
-
         *event_types : typing.Type[T]
             The additional event types to subscribe to.
 
--- a/hikari/api/event_manager.py
+++ b/hikari/api/event_manager.py
@@ -393,7 +393,6 @@ class EventManager(abc.ABC):
     @abc.abstractmethod
     def listen(
         self,
-        event_type: typing.Optional[typing.Type[EventT_co]] = None,
         *event_types: typing.Type[EventT_co],
     ) -> typing.Callable[[CallbackT[EventT_co]], CallbackT[EventT_co]]:
         """Generate a decorator to subscribe a callback to an event type.
@@ -402,14 +401,12 @@ class EventManager(abc.ABC):
 
         Parameters
         ----------
-        event_type : typing.Optional[typing.Type[T]]
-            The event type to subscribe to. The implementation may allow this
+        *event_types : typing.Optional[typing.Type[T]]
+            The event types to subscribe to. The implementation may allow this
             to be undefined. If this is the case, the event type will be inferred
             instead from the type hints on the function signature.
 
             `T` must be a subclass of `hikari.events.base_events.Event`.
-        *event_types : typing.Type[T]
-            The additional event types to subscribe to.
 
         Returns
         -------
--- a/hikari/api/event_manager.py
+++ b/hikari/api/event_manager.py
@@ -394,6 +394,7 @@ class EventManager(abc.ABC):
     def listen(
         self,
         event_type: typing.Optional[typing.Type[EventT_co]] = None,
+        *event_types: typing.Type[EventT_co],
     ) -> typing.Callable[[CallbackT[EventT_co]], CallbackT[EventT_co]]:
         """Generate a decorator to subscribe a callback to an event type.
 
@@ -407,6 +408,9 @@ class EventManager(abc.ABC):
             instead from the type hints on the function signature.
 
             `T` must be a subclass of `hikari.events.base_events.Event`.
+
+        *event_types : typing.Type[T]
+            The additional event types to subscribe to.
 
         Returns
         -------
--- a/hikari/impl/event_manager_base.py
+++ b/hikari/impl/event_manager_base.py
@@ -50,8 +50,6 @@ from hikari.internal import reflect
 from hikari.internal import ux
 
 if typing.TYPE_CHECKING:
-    import types
-
     from hikari import intents as intents_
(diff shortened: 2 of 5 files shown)


hikari/api/event_manager.py
```diff
@@ -408,6 +409,9 @@ def listen(
 
             `T` must be a subclass of `hikari.events.base_events.Event`.
 
+        *event_types : typing.Type[T]
```
davfsa, line 412: Extra space here

tests/hikari/impl/test_event_manager_base.py
```diff
@@ -699,6 +700,25 @@ async def test(event):
...
+
+            @event_manager.listen(member_events.MemberCreateEvent, member_events.MemberDeleteEvent)
+            async def test(event):
+                ...
+
+        assert subscribe.call_count == 2
+        resolve_signature.assert_not_called()
+        subscribe.assert_has_calls(
+            subscribe(member_events.MemberCreateEvent, test, _nested=1),
+            subscribe(member_events.MemberDeleteEvent, test, _nested=1),
```
davfsa, line 719:
```suggestion
            mock.call(member_events.MemberCreateEvent, test, _nested=1),
            mock.call(member_events.MemberDeleteEvent, test, _nested=1),
```

Although if this test passes, that is a new thing i learned about mock today. But would rather keep it consistent :)

Jonxslays: that was just me being dumb and forgetting to change it.

thesadru:
Ideally, the origin of the annotation should be checked too in case a type with `__args__` is passed in. 
Note that in 3.9 there's only `typing.Union` but in 3.10 you must also check for `types.UnionType`

davfsa:
> Ideally, the origin of the annotation should be checked too in case a type with `__args__` is passed in.
> Note that in 3.9 there's only `typing.Union` but in 3.10 you must also check for `types.UnionType`

This works with all python versions we support, as it never actually checks for the union. It just expands the arguments of the annotation using a typing function, which will support each python version

thesadru:
Oops, I wanted to suggest also checking the origin for the edge cases when the passed-in type is not a union. Somehow managed to delete that part 😅

davfsa:
> Oops, I wanted to suggest also checking the origin for the edge cases when the passed-in type is not a union. sweat_smile

In the case it is not a union, it will get expanded as far as it can go and then passed onto subscribe. Subscribe will then make the necessary checks and error accordingly

hikari/impl/event_manager_base.py
```diff
@@ -371,31 +371,36 @@ def unsubscribe(
...
+                annotation = event_param.annotation
 
-                if event_param.annotation is event_param.empty:
+                if annotation is event_param.empty:
                     raise TypeError("Must provide the event type in the @listen decorator or as a type hint!")
 
-                event_type = event_param.annotation
+                # If annotation is not a Union, get_args will return an empty tuple
+                # so just return a tuple with the annotation
+                resolved_types = typing.get_args(annotation) or (annotation,)
```
FasterSpeeding, line 399: you should check that typing.get_origin(annotation) is either typing.Union or types.Union since typing.get_args works for other generic types which presumably is not what you want here

tests/hikari/impl/test_event_manager_base.py
```diff
@@ -707,3 +729,18 @@ async def test(event: member_events.MemberCreateEvent):
                 ...
 
         subscribe.assert_called_once_with(member_events.MemberCreateEvent, test, _nested=1)
+
+    def test_listen_when_multiple_params_provided_as_union_in_typehint(self, event_manager):
```
FasterSpeeding, line 733: There should be a separate test case for 3.10+ style unions as these are a different class

hikari/impl/event_manager_base.py
```diff
@@ -371,31 +371,36 @@ def unsubscribe(
     def listen(
         self,
         event_type: typing.Optional[typing.Type[event_manager_.EventT_co]] = None,
```
FasterSpeeding, line 373: you can really just do
```py
def listen(self, *event_types: typing.Type[event_manager_.EventT_co])
```

then implicitly check if the tuple is empty with an if statement since we shouldn't care about the first arg's name here

FasterSpeeding:
Also this needs a changelog entry

FasterSpeeding:
Linting is failing

hikari/impl/event_manager_base.py
```diff
@@ -499,32 +506,45 @@ def unsubscribe(
...
+                annotation = event_param.annotation
 
-                if event_param.annotation is event_param.empty:
+                if annotation is event_param.empty:
                     raise TypeError("Must provide the event type in the @listen decorator or as a type hint!")
 
-                event_type = event_param.annotation
+                if sys.version_info >= (3, 10):
+                    # We can use types.UnionType on 3.10+
+                    UNIONS = {typing.Union, types.UnionType}
```
FasterSpeeding, line 534: `UNIONS` can just be a module level constant

davfsa:
Lining failing due to https://github.com/psf/black/issues/2964

Will be fixed by updating black to the latest version, which I will open a PR for in a sec. Then this PR should be rebased

FasterSpeeding (approved):

davfsa (approved):

PR #1108, opened 2022-03-27
Improve pyright support

thesadru:
Ignored all pyright errors that go against the design patterns of hikari.

All of the ignored errors should be unfixable unless hikari underwent a large rewrite so there should be no problems ignoring them.


hikari/impl/rest.py
```diff
@@ -736,6 +736,7 @@ async def _request(
                     if not no_auth:
                         await live_attributes.still_alive().global_rate_limit.acquire()
 
+                    start = time.monotonic()
```
davfsa, line 739: This should only be registered when trace_logging is enabled, so this is incorrect

hikari/impl/entity_factory.py
```diff
@@ -1375,10 +1375,9 @@ def deserialize_member(
             time.iso8601_datetime_string_to_datetime(raw_premium_since) if raw_premium_since is not None else None
         )
 
+        communication_disabled_until: typing.Optional[datetime.datetime] = None
         if raw_communication_disabled_until := payload.get("communication_disabled_until"):
             communication_disabled_until = time.iso8601_datetime_string_to_datetime(raw_communication_disabled_until)
```
davfsa, line 1380: Double definition for no reason

pyproject.toml
```diff
@@ -59,10 +59,26 @@ exclude_lines = [
 ]
 
 [tool.pyright]
-include = ["examples", "hikari"]
```
davfsa, line 62: This removes examples, which we dont want

pyproject.toml
```diff
@@ -59,10 +59,26 @@ exclude_lines = [
...
+include = ["hikari"]
+exclude = ["**/__init__.py", "hikari/internal/enums.py", "hikari/internal/fast_protocol.py"]
```
davfsa, line 63: This excludes them from verify-types, which we dont want

pyproject.toml
```diff
@@ -59,10 +59,26 @@ exclude_lines = [
...
+reportOverlappingOverload = "none" # Type-Vars in last overloads may interfere
+reportIncompatibleVariableOverride = "none" # Cannot overwrite abstract properties using attrs
+
+# Attrs validators will always be unknown
+reportUnknownMemberType = "warning"
+reportUntypedFunctionDecorator = "warning"
+reportOptionalMemberAccess = "warning"
+
+# for mypy
+reportUnnecessaryIsInstance = "none"
```
davfsa, line 79: For all these, we might as well fix them instead of ignoring them, if not, they will just keep flying under the radar and we will have no idea of them

hikari/impl/entity_factory.py
thesadru: This is exactly how it is done in many other methods.

pyproject.toml
thesadru: The whole point of this Pull Request is to remove those that are completely unfixable. You cannot fix them unless you completely drop attrs which won't happen.

hikari/impl/rest.py
thesadru: This is for pyright support. We do much more for the sake of satisfying mypy.

pyproject.toml
thesadru: They generate stubs so it doesn't matter.

hikari/impl/rest.py
davfsa: But this has runtime impact, which we dont want

pyproject.toml
thesadru: Rillrate is uncheckable so I'll exclude only that and keep all other examples.

pyproject.toml
davfsa: There is an issue open to improve attrs support for pyright: https://github.com/python-attrs/attrs/issues/795, so this will most probably change in the future

thesadru: Great, we can re-enable them once support is rolled out.

FasterSpeeding (dismissed):

hikari/impl/rest.py
```diff
@@ -737,6 +737,7 @@ async def _request(
                         await live_attributes.still_alive().global_rate_limit.acquire()
 
                     if trace_logging_enabled:
+                        start = time.monotonic()
```
davfsa, line 740: Should probably be moved back down where it was so that it doesnt take into account the time of stringifying the http message

pyproject.toml
```diff
@@ -59,10 +59,26 @@ exclude_lines = [
...
+include = ["hikari", "examples"]
+exclude = ["examples/simple_dashboard.py", "**/__init__.py", "hikari/internal/enums.py", "hikari/internal/fast_protocol.py"]
 pythonVersion = "3.8"
 typeCheckingMode = "strict"
 
+reportMissingTypeStubs = "none"
+reportImportCycles = "none" # Doesn't account for TYPE_CHECKING
+reportIncompatibleMethodOverride = "none"  # This relies on ordering for keyword-only arguments
+reportOverlappingOverload = "none" # Type-Vars in last overloads may interfere
+reportIncompatibleVariableOverride = "none" # Cannot overwrite abstract properties using attrs
```
davfsa, line 71: Just so we can remove the ones that get fixed and not have them lingering around
`reportUnnecessaryTypeIgnoreComment = true`

pyproject.toml
```diff
@@ -59,10 +59,26 @@ exclude_lines = [
 ]
 
 [tool.pyright]
-include = ["examples", "hikari"]
+include = ["hikari", "examples"]
+exclude = ["examples/simple_dashboard.py", "**/__init__.py", "hikari/internal/enums.py", "hikari/internal/fast_protocol.py"]
```
davfsa, line 63: Why do the init's, enum and fast protocol need to be ignored?

And for simple dashboard, wouldnt it be better to just ignore rillrate there instead?

davfsa (commented):
Rebase needed as well

pyproject.toml
thesadru: `__init__.py` must be ignored because there will be some unused variables when importing directly.

`enums.py` and `fast_protocol.py` have a dynamically generated namespace. Mypy already ignores these files and uses stubs instead. 
I'd love to fix all of the errors in those files but that is kinda outside the scope of this PR and should be resolved separately.

I doubt there is any way to ignore specifically rillrate. If there is it'd be very helpful in other places too I think.

hikari/impl/event_manager_base.py
```diff
@@ -103,8 +103,8 @@ def __events_bitmask__(self) -> int:
 
 def _generate_weak_listener(
     reference: weakref.WeakMethod[typing.Any],
-) -> typing.Callable[[base_events.EventT], typing.Coroutine[typing.Any, typing.Any, None]]:
-    async def call_weak_method(event: base_events.EventT) -> None:
+) -> typing.Callable[[base_events.Event], typing.Coroutine[typing.Any, typing.Any, None]]:
+    async def call_weak_method(event: base_events.Event) -> None:
```
davfsa, line 107: Not really correct, as we want to keep the type

hikari/impl/event_manager_base.py
```diff
@@ -63,7 +63,7 @@
         typing.List[event_manager_.CallbackT[base_events.EventT]],
     ]
     _WaiterT = typing.Tuple[
-        typing.Optional[event_manager_.PredicateT[base_events.EventT]], asyncio.Future[base_events.EventT]
+        typing.Optional[event_manager_.PredicateT[base_events.EventT]], "asyncio.Future[base_events.EventT]"
```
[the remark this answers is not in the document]
thesadru: It would theoretically fail on runtime and pyright gives a warning for that.

hikari/impl/event_manager_base.py
thesadru: Pyright says that the TypeVar has no meaning in that context. Since the actual meaning for the dev is questionable and it truly has no meaning for type-checkers it should be the correct move to remove it.

hikari/impl/event_manager_base.py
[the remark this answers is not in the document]
davfsa: Just did some search on it, turns out its only generic from 3.9 onwards (https://github.com/python/typing/issues/446#issuecomment-623251451). All good then :+1:

hikari/impl/rest.py
```diff
@@ -1246,24 +1243,6 @@ def _build_message_payload(  # noqa: C901- Function too complex
...
-            raise TypeError(
-                "You passed a non-collection to 'components', but this expects a collection. Maybe you meant to "
-                "use 'component' (singular) instead?"
-            )
-
-        if embeds not in _NONE_OR_UNDEFINED and not isinstance(embeds, typing.Collection):
-            raise TypeError(
-                "You passed a non-collection to 'embeds', but this expects a collection. Maybe you meant to "
-                "use 'embed' (singular) instead?"
-            )
```
davfsa, line 1265: Forgot to do this, but it raising an error has to be undocumented in all the functions in rest that use `_build_message_payload`

hikari/internal/collections.py
```diff
@@ -366,10 +344,7 @@ def get_index_or_slice(
...
-        try:
-            return next(itertools.islice(mapping.values(), index_or_slice, None))
-        except StopIteration:
-            raise IndexError(index_or_slice) from None
-
-    raise TypeError(f"sequence indices must be integers or slices, not {type(index_or_slice).__name__}")
+    try:
+        return next(itertools.islice(mapping.values(), index_or_slice, None))
+    except StopIteration:
+        raise IndexError(index_or_slice) from None
```
davfsa, line 350: Docstring needs updating (forgot to do so)

hikari/colors.py
```diff
@@ -517,12 +517,12 @@ def of(cls, value: Colorish, /) -> Color:
             if len(value) != 3:
                 raise ValueError(f"Color must be an RGB triplet if set to a {type(value).__name__} type")
 
-            if any(isinstance(c, float) for c in value):
-                r, g, b = value
+            r, g, b = value
+
+            if isinstance(r, float) and isinstance(g, float) and isinstance(b, float):
```
davfsa, line 522: For future reference: This has changed from being `any` to being the equivalent of `all`. This was never documented to work with a mixture of floats and ints, and even the examples dont show that behaviour, so technically just impl behaviour

hikari/internal/collections.py
thesadru: To what? TypeError can still be raised can't it?

davfsa: Yes, but not by us. We only document what we explicitly raise

thesadru: sussy

davfsa (approved):

davfsa (approved):

FasterSpeeding (approved):

PR #1132, opened 2022-04-12
add retry on timeout

Le0Developer:
### Summary

Retry on `asyncio.TimeoutError` that aiohttp raises when a request times out.

Currently testing this in production, hence draft, no changelog fragment and a lack of formatting/tests.

### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -715,7 +715,7 @@ class RESTClientImpl(rest_api.RESTClient):
 
         url = compiled_route.create_url(self._rest_url)
 
-        # This is initiated the first time we hit a 5xx error to save a little memory when nothing goes wrong
+        # This is initiated the first time we time our or hit a 5xx error to save a little memory when nothing goes wrong
         backoff: typing.Optional[rate_limits.ExponentialBackOff] = None
         retry_count = 0
 
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -680,7 +680,7 @@ class RESTClientImpl(rest_api.RESTClient):
             return None
 
     @typing.final
-    async def _request(
+    async def _request(  # noqa: C901 CFQ001
         self,
         compiled_route: routes.CompiledRoute,
         *,
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -758,7 +758,10 @@ class RESTClientImpl(rest_api.RESTClient):
                             proxy=self._proxy_settings.url,
                             proxy_headers=self._proxy_settings.all_headers,
                         )
-                    except asyncio.TimeoutError:  # Request timed out
+                    except (
+                        asyncio.TimeoutError,
+                        aiohttp.ClientOSError,
+                    ):  # Request timed out or failed due to connection issues
                         if retry_count >= self._max_retries:
                             raise
 
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -715,7 +715,7 @@ class RESTClientImpl(rest_api.RESTClient):
 
         url = compiled_route.create_url(self._rest_url)
 
-        # This is initiated the first time we time our or hit a 5xx error to save a little memory when nothing goes wrong
+        # This is initiated the first time we time out or hit a 5xx error to save a little memory when nothing goes wrong
         backoff: typing.Optional[rate_limits.ExponentialBackOff] = None
         retry_count = 0
 
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -745,18 +745,36 @@ class RESTClientImpl(rest_api.RESTClient):
                         start = time.monotonic()
 
                     # Make the request.
-                    response = await live_attributes.still_alive().client_session.request(
-                        compiled_route.method,
-                        url,
-                        headers=headers,
-                        params=query,
-                        json=json,
-                        data=form,
-                        allow_redirects=self._http_settings.max_redirects is not None,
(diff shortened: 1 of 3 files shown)


Le0Developer:
No idea what to do about the flake8 max complexity error 🤷

davfsa:
I am a bit torn on this. It is similar to a similar PR I did some time ago to attempt a restart on timeout errors.

Now that I think about it, it might be better to just recommend people to edit their HTTPSettings to their liking if they see this error too often.

cc @FasterSpeeding Opinions?

Le0Developer:
Added aiohttp.ClientOSError to retry too.

davfsa:
> Tbh not exactly sure what you're getting at there, wdym by restart?

Currently, if any error is thrown while doing the initial request to fetch the sharding information when the bot is being started, it just exits with an error.

I had the idea of at least retrying that request but eventually scrapped it and just decided on just recommending people to edit their HTTPSettings to have no timeout or a higher one on unstable connections

FasterSpeeding:
> I had the idea of at least retrying that request but eventually scrapped it and just decided on just recommending people to edit their HTTPSettings to have no timeout or a higher one on unstable connections

Is this related to that though? Doesn't `asyncio.TImeoutError` and `aiohttp.ClientOSError` being raised effect all requests, not just startup and since they're not being caught currently unless i'm missing something changing HTTPSettings wouldn't help with that?

davfsa:
> Is this related to that though? Doesn't `asyncio.TImeoutError` and `aiohttp.ClientOSError` being raised effect all requests, not just startup and since they're not being caught currently unless i'm missing something changing HTTPSettings wouldn't help with that?

Initially this pr was for `asyncio.TimeoutError` only, it later got extended to include `aiohttp.ClientOSError`, which I wasn't talking about in that comment.

And yes, when the timeout is exceeded, aiohttp raises a timeout error.

The comment I made before is about a change I worked on but never made (which only thought about timeout errors), since I abandoned it.

FasterSpeeding (approved):

davfsa:
Personally dont like the current state of the PR, will comment on it later :)

FasterSpeeding:
It needs rebasing now anyways lol

hikari/impl/rest.py
```diff
@@ -680,7 +680,7 @@ def __exit__(
             return None
 
     @typing.final
-    async def _request(
+    async def _request(  # noqa: C901 CFQ001
```
davfsa, line 683:
```suggestion
    async def _request(  # noqa: C901 CFQ001
```
If these are ignored, they should also be documented above

hikari/impl/rest.py
```diff
@@ -745,18 +745,39 @@ async def _request(
...
-                        headers=headers,
-                        params=query,
-                        json=json,
-                        data=form,
-                        allow_redirects=self._http_settings.max_redirects is not None,
-                        max_redirects=self._http_settings.max_redirects,
-                        proxy=self._proxy_settings.url,
-                        proxy_headers=self._proxy_settings.all_headers,
-                    )
+                    try:
```
davfsa, line 748: The try except should be taken out of here and put before the `with` (in the base of this PR, a try already exists there, but it is not there in the current state of master, so a rebase will first be necessary)

hikari/impl/rest.py
```diff
@@ -745,18 +745,39 @@ async def _request(
...
+                    ):  # Request timed out or failed due to connection issues
+                        if retry_count >= self._max_retries:
+                            raise
+
+                        if backoff is None:
+                            backoff = rate_limits.ExponentialBackOff(maximum=_MAX_BACKOFF_DURATION)
+
+                        sleep_time = next(backoff)
+                        _LOGGER.warning(
+                            "Timed out on request, backing off for %.2fs and retrying. Retries remaining: %s",
```
davfsa, line 773: Since we catch `aiohttp.ClientOSError` too, this is no longer true

davfsa (changes requested):
@Le0Developer would you mind showing me some context on this PR? Mostly these issues:
- Under what situations are you seeing the `asyncio.TimeoutError`? And wouldnt it be nicer to just tweak your `hikari.impl.HTTPSetting.timeouts` and tweak the values there (potentially only `total`)
  - This already defaults to 30 seconds, so the fact that you are seeing so many of them is a bit worrying. Might want to double check the network to the raspberry pi is stable (try using an ethernet cable directly to the router, if not done already)
- `aiohttp.ClientOSError` is a really weird case and not something i have seen in a while, specially on linux. Mind trying to upgrade hikari + aiohttp to their respective latest versions and give it another shot?

*Sorry that this took so long, I kinda forgot about this pr 😅*

Le0Developer:
> Might want to double check the network to the raspberry pi is stable (try using an ethernet cable directly to the router, if not done already)

My raspberry pi is directly plugged into my router, it just happens *sometimes* (possibly happens when internet connection is shortly cutoff?).

> `aiohttp.ClientOSError` is a really weird case and not something i have seen in a while, specially on linux.

I only have the discord notification of sentry, since sentry deletes them after 90d:
image

davfsa:
> My raspberry pi is directly plugged into my router, it just happens sometimes (possibly happens when internet connection is shortly cutoff?).

After a 30s timeout, i think the issue might be bigger than just a simple timeout, but regardless, maybe this is worth implementing for the few it would affect

> I only have the discord notification of sentry, since sentry deletes them after 90d

Hmmm. Under normal functions you should be receiving this, but i guess we could handle it for the odd case

davfsa:
Closing in favor of https://github.com/hikari-py/hikari/pull/1648

PR #1134, opened 2022-04-13
Allow async iterators in delete_messages

### LINKED ISSUE #1127: Allow async iterators in `delete_messages`
### Summary

`rest.delete_messages` should allow the use of both sync and async/lazy iterators.

### Why is this needed?

Users should not be forced to chunk iterators themselves if they want to delete and fetch the history at the same time.

### Ideal implementation

Instead of having a list of pending messages, it would be better to cast the iterable into a LazyIterator and use `iterator.chunk()`. While this may reduce the performance by a few nanoseconds it will improve user experience.

thesadru:
Made `rest.delete_messages` allow the use of both sync and async/lazy iterators.

closes #1127


hikari/impl/rest.py
```diff
@@ -1446,26 +1446,25 @@ async def delete_messages(
...
         route = routes.POST_DELETE_CHANNEL_MESSAGES_BULK.compile(channel=channel)
 
-        pending: typing.List[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = []
         deleted: typing.List[snowflakes.SnowflakeishOr[messages_.PartialMessage]] = []
 
-        if isinstance(messages, typing.Iterable):  # Syntactic sugar. Allows to use iterables
-            pending.extend(messages)
-
+        if isinstance(messages, typing.AsyncIterable):
+            iterator = iterators.NOOPLazyIterator(messages)
```
thesadru, line 1460: maybe add a `LazyIterator.__new__` that can return the appropriate iterator?

iterable -> flat lazy iterator
async iterable -> noop iterator
lazy iterator -> identical

thesadru: Probably outside the scope of this PR

hikari/impl/rest.py
```diff
@@ -1495,14 +1495,12 @@ async def delete_messages(
...
                     body = data_binding.JSONObjectBuilder()
-                    chunk = pending[:100]
                     body.put_snowflake_array("messages", chunk)
                     await self._request(route, json=body)
                     deleted += chunk
 
-                pending = pending[100:]
             except Exception as ex:
-                raise errors.BulkDeleteError(deleted, pending) from ex
+                raise errors.BulkDeleteError(deleted, await iterator) from ex
```
thesadru, line 1503: Might potentially be cool to optimize this part further.

hikari/impl/rest.py
```diff
@@ -1481,8 +1481,8 @@ async def delete_messages(
             # (albeit at a cost of maybe a couple-dozen milliseconds per call),
             # I am just gonna invoke these sequentially instead.
             try:
-                if len(pending) == 1:
-                    message = pending[0]
+                if len(chunk) == 1:
+                    message = chunk[0]
                     try:
                         await self.delete_message(channel, message)
                     except errors.NotFoundError as exc:
```
GoogleGenius, line 1488: Not a specific problem with this PR, just wondering why this uses except as `exc` vs `ex` below...

GoogleGenius: Saw this in other files too but whatever.

thesadru: Was already there when I got to it. I'd personally use `e` but oh well.

GoogleGenius: Since 39767a9 should this be changed across the whole lib? Preferably in another PR obviously.

thesadru: does it even matter that much? It's just a small detail

GoogleGenius: True but now there's 3 conventions for errors lol

thesadru: L

hikari/impl/rest.py
```diff
@@ -1481,28 +1481,26 @@ async def delete_messages(
...
             # I am just gonna invoke these sequentially instead.
             try:
-                if len(pending) == 1:
-                    message = pending[0]
+                if len(chunk) == 1:
+                    message = chunk[0]
                     try:
                         await self.delete_message(channel, message)
-                    except errors.NotFoundError as exc:
+                    except errors.NotFoundError as e:
```
davfsa, line 1488: `e` gives no context to what it may refer to.

Please change it back to `exc` or `ex`, which is at least an acronym to "exception"

hikari/impl/rest.py
```diff
@@ -1481,28 +1481,26 @@ async def delete_messages(
...
-                    chunk = pending[:100]
                     body.put_snowflake_array("messages", chunk)
                     await self._request(route, json=body)
                     deleted += chunk
 
-                pending = pending[100:]
-            except Exception as ex:
-                raise errors.BulkDeleteError(deleted, pending) from ex
+            except Exception as e:
+                raise errors.BulkDeleteError(deleted, await iterator) from e
```
davfsa, line 1503: Got a little problem with this, as we are effectively consuming the whole iterator. We no longer know it has a finate number of items, so this could potentially go on for ever, which would cause it to never return/raise the exception

tests/hikari/impl/test_rest.py
```diff
@@ -2668,6 +2669,31 @@ async def test_delete_messages_with_iterable(self, rest_client):
...
+                ),
+                mock.call(
+                    routes.POST_DELETE_CHANNEL_MESSAGES_BULK.compile(channel=channel),
+                    json={"messages": ["100", "101", "102"]},
+                ),
+            ]
+        )
+
+    async def test_delete_messages_with_async_iterable_and_args(self, rest_client):
+        with pytest.raises(TypeError):
```
davfsa, line 2694: Should match the error string to make sure the error raised is the one you expect

hikari/iterators.py
```diff
@@ -815,6 +815,18 @@ async def __anext__(self) -> ValueT:
             self._complete()
 
 
+class NOOPLazyIterator(typing.Generic[ValueT], LazyIterator[ValueT]):
+    """A lazy iterator that uses an underlying async iterator and does nothing."""
+
+    __slots__: typing.Sequence[str] = "_iterator"
```
davfsa, line 821:
```suggestion
    __slots__: typing.Sequence[str] = ("_iterator",)
```

hikari/iterators.py
```diff
@@ -815,6 +815,18 @@ async def __anext__(self) -> ValueT:
             self._complete()
 
 
+class NOOPLazyIterator(typing.Generic[ValueT], LazyIterator[ValueT]):
```
davfsa, line 818: Instead of a NOOPLazyIterator, LazyIterator could be a no-op by default instead.

I have already been thinking about splitting async LazyIterators from sync LazyIterators which would help with cases like the CacheView iterator needing to be awaited even if it is sync.

This would be part of a greater revamp of the LazyIterators, so might be worth waiting for that to be done (by me or someone else interested in taking a look) instead of half-doing it here.

hikari/impl/rest.py
```diff
@@ -1499,8 +1499,12 @@ async def delete_messages(
                     await self._request(route, json=body)
                     deleted += chunk
 
-            except Exception as e:
-                raise errors.BulkDeleteError(deleted, await iterator) from e
+            except Exception as ex:
+                skipped = list(chunk)
+                if isinstance(iterator, iterators.FlatLazyIterator):
+                    skipped += await iterator
```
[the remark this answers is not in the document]
thesadru: I wanted to avoid a breaking change but oh well.

How should the current chunk be handled though? Somehow get it back into the iterator?

FasterSpeeding (approved):

thesadru:
woo

PR #1137, opened 2022-04-13
Fix docs examples for hikari.Permissions

### LINKED ISSUE #1133: Fix hikari.Permissions bitwise operation examples
### Summary

In the docs, the description for an IntFlag enum gives this example:

```
if my_perms & (Permissions.MANAGE_CHANNELS | Permissions.MANAGE_GUILD):
    print("I have the permission to both manage the guild and the channels in it!")
```

The expression in the example evaluates to true when my_perms contains either permission, not both permissions.

### Why is this needed?

Some of the examples in the docs are misleading.

Like the description for the hikari.Intents enum, the description for hikari.Permissions should make it more clear how combining the & and == operators allows you to check that the entire intersection of two sets of enums are in another set:

```
if my_perms & (Permissions.MANAGE_CHANNELS | Permissions.MANAGE_GUILD) == Permissions.MANAGE_CHANNELS | Permissions.MANAGE_GUILD):
    print("I have the permission to both manage the guild and the channels in it!")
```

I think a more clear and concise example would be:

```
if my_perms & required_perms == required_perms
    print("I have all of the required permissions!")
```

Also the docs said that If the permission is not present, the bitwise xor will return a non-zero value, otherwise if it is present, it will return 0.
Actually the bitwise xor operation only returns a nonzero value whenever both operands are not exactly the same, and the example in the docs will only work if both my_perms and required_perms are made of just 1 permission.

### Ideal implementation

To demonstrate how to check whether all the required permissions are present, put something like
```
if my_perms & required_perms == required_perms
    print("I have all of the required permissions!")
```

Bitwise Xor:
For hikari.Permissions, I can't really think of a practical application. Technically if we wanted to fix the example in the docs we would do:

```
missing_perms = (my_perms ^ required_perms) & required_perms
if (missing_perms):
    print(f"I'm missing these permissions! {missing_perms}")
```

But maybe this should just be removed because this is a really specific example, and we're not trying to explain how bitwise operators work. lmk what you think thanks

### Checklist

lilapela:
### Summary

Fix the errors in the examples, and make the explanations more clear and concise

### Related issues

Closes #1133

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -29,7 +29,7 @@ sphobjinv==2.2.2
 #################
 
 mypy==0.942
-pyright==1.1.236
+pyright==1.1.237
 
 #######################
 # DEPENDENCY CHECKING #
--- a/hikari/permissions.py
+++ b/hikari/permissions.py
@@ -41,7 +41,8 @@ class Permissions(enums.Flag):
     where each bit represents a permission. You can use bitwise operators
     to efficiently manipulate and compare permissions.
 
-    Examples:
+    Examples
+    --------
     You can create an enum which combines multiple permissions using the bitwise OR operator (`|`):
 
        my_perms = Permissions.MANAGE_CHANNELS | Permissions.MANAGE_GUILD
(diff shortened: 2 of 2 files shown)


davfsa (approved):
Thanks!

davfsa (approved):

PR #1148, opened 2022-04-29
Application permissions v2

### LINKED ISSUE #1145: Implement Permissions v2
### Summary

Discord has recently announced the new [Permissions v2](https://discord.com/blog/slash-commands-permissions-discord-apps-bots) system for Application Commands. This is a formal request for it to be implemented in Hikari.

### Why is this needed?

This adds new features to the Permissions system specifically for application commands including being able to use Discord Permission bitfields, disable commands per user, channel, or role, and remove global commands from being used in DMs, etc. 

More documentation is [here](https://github.com/discord/discord-api-docs/pull/4830)

### Ideal implementation

N/A

### Checklist

davfsa:
### Checklist

### Related issues
Close #1145


hikari/applications.py
```diff
@@ -137,15 +137,18 @@ class OAuth2Scope(str, enums.Enum):
     """
 
     APPLICATIONS_COMMANDS = "applications.commands"
-    """Allows your application's (slash) commands to be used in a guild.
+    """Allows your application's application commands to be used in a guild.
```
thesadru, line 140: Might as well just only talk about an application's commands instead of application commands. 

Does it have to be your application anyways?

davfsa: Yes, it is only that applications commands it can modify:
image

And that is a typo, will fix it. Thanks for pointing it out

hikari/api/rest.py
```diff
@@ -6917,6 +6925,17 @@ async def create_slash_command(
 
             Defaults to `builtins.True`.
 
+            .. deprecated:: 2.0.0.dev109
+                Use `default_member_permissions` instead.
+        default_member_permissions : hikari.undefined.UndefinedOr[hikari.permissions.Permissions]
+            Member permissions necessary to utilize this command by default.
+
+            If `0`, then it will be disabled by default.
```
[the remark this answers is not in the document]
GoogleGenius: As outlined by the Discord API docs, the commands by default are disabled except for administrators if set to `0`.

GoogleGenius (changes requested):

hikari/applications.py
```diff
@@ -137,15 +137,18 @@ class OAuth2Scope(str, enums.Enum):
...
     `OAuth2Scope.BOT` in-order to join an application into a guild as an
     application command providing integration.
     """
 
     APPLICATIONS_COMMANDS_UPDATE = "applications.commands.update"
-    """Allows your application to update it's (slash) commands via a bearer token."""
+    """Allows your application to update it's commands via a bearer token."""
+
+    APPLICATIONS_COMMANDS_PERMISSION_UPDATE = "applications.commands.permissions.update"
+    """Allows your application to update it's commands permissions via a bearer token."""
```
GoogleGenius, line 151:
```suggestion
    """Allows your application to update its commands permissions via a bearer token."""
```

GoogleGenius: Typo

hikari/api/rest.py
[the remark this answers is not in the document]
davfsa: These are defaults. They can of course be overwritten per server. But I dont think we need to document that specifically, we already clarify these are default values that can be modified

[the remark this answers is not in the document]
HyperGH: It may be a good idea to point out specifically still, as some users may incorrectly believe that this, for example, may allow for the creation of developer-only slash commands.

hikari/api/special_endpoints.py
```diff
@@ -1006,7 +1017,7 @@ def set_id(self: _T, id_: undefined.UndefinedOr[snowflakes.Snowflakeish], /) ->
 
     @abc.abstractmethod
     def set_default_permission(self: _T, state: undefined.UndefinedOr[bool], /) -> _T:
-        """Whether this command should be enabled by default (without any permissions).
+        """Set whether this command should be enabled by default (without any permissions).
```
GoogleGenius, line 1020: I am not sure, does this disable commands regardless of any permission (including disabling for admins here)?

hikari/api/rest.py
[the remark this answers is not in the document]
GoogleGenius: > It may be a good idea to point out specifically still, as some users may incorrectly believe that this, for example, may allow for the creation of developer-only slash commands.

Discord also updated their documentation later to point this out as well (though it really should have done that in the first place). I still recommend to do the same here for clarity purposes.

hikari/api/special_endpoints.py
davfsa: Yep, but this is deprecated and will be removed. I havent modified the logic for it in this PR

GoogleGenius: Ah I see, thanks for letting me know!

hikari/applications.py
```diff
@@ -137,15 +137,18 @@ class OAuth2Scope(str, enums.Enum):
...
+    """Allows your application's commands to be used in a guild.
 
     This is used in Discord's special Bot Authorization Flow like
     `OAuth2Scope.BOT` in-order to join an application into a guild as an
     application command providing integration.
     """
 
     APPLICATIONS_COMMANDS_UPDATE = "applications.commands.update"
-    """Allows your application to update it's (slash) commands via a bearer token."""
+    """Allows your application to update it commands via a bearer token."""
```
GoogleGenius, line 148:
```suggestion
    """Allows your application to update its commands via a bearer token."""
```

hikari/applications.py
```diff
@@ -137,15 +137,18 @@ class OAuth2Scope(str, enums.Enum):
...
+
+    APPLICATIONS_COMMANDS_PERMISSION_UPDATE = "applications.commands.permissions.update"
+    """Allows your application to update it commands permissions via a bearer token."""
```
GoogleGenius, line 151:
```suggestion
    """Allows your application to update its commands permissions via a bearer token."""
```

GoogleGenius (changes requested):
This is still grammatically incorrect.

hikari/applications.py
davfsa: I swear I only remove the `'`. What happened here

thesadru (dismissed):
Looks great

GoogleGenius (approved):

hikari/api/rest.py
```diff
@@ -3275,6 +3275,9 @@ async def add_user_to_guild(
             Requires the `MANAGE_NICKNAMES` permission on the guild.
         nick : hikari.undefined.UndefinedOr[builtins.str]
             Deprecated alias for `nickname`.
+
+            .. deprecated:: 2.0.0.dev106
+                Use `nickname` instead.
         roles : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishSequence[hikari.guilds.PartialRole]]
             If provided, the roles to add to the user when he joins the guild.
```
FasterSpeeding, line 3282: you need to document the new restrictions around using these endpoints

FasterSpeeding (changes requested):

hikari/impl/rest.py
```diff
@@ -3341,6 +3356,7 @@ async def fetch_application_command_permissions(
         assert isinstance(response, dict)
         return self._entity_factory.deserialize_guild_command_permissions(response)
 
+    @deprecation.deprecated("2.0.0.dev112", "set_application_command_permissions")
```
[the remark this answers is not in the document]
davfsa: Breaking change it is then :)

hikari/api/rest.py
```diff
@@ -6983,6 +7004,17 @@ async def create_context_menu_command(
 
             Defaults to `builtins.True`.
 
+            .. deprecated:: 2.0.0.dev109
+                Use `default_member_permissions` instead.
```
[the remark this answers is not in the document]
davfsa: Would clarifying that you can use `default_member_permissions` instead to get the same functionality better?

hikari/api/rest.py
davfsa: What restrictions?

hikari/api/rest.py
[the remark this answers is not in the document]
HyperGH: But it isn't the same functionality, only similar, unless I understand this incorrectly.

hikari/api/rest.py
FasterSpeeding: oh put that on the wrong line
but for the PUT permissions endpoint you need to document that its not a bot endpoint anymore and can must be used through the oauth2 flow with the perms scope.

hikari/impl/rest.py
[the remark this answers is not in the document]
FasterSpeeding: technically its not really a breaking change on our behalf if Discord already broke it.
Like either way this is going to error lol, you're just switching out a guaranteed BadRequestError for a guaranteed AttributeError (at least under normal usage)

Also i kinda suspect this method was a pretty cold path anyways

hikari/api/rest.py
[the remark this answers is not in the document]
FasterSpeeding: They're not even really similar tbh: v1 let bots control who can use their commands with `default_permission` deciding whether it could be used without explicit auth from the bot and v2 lets server moderators decide who can use commands and just lets bots suggest the initial perms; its a total inversion of v1 control wise and v2 just can't be used to achieve the behaviour of v1.

also random point but v2 broke default_permission (introducing a security hole for bots relying on it) by making it a suggestion rather than a strict rule so keeping it around is a bit iffy.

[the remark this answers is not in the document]
davfsa: Ok, will remove it then. Unfortunate it has to be this way :c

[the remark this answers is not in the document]
FasterSpeeding: Yeah its really annoying that discord did this with zero public notice but it is how it is

hikari/api/rest.py
```diff
@@ -3275,6 +3275,9 @@ async def add_user_to_guild(
             Requires the `MANAGE_NICKNAMES` permission on the guild.
         nick : hikari.undefined.UndefinedOr[builtins.str]
             Deprecated alias for `nickname`.
+
+            .. deprecated:: 2.0.0.dev106
+                Use `nickname` instead.
         roles : hikari.undefined.UndefinedOr[hikari.snowflakes.SnowflakeishSequence[hikari.guilds.PartialRole]]
             If provided, the roles to add to the user when he joins the guild.
             This may be a collection objects or IDs of existing roles.
```
FasterSpeeding, line 3283: (ignore the code this is pointing to, just picked a random spot)
Can we also document special cases such as https://github.com/discord/discord-api-docs/pull/4872 and https://discord.com/developers/docs/interactions/application-commands#application-command-permissions-object-application-command-permissions-constants for the perm endpoints and ig also the perm object maybe?

hikari/api/rest.py
```diff
@@ -6889,6 +6895,8 @@ async def create_slash_command(
         guild: undefined.UndefinedOr[snowflakes.SnowflakeishOr[guilds.PartialGuild]] = undefined.UNDEFINED,
         options: undefined.UndefinedOr[typing.Sequence[commands.CommandOption]] = undefined.UNDEFINED,
         default_permission: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
+        default_member_permissions: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
```
FasterSpeeding, line 6898: Probably want to also allow int for these

hikari/api/special_endpoints.py
```diff
@@ -982,11 +982,23 @@ def default_permission(self) -> undefined.UndefinedOr[bool]:
...
-        Returns
-        -------
-        undefined.UndefinedOr[builtins.bool]
-            Whether the command should be enabled by default (without any permissions).
+    @property
+    @abc.abstractmethod
+    def default_member_permissions(self) -> undefined.UndefinedOr[permissions_.Permissions]:
+        """Member permissions necessary to utilize this command by default.
+
+        If `0`, then it will be disabled by default. This excludes administrators
```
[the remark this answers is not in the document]
GoogleGenius: But really an UNDEFINED in this case should be the default `@everyone` permission without overwrites. Unless I'm misunderstanding you?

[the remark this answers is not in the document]
GoogleGenius: I see. `0` is admins, etc. with the API correct? So you're suggesting to hide this behavior and take `8` for admins instead?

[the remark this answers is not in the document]
davfsa: > Did you ever consider trying to hide this behaviour on a library level considering that it just doesn't really make any sense since it just makes setting it to 0 equal to setting it to 8 while leaving setting it to UNDEFINED as the real 0?

Yea, that is exactly what it is at the moment. Its just the edge case of 0 basically meaning 8 that I would document (as have done Discord)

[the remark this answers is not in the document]
FasterSpeeding: > > Did you ever consider trying to hide this behaviour on a library level considering that it just doesn't really make any sense since it just makes setting it to 0 equal to setting it to 8 while leaving setting it to UNDEFINED as the real 0?
> 
> Yea, that is exactly what it is at the moment. Its just the edge case of 0 basically meaning 8 that I would document (as have done Discord)

> Did you ever consider trying to hide this behaviour on a library level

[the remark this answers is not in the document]
GoogleGenius: Honestly, I don't think it's a good idea as it abstracts too much from the API docs and info. Even though technically it should be the same, it could be confusing when comparing the behavior of Hikari vs discord API docs.

hikari/api/rest.py
```diff
@@ -7081,6 +7030,10 @@ async def edit_application_command(
         name: undefined.UndefinedOr[str] = undefined.UNDEFINED,
         description: undefined.UndefinedOr[str] = undefined.UNDEFINED,
         options: undefined.UndefinedOr[typing.Sequence[commands.CommandOption]] = undefined.UNDEFINED,
+        default_member_permissions: typing.Union[
```
FasterSpeeding, line 7033: unless you special case this so that Permissions.NONE is special cased to be treated as null here then you're going to have to also allow None to be passed here

hikari/api/special_endpoints.py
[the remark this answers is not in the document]
davfsa: After revisiting this PR this much time later I have realized how funky that behavior could be and we might as well account for it ourselves. Changed the behavior to account for Discord's interpretation of `default_member_permissions` being 0.

hikari/api/rest.py
```diff
@@ -6852,7 +6852,7 @@ async def create_slash_command(
         default_member_permissions : typing.Union[hikari.undefined.UndefinedType, int, hikari.permissions.Permissions]
             Member permissions necessary to utilize this command by default.
 
-            If `0`, then it will be disabled by default. This excludes administrators
+            If `0`, then it will be available for all members. This excludes administrators
```
GoogleGenius, line 6855: I think this might need to be rewritten clearer, this confused me.

As with the other occurrences.

hikari/impl/special_endpoints.py
```diff
@@ -1191,9 +1206,7 @@ async def create(
         *,
         guild: undefined.UndefinedOr[snowflakes.SnowflakeishOr[guilds.PartialGuild]] = undefined.UNDEFINED,
     ) -> commands.ContextMenuCommand:
-        return await rest.create_context_menu_command(
-            application, self._type, self._name, guild=guild, default_permission=self._default_permission
-        )
+        return await rest.create_context_menu_command(application, self._type, self._name, guild=guild)
```
FasterSpeeding, line 1209: Don't you need to pass through `default_member_permissions` and `dm_enabled` here?

hikari/impl/special_endpoints.py
```diff
@@ -1165,7 +1181,6 @@ async def create(
             self._name,
             self._description,
             guild=guild,
-            default_permission=self._default_permission,
```
FasterSpeeding, line 1168: Don't you need to pass through `default_member_permissions` and `dm_enabled` here?

hikari/impl/special_endpoints.py
davfsa: Nice catch, I completely forgot to do so. Fixed now!

FasterSpeeding (approved):

PR #1155, opened 2022-05-10
Remove Tanjun from README.md

FasterSpeeding:
It's just wasting space there and shouldn't be there

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,6 @@ implementing your own command handler.
 Hikari does not include a command framework by default, so you will want to pick a third party library to do it:
 
 - [`lightbulb`](https://github.com/tandemdude/hikari-lightbulb) - a simple and easy to use command framework for Hikari.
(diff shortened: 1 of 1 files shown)


AlexanderHOtt:
I still use tanjun, so I see no reason to remove it if there is still a userbase.

GoogleGenius:
I am also actively developing and dependent on the Tanjun framework. Please reconsider this.

thesadru:
Tanjun is essential to the Hikari ecosystem. Lightbulb targets a completely different demographic.

davfsa (approved):
Considering it looks like you are still updating Tanjun, all I can do is ask for you keep it there. It takes absolutely no place, since its just a text file. There is nothing better that could be replacing it.

But then again, this is not my place to talk. All I can do is ask for you to keep it there. But if you want to remove it, its you lib, your call

I would love for it to stay :)

PR #1164, opened 2022-05-19
Make sure build methods are typed consistently

FasterSpeeding:
### Summary
Returning JSONObjectBuilder introduces issues for anybody who's overriding these methods as they'd have to type their return methods as JSONObjectBuilder which is an internal type.

And  MutableMapping is preferred over JSONObject for a return type

### Checklist

### DIFF
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -1208,7 +1208,7 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
             payload["author"] = author_payload
 
         if embed.fields:
-            field_payloads: list[data_binding.JSONObject] = []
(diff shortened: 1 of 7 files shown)


hikari/impl/entity_factory.py
```diff
@@ -1209,7 +1208,7 @@ def serialize_embed(  # noqa: C901 - Function too complex
             payload["author"] = author_payload
 
         if embed.fields:
-            field_payloads: data_binding.JSONArray = []
+            field_payloads: list[data_binding.JSONObject] = []
```
davfsa, line 1211:
```suggestion
            field_payloads: typing.List[data_binding.JSONObject] = []
```

davfsa (changes requested):

davfsa (approved):

PR #1179, opened 2022-06-02
Implement text in voice

HyperGH:
### Summary
Implement the ability to handle messages in voice channels, with the (upcoming-ish) release of text-in-voice.

### Checklist

### DIFF
--- a/hikari/channels.py
+++ b/hikari/channels.py
@@ -1288,7 +1288,7 @@ class GuildNewsChannel(TextableGuildChannel):
 
 
 @attr.define(hash=True, kw_only=True, weakref_slot=False)
-class GuildVoiceChannel(GuildChannel):
+class GuildVoiceChannel(TextableGuildChannel):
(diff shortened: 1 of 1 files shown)


hikari/channels.py
```diff
@@ -1288,7 +1288,7 @@ class GuildNewsChannel(TextableGuildChannel):
 
 
 @attr.define(hash=True, kw_only=True, weakref_slot=False)
-class GuildVoiceChannel(GuildChannel):
+class GuildVoiceChannel(TextableGuildChannel):
```
FasterSpeeding, line 1291: You need to add the `last_message_id` field to this class

FasterSpeeding: this still needs to be added

HyperGH: Whoops, looks like I completely forgot about this PR, should be resolved now.

FasterSpeeding (approved):

PR #1185, opened 2022-06-11
Switch docs to Sphinx

davfsa:
### Summary
https://www.youtube.com/watch?v=iCMcSx1as3c

It was finally a full moon Sunday and I got Sphinx working. I am not able to make or keep nice HTML as I was aiming for to use `pdoc`. Instead we can outsource that to the furo project and, with the sphinx ecosystem, we can directly add guides and what not to the docs.

This could be immediately merged into master without any problems. When releasing the new version, the `gh-pages` branch would have to be manually pushed with the landing page only and redirects for where the documentation used to be should be setup in Cloudflare

### Showcase
https://hikari-py--1185.org.readthedocs.build/en/1185/

image
image

### Checklist


GoogleGenius:
👍

While I did like the old docs for its design, I do agree with this change to improve them in many other aspects i.e. search, visibility, usability, guides, etc.

davfsa:
Would also like to note, the read the docs check that runs on PRs is only temporarily enabled to live-view these docs

FasterSpeeding:
Did you consider alternatives other than just sphinx?

davfsa:
> Did you consider alternatives other than just sphinx?

Trust me, sphinx was not the first thing I went for, mostly because how badly I remember having to deal with stupid bugs of things not liking correctly in the past. I have tried sphinx, mkdocs (w/ pdoc) and pure pdoc before trying with sphinx again (as we had in the past). None of them managed to really make go "yes, this is what I want". Each of them has their benefits and their drawbacks:

- Mkdocs (w/pdoc): It works nicely, because pdoc generates perfect documentation from the source, it was built for it. The addition of mkdocs makes it customizable, but I just dont like the themes that are available and I lack the knowledge to make one myself (as can be seen with the PRs before this)
- Pure Pdoc: Same benefits as Mkdocs, but the default theme doesnt do it for me and I completely failed to make one. There seems to be bugs around all the corners and I dont want to have to maintain it after it breaks each time. Additionally, there is no way to add custom pages, which makes things like a guide impossible
- Sphinx: Overall I don't like sphinx. It is clunky and bound to break at any moment as we have seen in the past. But after using it in other projects and being relatively successful I decided to try again. I found [sphinx-autoapi](https://github.com/readthedocs/sphinx-autoapi) some time ago and it worked wonders, like it does here. The only problem with it is that it is slow and rebuilds everything every time you run it. I have been trying to get rid of it as a dependency but still get the same advantages as having it, but not amazing results so far (still trying tho). Sphinx has a lot of themes available (like furo) that provide an amazing light/dark documentation view for accessibility reasons and is actively maintained, meaning any bugs can be fixed upstream, which fixes my incapability of making and maintaining a theme. Furthermore, the sphinx environment makes adding an extension and it just working (most of the time) very helpful. Lastly, we can add any amount of `.rst` or `.md` files for guides, extra pages or what not

FasterSpeeding:
Why mcdocs with pdoc over mkdocstrings?

davfsa:
> Why mcdocs with pdoc over mkdocstrings?

First time I hear about mkdocstrings ngl. Will have a look at it and compare :)

GoogleGenius:
Mkdocs Material too? Or is there a different theme you used before.

davfsa:
> Why mcdocs with pdoc over mkdocstrings?

I played around with mkdocstrings for a bit the past couple of days and the results are on Discord, but I am not overly happy with the outcome, so I would rather stick with sphinx for now. Mkdocs did look really nice, but I dont think will meet the cut for now

hikari/api/event_factory.py
```diff
@@ -300,7 +300,7 @@ def deserialize_guild_update_event(
 
         Other Parameters
         ----------------
-        old_guild : typing.Optional[hikari.guilds.GatewayGuild]
+        old_guild  : typing.Optional[hikari.guilds.GatewayGuild]
```
davfsa, line 303: One too many here

dev-requirements.txt
```diff
@@ -21,9 +21,13 @@ async-timeout==4.0.2  # Used for timeouts in some test cases.
 # DOCUMENTATION #
 #################
 
-pdoc==11.0.0
-sphobjinv==2.2.2
-minify-html==0.8
+sphinx==5.0.1
+sphinx-autoapi==1.8.4
```
davfsa, line 25: Attempt to get rid of this dependency

davfsa (commented):
Just some notes for me

dev-requirements.txt
PerchunPak: Better do not delete `sphinx-autoapi`. Build-in auto API tool requires write reference to every module by hand. And then it's importing everything for documentation, while `sphinx-autoapi` parsing source code and automaticly find all source files, as expected from such tool.

You can generate *stubs* for build-in API tool on every run, so no need to reference in docs every module. **BUT** this will **import** the code plus it's too slow and often breaks.

But `sphinx-autoapi` also have some dissadvantages, like unsuporting redefing (althought mypy also doesn't support it).

davfsa: I left this purely as a note to remind me what to finish. I was making an alternative to it that works better in some parts of the code that this currently breaks, mainly enums (off the top of my head), but I still got to play around with it a bit

davfsa:
For the time being, I am going to be going with these docs. In the future, I would love to switch Mkdocs with the use of Mkdocstrings, as I find the documentation clearer and the whole environment nicer to work with.

Unfortunately, that is a bit further down the line as there is no attrs support, ending in not so nice docs. I'll keep an eye out for when third party extensions are supported to allow for it and (eventually) move to Mkdocs!

For now, sphinx will have to suffice.

PR #1186, opened 2022-06-12
Dont increase/decrease group counts for always enabled consumers

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/event_manager_base.py
+++ b/hikari/impl/event_manager_base.py
@@ -343,7 +343,22 @@ class _Consumer:
 
     @property
     def is_enabled(self) -> bool:
+        """Whether this consumer is enabled."""
         return self.is_caching or self.listener_group_count > 0 or self.waiter_group_count > 0
+
+
+@attr.define(weakref_slot=False)
+class _AlwaysEnabledConsumer(_Consumer):
+    # By using 0 here, we avoid increasing/decreasing listener_group_count and/or waiter_group_count
+    # when a new group gets added/removed.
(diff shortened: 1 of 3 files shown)


hikari/impl/event_manager_base.py
```diff
@@ -343,7 +343,22 @@ class _Consumer:
...
     def is_enabled(self) -> bool:
-        return self.is_caching or self.listener_group_count > 0 or self.waiter_group_count > 0
+        """Whether this consumer is enabled."""
+        return self.is_caching or self.listener_group_count or self.waiter_group_count
+
+
+@attr.define(weakref_slot=False)
+class _AlwaysEnabledConsumer(_Consumer):
+    # By using 0 here, we avoid increasing/decreasing listener_group_count and/or waiter_group_count
+    # when a new group gets added.
```
davfsa, line 353:
```suggestion
    # when a new group gets added/removed.
```

FasterSpeeding:
What's the benefit of this change/reason for it?

davfsa:
If you were to currently look at the consumer classes for unfiltered events like `GUILD_CREATE` and look at the `listener_group_count` and `waiter_group_count`, the number reflects how many global listener and waiter group counts there are respectively, as it is always increased.

The easy fix for this would have been:
```py
_Consumer(member, 0, True)
```
Which tbh might actually be a better diff, but I thought that a new class would make this more explicit.

Could change to that instead tho

davfsa:
After further consideration, I think I will just keep it like it is

PR #1187, opened 2022-06-13
Change SelectMenuBuilder min_value to 1 by default

supershadoe:
### Summary
Changed the default min_value in SelectMenuBuilder from 0 to 1.

As per Discord API docs, the default value of min_value is 1 while the least it can go to is 0. (Also, when using select menu on phone, before this PR, the select options would appear with a checkbox near it if set_min_value() isn't used although the max_value is 1.)

### Checklist

### Related issues
Found no issue related to this.

### DIFF
--- a/changes/1187.bugfix.md
+++ b/changes/1187.bugfix.md
@@ -1,0 +1,1 @@
+Changed the default `min_value` in `SelectMenuBuilder` from 0 to 1.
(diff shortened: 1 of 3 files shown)


davfsa:
Discord API defaults are not the same as the ones we impose. If we wanted them to be, we would exclude it from the payload we send.

You mentioned that this makes it not work nicely on Android, but is the same behaviour in desktop and would it just be worth reporting it as a bug to Discord?

PR #1195, opened 2022-06-26
Rename `edit_permission_overwrites` to `edit_permission_overwrite`

GoogleGenius:
### Summary

**Deprecation**. Renames `edit_permission_overwrites` to `edit_permission_overwrite`.

### Checklist

### Related issues

https://discord.com/channels/574921006817476608/700378161526997003/979493994092449802


parafoxia:
Pretty sure "overwrites" is correct here. The phrase "edit permission overwrite" makes no sense.

Lunarmagpie:
> Pretty sure "overwrites" is correct here. The phrase "edit permission overwrite" makes no sense.

The discord API docs also refers to it as "permission overwrites"

GoogleGenius:
> Pretty sure "overwrites" is correct here. The phrase "edit permission overwrite" makes no sense.

Hmm I thought and agreed with davfsa that "overwrite" made more sense because it edits a specific entry if I understand the method correctly.

GoogleGenius:
> > Pretty sure "overwrites" is correct here. The phrase "edit permission overwrite" makes no sense.
> 
> The discord API docs also refers to it as "permission overwrites"

This form is used throughout hikari as well but it's this specific method that edits only a single entry I believe for an overwrite?

GoogleGenius:
I'll add the fragment later by the way.

davfsa (commented):
This would need deprecation of the old method instead of removal to not turn it into a breaking change. Fragment file should update to reflect that too

davfsa:
> Pretty sure "overwrites" is correct here. The phrase "edit permission overwrite" makes no sense.

There is currently an inconsistency in the lib with this name, with one method calling it overwrites and another one overwrite. Since you are editing a specific permission overwrite, I think the singular is more fitting.

This was fact a todo of mine that seems to have been snipped, lol

GoogleGenius:
> This was fact a todo of mine that seems to have been snipped, lol

Heh it was, I just thought I could do it anyway haha.

GoogleGenius:
@davfsa Should the tests should use the old name or new name?

changes/1195.deprecation.md
```diff
@@ -0,0 +1,2 @@
+`RESTClient.edit_permission_overwrites` deprecated
+ - Renamed to `RESTClient.edit_permission_overwrite`
```
GoogleGenius, line 2: Let me know if you want this to be more specific or wording changed.

GoogleGenius:
Nvm about review, this is failing CI, let me fix.

Jonxslays:
> @davfsa Should the tests should use the old name or new name?

The old method deprecated method and the new method should both be tested up until the deprecated method is removed.

GoogleGenius:
> > @davfsa Should the tests should use the old name or new name?
> 
> The old deprecated method and the new method should both be tested up until the deprecated method is removed.

What about methods that use that method? They can just be updated with the new name and the test changed right?

parafoxia:
> > Pretty sure "overwrites" is correct here. The phrase "edit permission overwrite" makes no sense.
> 
> There is currently an inconsistency in the lib with this name, with one method calling it overwrites and another one overwrite. Since you are editing a specific permission overwrite, I think the singular is more fitting.
> 
> This was fact a todo of mine that seems to have been snipped, lol

Don't you have the ability to edit multiple overwrites with the same method though?

GoogleGenius:
> > > Pretty sure "overwrites" is correct here. The phrase "edit permission overwrite" makes no sense.
> > 
> > 
> > There is currently an inconsistency in the lib with this name, with one method calling it overwrites and another one overwrite. Since you are editing a specific permission overwrite, I think the singular is more fitting.
> > This was fact a todo of mine that seems to have been snipped, lol
> 
> Don't you have the ability to edit multiple overwrites with the same method though?

Hmm, now I'm a bit conflicted here. It makes sense as singular when you look at it from an "overwrite entry" but if you're editing multiple permissions, you can use the bitwise `|` operator on the `hikari.Permissions` enum. Then it would be `edit_overwrite_permissions` rather.

GoogleGenius:
I'm also looking at how discord.js and discord.py named this and utterly confused 😕

davfsa:
A permission overwrite is the singular object that consists of a target, what permissions are allowed and which ones are denied.

Since this endpoint only edits a specific one, then `edit_permission_overwrite` makes more sense than `edit_permission_overwrites`, as you are only editing one.

I currently can't check and I don't know it of the top of my head, but the method to bulk ediy them is a different one @parafoxia

changes/1195.deprecation.md
davfsa: Could just be simplified to: ```py
```suggestion
`RESTClient.edit_permission_overwrites` renamed to `RESTClient.edit_permission_overwrite`
```

hikari/api/rest.py
```diff
@@ -530,6 +530,107 @@ async def edit_permission_overwrites(
...
+        target_type: typing.Union[channels_.PermissionOverwriteType, int],
+        allow: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
+        deny: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
+        reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
+    ) -> None:
+        """Edit permissions for a given entity ID and type.
+
+        .. deprecated:: 2.0.0.dev110
+            Use `RESTClient.edit_permission_overwrite` instead.
+        """
```
davfsa, line 632: Since they are deprecated, this could be removed from here and the docstrings could be placed in `impl.rest`, as then people dont need to necessarily implement these functions when subclassing

parafoxia:
> A permission overwrite is the singular object that consists of a target, what permissions are allowed and which ones are denied.
> 
> Since this endpoint only edits a specific one, then `edit_permission_overwrite` makes more sense than `edit_permission_overwrites`, as you are only editing one.
> 
> I currently can't check and I don't know it of the top of my head, but the method to bulk ediy them is a different one @parafoxia 

Right yeah, I'm seeing what you mean. I'd always taken a "permission overwrite" to be an overwrite of a specific permission, rather than all permissions for a given entity.

parafoxia:
@GoogleGenius I've spotted a typo which might as well be fixed in this PR. The docs (existing and updated) say "vale" instead of "value" in the allowed and denied fields.

GoogleGenius:
> @GoogleGenius I've spotted a typo which might as well be fixed in this PR. The docs (existing and updated) say "vale" instead of "value" in the allowed and denied fields.

Thanks, will fix.

davfsa (approved):

FasterSpeeding (dismissed):

hikari/impl/rest.py
```diff
@@ -1078,6 +1078,78 @@ async def edit_permission_overwrites(
...
+            snowflakes.Snowflakeish, users.PartialUser, guilds.PartialRole, channels_.PermissionOverwrite
+        ],
+        *,
+        target_type: undefined.UndefinedOr[typing.Union[channels_.PermissionOverwriteType, int]] = undefined.UNDEFINED,
+        allow: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
+        deny: undefined.UndefinedOr[permissions_.Permissions] = undefined.UNDEFINED,
+        reason: undefined.UndefinedOr[str] = undefined.UNDEFINED,
+    ) -> None:
+        """Edit permissions for a specific entity in the given guild channel.
+
```
FasterSpeeding, line 1095: This doc string doesn't really need to be this long considering the only important info is that this is a deprecated alias of `edit_permission_overwrites`

FasterSpeeding (approved):

davfsa (approved):

PR #1204, opened 2022-07-06
Improve components lifetime

davfsa:
# Short summary

Bit of a rewrite of hikari internals to improve the lifetime, stability and consistency of the components.

# Extended summary

## Breaking changes
- `GatewayBot.join`'s `until_close` argument removed.
- `GatewayShardImpl.get_user_id` is no longer async and will now always be available.

## Bugfixes
- `GatewayShardImpl` can now be instantiated out of an async environment for consistency with other components.
- Correct signal handling in `RESTBot`.

## Improvements
- `GatewayBot`:
  - General speedups.
  - Fix a lot of edge cases of hard crashes if the application shuts unexpectedly.
  - More consistent signal handling.
  - `run`' `shard_ids` argument can now be a `typing.Sequence`.
  - Improved logging.
- `RESTBot`:
  - Consistent signal handling inline with `GatewayBot`.
  - Improved logging.
  - Improved loop closing.
- `GatewayShardImpl`:
  - New `is_connected` property to determine whether the shard is connected to the gateway.
  - Faster websocket pulling and heartbeating.
  - Improved error handling.
  - New gateway reconnect logic to account for `resume_gateway_url`.
  - Rate limiting changes:
    - Chunking no longer has its own special ratelimit. Now it is shared with the rest of
    "non-priority" packages sent, which is of 117/60s (3 less than the hard limit).
      - "priority" packages currently only include hearbeating.

---

*To @FasterSpeeding: I didnt realize until now how many "small fixes" there are in this pr, which makes it a bit hard to review. If you want, I could make smaller separate PRs for some of the issues if they are better for you to review*


davfsa:
Implemented https://github.com/discord/discord-api-docs/pull/5282 as part of https://github.com/hikari-py/hikari/pull/1204/commits/5c1b5e7a1a6e847ca9693b4d53711d870e9b4d41.

If this PR were not to be merged before the 7th of August, I will open another smaller PR to merge directly into master (or if @FasterSpeeding rathers)

And bump for review :)

davfsa:
@FasterSpeeding Bump :)

changes/1204.feature.md
```diff
@@ -0,0 +1,20 @@
+Improve components lifetimes:
+- `GatewayBot`:
+  - General speedups.
+  - Fix a lot of edge cases of hard crashes if the application shuts unexpectedly.
+  - More consistent signal handling.
+  - `run`' `shard_ids` argument can now be a `typing.Sequence`.
```
FasterSpeeding, line 6: typo

hikari/traits.py
```diff
@@ -560,133 +560,6 @@ class GatewayBotAware(
...
-        activity: typing.Optional[presences.Activity] = None,
-        afk: bool = False,
-        close_passed_executor: bool = False,
-        idle_since: typing.Optional[datetime.datetime] = None,
-        ignore_session_start_limit: bool = False,
-        large_threshold: int = 250,
-        status: presences.Status = presences.Status.ONLINE,
-        shard_ids: typing.Optional[typing.AbstractSet[int]] = None,
-        shard_count: typing.Optional[int] = None,
-    ) -> None:
```
FasterSpeeding, line 594: this is a pretty important breaking change which should be in the changelog, also why was this removed?

davfsa: For consistency with `RESTBotAware`. They are both now `Runnable` and how their lifetime in regards to parameters to `start`, `run`, `join` and `close` are left up to implementation detail.

I forgot to add it to the changelog, will do so

FasterSpeeding: For consistency with a class where all the run and start args were considered impl specific you removed the generic run and start args from the trait for a class where some of the run and start args were considered a part of the abstract interface since they were discord detail rather than bot impl detail?

davfsa: They do not have to be Discord detail. They are left up to however the developers want to define them. The GatewayBotAware trait just specifies all the components it should have and thats about it. There is no reason to enforce this as abstraction detail here when it comes to how it should run

FasterSpeeding: > There is no reason to enforce this as abstraction detail here when it comes to how it should run.

There is? so ppl can run or start a hikari gateway bot (in-memory shard manager) generically while still being able to control what it declares to Discord on startup (which is what most of these args are for).

Also the field `close_passed_executor` make sense there for the run method since these are both `ExecuterAware` (meaning they manage an executor internally) and I'd argue missing that field for `RESTBotAware.run` was a mistake.

hikari/impl/shard.py
```diff
@@ -519,61 +516,63 @@ def intents(self) -> intents_.Intents:
...
+        except asyncio.CancelledError:
+            pass
+
+        self._keep_alive_task = None
+        self._non_priority_rate_limit.close()
+        self._total_rate_limit.close()
+        self._is_closing = False
+        self._logger.info("shard shutdown successfully")
+
+    def get_user_id(self) -> snowflakes.Snowflake:
```
FasterSpeeding, line 552: what's the point of this breaking change?

davfsa: With the new lifetime, we can ensure it will always be available, so there is no need for it to be async anymore

hikari/traits.py
davfsa: But they don't need to necessarily manage that in the run method. They could be done other ways. I just don't see the need for them to be enforced like this here

hikari/impl/shard.py
FasterSpeeding: >         If the shard has not connected fully yet, this should wait until the ID
>         is set before returning.


The old code was just waiting until the shard had started to return the user (and returning if the shard was already active); how are you ensuring that it's available before the shard has started?

hikari/traits.py
FasterSpeeding: > I just don't see the need for them to be enforced like this here

So ppl can manage them generically, like not everyone just locks their stuff to 1 implementation of Hikari's gateway bot and without that you just won't be able to set some of these fields generically and others you'll need to send another message over the gateway on start-up which is kind of geh.

Also ppl can just not implement GatewayBot specifically if these do not apply to them, like I already do cause GatewayBot also enforces having a cache and none of my implementations have caches.

Also you like enforce that Shard has a start and close method even though not all Shard implementations will be "started" like that.

hikari/impl/shard.py
davfsa: With the new lifetime, we can ensure that after `start` completes, we will have the information available and there wont be an attempt to access it until we return from `start`, which is when it would be reasonably expected for the information to be available. There is no need for the function to be async, as you shouldnt be waiting for the user_id if you never start the shard.

hikari/traits.py
davfsa: In the same way we dont enforce what the parameters to start, join and close a shard or a `RESTBot` should be, we shouldnt be enforcing it for `GatewayBot` either. Our API standard says that they should be runnable, but it stops there.

With the example you gave before of `ExecutorAware`, we never enforce how that executor should be handled, we just type that an ExecutorAware object will have a executor property.

FasterSpeeding: > In the same way we dont enforce what the parameters to start, join and close a shard or a RESTBot should be, we shouldnt be enforcing it for GatewayBot either. Our API standard says that they should be runnable, but it stops there.

There's a big distinction here though, all gateway bots will do stuff like declare an activity on startup. Rest bots don't have anything like that which is why it doesn't have any extra run or start arguments.

> With the example you gave before of ExecutorAware, we never enforce how that executor should be handled, 

Wdym, we were enforcing it and now you're removing it.

hikari/impl/shard.py
FasterSpeeding: > There is no need for the function to be async, as you shouldnt be waiting for the user_id if you never start the shard.

Ppl could call that method before the shard has started, this is why it was async before to wait for the shard to start. Idk what you're getting at here.

hikari/traits.py
FasterSpeeding: > In the same way we dont enforce what the parameters to start, join and close a shard

Even just having the method on Shard is enforcing behaviour, its the same thing as having optional arguments. A start method like this would make no sense for a Shard impl which just wraps around a shared message queue as an example and yet its still enforced. Start methods also aren't how you'd want to handle this for a gateway bot impl or shard impl which targets trio or anyio and having start methods on the interface like this is exposing asyncio specific async structuring.

davfsa: We have as an API design for Shard to be be Runnable (this includes having a `start`, `join` and `close`). I am aware that you can have shards without those methods, but by our API design, they should. If someone doesn't want to use this API design, they can define and make their own.

But just like we enforce a shard being runnable, we don't enfoce it to have any specific arguments. This is done in Shard, RESTBotAware, and now GatewayBotAware.

I don't think we should enfoce those arguments at a trait level, which is why I  removed them in this pr.

FasterSpeeding: > We have as an API design for Shard to be be Runnable (this includes having a start, join and close). I am aware that you can have shards without those methods, but by our API design, they should. If someone doesn't want to use this API design, they can define and make their own.

Ok so that wouldn't be compatible with Hikari at all so you're just saying "don't use Hikari" here and also why does it need to be a part of the interface? is there even any reason for shard to be directly runnable on the abstract interface when this it even breaks compatibility with the standard gateway bot impl let alone custom implementations.

> But just like we enforce a shard being runnable, we don't enfoce it to have any specific arguments. This is done in Shard, RESTBotAware, and now GatewayBotAware.

Also that's just wrong, the only reason no args are enforced for shard.start is because it just straight out doesn't take any args on the standard implementation (since api `shard.start` is just a copy of the standard implementation's start method) and for `RESTBotAware.run` the reason is because I just missed the close executor arg when i implement it.

PR #1206, opened 2022-07-07
Remove field validation from `serialize_embed`

davfsa:
This validation only checks fields and not all the other possibilities that Discord will complain about. It is best to just leave it to Discord to decide whether to error or not.

### DIFF
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -1215,35 +1215,9 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
             payload["author"] = author_payload
 
         if embed.fields:
-            field_payloads: typing.List[data_binding.JSONObject] = []
-            for i, field in enumerate(embed.fields):
-
-                # Yep, these are technically two unreachable branches. However, this is an incredibly
-                # common mistake to make when working with embeds and not using a static type
-                # checker, so I have added these as additional safeguards for UX and ease
-                # of debugging. The case that there are `None` should be detected immediately by
-                # static type checkers, regardless.
-                name = str(field.name) if field.name is not None else None
-                value = str(field.value) if field.value is not None else None
-
-                if name is None:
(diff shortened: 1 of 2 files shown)


FasterSpeeding:
You won't get an error response from discord in the interaction server flow

davfsa:
> You won't get an error response from discord in the interaction server flow, could move this logic to the response builders instead maybe? 

I personally dislike builders a tiny bit because, for other languages it would be great, but for python it just adds overhead, so I really don't know.

Maybe it could be added there but, and I havent played around with an interaction server yet so idk, if Discord don't reply with error codes, handing this one specific case for embeds and not all the other things Discord could raise I as a bit unnecessary. We might as well not handle anything and skip all these checks all the time.

FasterSpeeding:
> I personally dislike builders a tiny bit because, for other languages it would be great, but for python it just adds overhead, so I really don't know.

Adds overhead compared to what?

> Maybe it could be added there but, and I havent played around with an interaction server yet so idk, if Discord don't reply with error codes, handing this one specific case for embeds and not all the other things Discord could raise I as a bit unnecessary. We might as well not handle anything and skip all these checks all the time, as it would be more in-line with the rest of the lib, don't you think?

It being more inline with the rest of the lib would be bad ux there; the plan has always been to add some level of validation to the server response builders as you don't get a response to your initial response and therefore will get no info/feedback on why it failed (just this is really low priority).

davfsa:
> Adds overhead compared to what?

Sorry, wrote that as I woke up and idk what I was thinking. Embed is basically a builder anyways, so yea, the validation could be moved there

> It being more inline with the rest of the lib would be bad ux there; the plan has always been to add some level of validation to the server response builders as you don't get a response to your initial response and therefore will get no info/feedback on why it failed (just this is really low priority). 

Yea, there it would make sense there. My bad

PR #1207, opened 2022-07-09
Remove case of bang (`!`) from `Member.mention` property

GoogleGenius:
### Summary

Discord mentions should not have the `!` bang anymore per deprecation by Discord.

### Checklist

### DIFF
--- a/changes/1207.breaking.md
+++ b/changes/1207.breaking.md
@@ -1,2 +1,1 @@
-`Member.mention` now returns a mention string without the `!` bang regardless of nickname setting.
- - Example: `<@1234567890>` instead of `<@!1234567890>`
+Removed case of `Member.mention` returning bang (!) mention, as it is deprecated by Discord.
--- a/changes/1207.breaking.md
+++ b/changes/1207.breaking.md
@@ -1,1 +1,1 @@
-Removed case of `Member.mention` returning bang (!) mention, as it is deprecated by Discord.
+Removed case of `Member.mention` returning bang (`!`) mention, as it is deprecated by Discord.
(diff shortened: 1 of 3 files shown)


changes/1207.breaking.md
```diff
@@ -0,0 +1,2 @@
+`Member.mention` now returns a mention string without the `!` bang regardless of nickname setting.
+ - Example: `<@1234567890>` instead of `<@!1234567890>`
```
davfsa, line 2: This is not really necessary

And would this be considered a breaking change even?

GoogleGenius: I am not sure exactly about this. I thought this was old behavior by Discord that should not be there anymore. I was encountering some issues pertaining to this and noticed this here in hikari. 

> And would this be considered a breaking change even?

It modifies the return value of the mention by removing the bang. If this was implemented, it could be a breaking change as it may perhaps change the behavior of comparisons. I don't know how different clients behave...

Guess this isn't needed to be modified.

GoogleGenius: Resolved in Discord

hikari/guilds.py
```diff
@@ -486,26 +486,20 @@ def is_system(self) -> bool:
...
-        '<@!123456789123456789>'
         ```
 
         Returns
         -------
         builtins.str
             The mention string to use.
         """
-        return f"<@!{self.id}>" if self.nickname is not None else self.user.mention
+        return self.user.mention
```
davfsa, line 502: Member inherits from User, so this property could be removed all together from here and it will just fallback to User.mention

davfsa (changes requested):

hikari/guilds.py
GoogleGenius: I thought so too, however I thought it would be more consistent with the other properties: https://github.com/hikari-py/hikari/blob/b56b79c0a2c4ccfa10dab51ff169cb926c518307/hikari/guilds.py#L404-L549

davfsa: Oh, nvm, just remembered why we do it this way. You can keep it, maybe the docstring is what can go only

PR #1213, opened 2022-07-15
Improve BadRequestError error string

davfsa:
### Checklist

### DIFF
--- a/hikari/errors.py
+++ b/hikari/errors.py
@@ -285,6 +285,25 @@ class ClientHTTPResponseError(HTTPResponseError):
     """
 
 
+def _dump_errors(obj: data_binding.JSONObject, obj_string: str = "") -> str:
+    string = ""
+    for key, value in obj.items():
+        if isinstance(value, typing.Sequence):
+            string += obj_string + ":"
+
+            for item in value:
+                string += f"\n - {item['message']}"
+
+            string += "\n\n"
+
+            continue
+
+        current_obj_string = f"{obj_string}{'.' if obj_string else ''}{key}"
+        string += _dump_errors(value, current_obj_string)
+
+    return string
+
+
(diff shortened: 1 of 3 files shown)


hikari/errors.py
```diff
@@ -308,7 +308,27 @@ def __str__(self) -> str:
 
         value = super().__str__()
         if self.errors:
-            value += "\n" + data_binding.dump_json(self.errors, indent=2)
+            value += "\n\n"
+
+            def dump(obj: data_binding.JSONObject, obj_string: str = "") -> str:
```
FasterSpeeding, line 313: Does this function have to be defined inline and can thus logic just be merged into the parent function?

hikari/errors.py
```diff
@@ -308,7 +308,27 @@ def __str__(self) -> str:
...
+                string = ""
+                for k, v in obj.items():
+                    if isinstance(v, typing.Sequence):
+                        errors_string = obj_string + ":"
+
+                        for item in v:
+                            errors_string += f"\n - {item['message']}"
```
FasterSpeeding, line 320: Might be worth it not assuming the error will always be in this format and instead falling back to a more basic handling

hikari/errors.py
davfsa: Extracted the logic

hikari/errors.py
davfsa: Added a fallback for when a KeyError occurs to just dump the errors JSON (the previous behaviour). Personally I think that the errors structure is well defined enough in the docs for this not be an issue (can also be seen in the wild), but better safe than sorry ig.

Also, TIL that `str` failing when raising an exception will just add `str failed` as the str of the exception

FasterSpeeding (approved):

PR #1221, opened 2022-07-19
move mention property from `GuildChannel` to `PartialChannel`

AlexanderHOtt:
### Summary
Discord allows mentioning of all channels through the `<#id>` syntax, so it makes sense to move the mention property to the base class.

### Checklist

### DIFF
--- a/hikari/channels.py
+++ b/hikari/channels.py
@@ -333,7 +333,9 @@ class PartialChannel(snowflakes.Unique):
 
         !!! note
             As of writing, GuildCategory, GroupDMChannel, and DMChannel channels are
-            a special case for this and mentions of them will not resolve as clickable.
+            a special case for this. Mentions of them will not resolve as clickable
+            on all platforms, a "#" followed by the channel's name is rendered on the 
+            client instead.
 
         Returns
         -------
--- a/hikari/channels.py
+++ b/hikari/channels.py
@@ -327,6 +327,21 @@ class PartialChannel(snowflakes.Unique):
     type: typing.Union[ChannelType, int] = attr.field(eq=False, hash=False, repr=True)
     """The channel's type."""
 
+    @property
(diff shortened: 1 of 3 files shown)


hikari/channels.py
```diff
@@ -327,6 +327,21 @@ class PartialChannel(snowflakes.Unique):
     type: typing.Union[ChannelType, int] = attr.field(eq=False, hash=False, repr=True)
     """The channel's type."""
 
+    @property
+    def mention(self) -> str:
+        """Return a raw mention string for the channel.
+
+        !!! note
+            As of writing, GuildCategory, GroupDMChannel, and DMChannel channels are
+            a special case for this and mentions of them will not resolve as clickable.
```
Le0Developer, line 336: Doesn't this depend on the platform?

> Tbh, might as well move it to PartialChannels and add that GuildCategory and DmChannel don't render on all platforms correctly
> Android seems to work with everything
> Even renders categories
> But link just takes you to the first channel in the server

https://discord.com/channels/574921006817476608/644130410564943873/998865770970886194

davfsa: We are not going to through the different apps and check how they function. This is ok for now

AlexanderHOtt: updated it

GoogleGenius: I still prefer @Le0Developer's note better

AlexanderHOtt: sure, i'll change it

davfsa (approved):

davfsa:
Missing a fragment file @AlexanderHOtt. Mind adding it?

FasterSpeeding (approved):

davfsa (dismissed):

davfsa (approved):

davfsa (approved):

PR #1225, opened 2022-07-21
Add application commands badge

HyperGH:
### Summary

Add the new application commands badge to `ApplicationFlags`.

### Checklist

### DIFF
--- a/hikari/applications.py
+++ b/hikari/applications.py
@@ -96,7 +96,7 @@ class ApplicationFlags(enums.Flag):
     """Denotes that the application has message content access while pending verification."""
 
     APPLICATION_COMMAND_BADGE = 1 << 23
-    """Denotes that the application supports app commands and has the corresponding badge."""
+    """Denotes that the application has at least one global application command, and thus has the corresponding badge."""
 
 
 @typing.final
--- a/hikari/applications.py
+++ b/hikari/applications.py
@@ -96,7 +96,7 @@ class ApplicationFlags(enums.Flag):
     """Denotes that the application has message content access while pending verification."""
 
     APPLICATION_COMMAND_BADGE = 1 << 23
-    """Denotes that the application has at least one global application command, and thus has the corresponding badge."""
+    """Denotes that the application has at least one global application command."""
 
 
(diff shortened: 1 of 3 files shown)


hikari/applications.py
```diff
@@ -95,6 +95,9 @@ class ApplicationFlags(enums.Flag):
     MESSAGE_CONTENT_INTENT_LIMITED = 1 << 19
     """Denotes that the application has message content access while pending verification."""
 
+    APPLICATION_COMMAND_BADGE = 1 << 23
+    """Denotes that the application supports app commands and has the corresponding badge."""
```
davfsa, line 99:
```suggestion
    """Denotes that the application has at least one application command."""
```

davfsa (changes requested):
#

hikari/applications.py
HyperGH: If we want to be exact, then it should be "has at least one **global** application command".

hikari/applications.py
```diff
@@ -95,6 +95,9 @@ class ApplicationFlags(enums.Flag):
     MESSAGE_CONTENT_INTENT_LIMITED = 1 << 19
     """Denotes that the application has message content access while pending verification."""
 
+    APPLICATION_COMMAND_BADGE = 1 << 23
+    """Denotes that the application has at least one global application command, and thus has the corresponding badge."""
```
davfsa, line 99:
```suggestion
    """Denotes that the application has at least one global application command."""
```
 The rest of the information is redundant. It is already implied in the context

davfsa (changes requested):

davfsa (approved):

FasterSpeeding:
Btw might've jumped the gun on merging this just a tad bit considering this is an undocumented feature and that announcement said they were "testing" this feature

davfsa:
Oh shoot. I am currently not home and I saw the announcement and just assumed they had rolled it out. My bad

HyperGH:
The announcement mentions that it was activated yesterday for everyone, so I assumed it was okay to make a PR for it.

PR #1230, opened 2022-07-29
Simplify and improve rest lifetime

davfsa:
### Summary
- Move global ratelimit to bucket manager
- RESTApp now persists TCP and Client connections, as well as shared bucket manager
- Skip ratelimits for interactions `/callback` endpoint, due to it having 0 ratelimit (to improve memory)
- Add example of oauth server

### Checklist


davfsa:
@FasterSpeeding tests still need to be done and will be off for the next couple of days, so dont think ill be able to work much, but would appreciate it if i could get a couple of eyes on it and some general feedback :)

FasterSpeeding (approved):
Can't find anything obviously out of place here but this feels like a case that'll need to be battle tested in the real world so looks good to me

PR #1245, opened 2022-08-10
New gateway resume logic

davfsa:
### Checklist

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -880,13 +880,28 @@ class GatewayShardImpl(shard.GatewayShard):
 
         exit_stack = contextlib.AsyncExitStack()
 
-        query = {"v": _VERSION, "encoding": "json"}
+        url_parts = urllib.parse.urlparse(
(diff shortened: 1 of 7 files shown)


davfsa:
Looks like this follows the revamped gateway documentation, so should be good to merge

hikari/impl/shard.py
```diff
@@ -882,13 +880,21 @@ async def _run_once(self) -> bool:
 
         exit_stack = contextlib.AsyncExitStack()
 
+        query = {"v": _VERSION, "encoding": "json"}
+        if self._transport_compression:
+            query["compress"] = "zlib-stream"
+
+        url = self._resume_gateway_url or self._gateway_url
+        scheme, netloc, path, params, _, _ = urllib.parse.urlparse(url, allow_fragments=True)
+        encoded_query = urllib.parse.urlencode(query)
```
davfsa, line 889: It is possible for `resume_gateway_url` to contain a query, so we should append to the already existing ones instead of ignoring them

PR #1247, opened 2022-08-15
More precise snowflake datetime conversion

davfsa:
### Summary
Change to a more precise and reliable snowflake datetime conversion to fix float precision issues.

Special thanks to @rexor12 for investigating and providing the initial fix!

### Checklist

### DIFF
--- a/hikari/internal/time.py
+++ b/hikari/internal/time.py
@@ -52,12 +52,10 @@ This is an alias for `typing.Union[int, float, datetime.datetime]`,
 where `builtins.int` and `builtins.float` types are interpreted as a number of seconds.
 """
 
-DISCORD_EPOCH: typing.Final[datetime.datetime] = datetime.datetime.fromtimestamp(
-    1_420_070_400, tz=datetime.timezone.utc
-)
+DISCORD_EPOCH: typing.Final[datetime.timedelta] = datetime.timedelta(seconds=1_420_070_400)
 """Discord epoch used within snowflake identifiers.
 
-This is defined as the number of seconds between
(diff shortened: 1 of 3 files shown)


hikari/internal/time.py
```diff
@@ -134,7 +136,7 @@ def datetime_to_discord_epoch(timestamp: datetime.datetime) -> int:
     builtins.int
         Number of milliseconds since `1/1/2015 00:00:00 UTC`.
     """
-    return int((timestamp.timestamp() - DISCORD_EPOCH) * 1_000)
+    return int((timestamp - DISCORD_EPOCH).total_seconds() * 1_000)
```
FasterSpeeding, line 139: Did you consider just using timedeltas for these

hikari/internal/time.py
```diff
@@ -52,7 +52,9 @@
 where `builtins.int` and `builtins.float` types are interpreted as a number of seconds.
 """
 
-DISCORD_EPOCH: typing.Final[int] = 1_420_070_400
+DISCORD_EPOCH: typing.Final[datetime.datetime] = datetime.datetime.fromtimestamp(
+    1_420_070_400, tz=datetime.timezone.utc
+)
 """Discord epoch used within snowflake identifiers.
 
 This is defined as the number of seconds between
```
FasterSpeeding, line 60: if you're going to make this a datetime rather than a timedelta then this docstring needs to be updated

hikari/internal/time.py
davfsa: Changed to it :+1:

FasterSpeeding (approved):

PR #1260, opened 2022-08-18
Support v10 attachments edits

### LINKED ISSUE #713: Add interface for re-using previously uploaded attachments in message edits
### Summary
While editing a message previously uploaded Attachments can be referenced by name within embeds or by ID within the top level "attachments" array, while attachments re-use is implicitly possible within an embed context by using files.URL where the url is set to `attachments/{file_name}` this is not possible for top level attachments and is not documented behaviour nor suggested.

As a note, reusing attachments on a top-level within a message edit call looks like the following
```py
{
	"attachments": [
		{
			"id": "862815827518423080"
		}
	]
}
```

### Problem
Avoiding re-uploading attachments in message edits.

### Ideal implementation
TBD

### Checklist

davfsa:
### Checklist

### Related issues
Closes #713


changes/1260.breaking.md
```diff
@@ -0,0 +1,7 @@
+Support v10 attachments
+
+This includes breaking changes, features and things to look out for:
+- Modifying attachments in messages that contain embeds with any image attached to them now requires the images of that embed
+  image to be re-passed in the edit or they will be lost.
+- `replace_attachments` has been removed, as it is now the default.
+  - `attachment` and `attachments` now supports `None` to replicate the behaviour of fully removing all attachments.
```
davfsa, line 7:
```suggestion
  - `attachment` and `attachments` now supports `None` to replicate the behaviour of fully removing all attachments.
- `attachment` and `attachments` in message edits now support passing an `Attachment` object to keep existing attachments.
```

Note to self: Search replace affected message create too. Go ahead and revert those

hikari/impl/rest.py
```diff
@@ -1389,23 +1392,36 @@ def _build_message_payload(  # noqa: C901- Function too complex
...
             )
 
+        form_builder: typing.Optional[data_binding.URLEncodedFormBuilder] = None
         if final_attachments:
-            form_builder = data_binding.URLEncodedFormBuilder(executor=self._executor)
+            attachments_payload = []
+            attachment_id = 0
+
+            for f in final_attachments:
+                if isinstance(f, messages_.Attachment):
```
FasterSpeeding, line 1407: You prob need to update the interaction server's logic as well

davfsa: Good spot! Forgot about that

davfsa: Implemented!

FasterSpeeding (approved):

PR #1265, opened 2022-08-21
Add entry_point for extensions to show display version in cli

davfsa:
For any extension wanting to make use of this, add the following to your `setup.py` or equivalent:

```py
    entry_points={
        "hikari": ["extension_name = module_name"],
    },
```

The version is fetched from the `__version__` variable, accessed through the `module_name`

Note that this is only an suggestion for now and is bound to change until (and if) this pr is merged

### Checklist

### DIFF
--- a/hikari/cli.py
+++ b/hikari/cli.py
@@ -23,16 +23,25 @@
 """Provides the `python -m hikari` and `hikari` commands to the shell."""
 from __future__ import annotations
 
-import os
+import importlib.metadata
+import pathlib
 import platform
 import sys
+import typing
 
 from hikari import _about
 
 
+def _get_entry_points(group: str) -> typing.Optional[typing.Sequence[importlib.metadata.EntryPoint]]:
+    if sys.version_info >= (3, 10):
+        return importlib.metadata.entry_points(group=group)
+
+    return importlib.metadata.entry_points().get(group)
+
+
 def main() -> None:
     """Print package info and exit."""
-    path = os.path.abspath(os.path.dirname(_about.__file__))
+    path = pathlib.Path(_about.__file__).parent
     sha1 = _about.__git_sha1__[:8]
     version = _about.__version__
     py_impl = platform.python_implementation()
@@ -43,3 +52,14 @@ def main() -> None:
     sys.stderr.write(f"located at {path}\n")
     sys.stderr.write(f"{py_impl} {py_ver} {py_compiler}\n")
     sys.stderr.write(" ".join(frag.strip() for frag in platform.uname() if frag and frag.strip()) + "\n")
+
+    if extensions := _get_entry_points("hikari"):
+        sys.stderr.write("\nInstalled extensions:\n")
+
+        for extension in extensions:
+            name = extension.name
+            version = importlib.metadata.version(name)
+            module = extension.load()
+            module_path = pathlib.Path(module.__file__).parent
+
+            sys.stderr.write(f"  {name} ({version}) [{module_path}]\n")


Lunarmagpie:
Extension dev opinion here: I think this is a very good of doing this. I think you should go forward with this implementation.

hikari/cli.py
```diff
@@ -43,3 +44,13 @@ def main() -> None:
     sys.stderr.write(f"located at {path}\n")
     sys.stderr.write(f"{py_impl} {py_ver} {py_compiler}\n")
     sys.stderr.write(" ".join(frag.strip() for frag in platform.uname() if frag and frag.strip()) + "\n")
+
+    if extensions := importlib.metadata.entry_points().get("hikari"):
+        sys.stderr.write("\nInstalled extensions:\n")
+
+        for extension in extensions:
+            module = extension.load()
+            module_path = os.path.abspath(os.path.dirname(module.__file__))
```
FasterSpeeding, line 53: I would've used pathlib here personally lol but doesn't matter either eay

FasterSpeeding:
Also this would need to be documented somewhere (probably the readme or an extra doc page about extensions)

hikari/cli.py
```diff
@@ -43,3 +44,13 @@ def main() -> None:
...
+            version = getattr(module, "__version__", "unknown version")
```
FasterSpeeding, line 54: have you considered trying `importlib.metadata.version` here?

hikari/cli.py
```diff
@@ -43,3 +44,13 @@ def main() -> None:
     sys.stderr.write(f"located at {path}\n")
     sys.stderr.write(f"{py_impl} {py_ver} {py_compiler}\n")
     sys.stderr.write(" ".join(frag.strip() for frag in platform.uname() if frag and frag.strip()) + "\n")
+
+    if extensions := importlib.metadata.entry_points().get("hikari"):
```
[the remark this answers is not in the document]
davfsa: Unfortunately `.select` wasn't added until 3.10. Before, 3.10, it returned a normal dict.

Ig I'll have a version check for what call to make

davfsa:
> Also this would need to be documented somewhere (probably the readme or an extra doc page about extensions) 

I initially intended to make this a draft, just to get opinions on it. Will ofc add that :)

davfsa:
> Just wondering if you considered displaying extensions in the printout the Bot classes emit when they're initiated since that seems to be seen by more ppl? 

Would be nice too, will have a look

hikari/cli.py
davfsa: Initially did, but had a dumb mistake and couldnt get it work. Now i realized what that dumb mistake was and switched to it

Thanks for pointing it out and having me have another look at it!

hikari/cli.py
davfsa: I copy pasted this from above, but switched both to make use of pathlib

parafoxia:
Just had a peek at this, and it looks good! I can't see anything wrong with the current implementation.

parafoxia:
> Just wondering if you considered displaying extensions in the printout the Bot classes emit when they're initiated since that seems to be seen by more ppl? 

We'd also need to consider how best to handle this when it comes to custom banners too, as Crescent and Kasai both have their own in place of Hikari's.

If this was a config arg you could pass (which I imagine it probably would be), we'd need to decide how much that key actually handles, such as whether it would format them fully, or provide the option to have a custom delimiter, etc..

Lunarmagpie:
> Just wondering if you considered displaying extensions in the printout the Bot classes emit when they're initiated since that seems to be seen by more ppl?

I think this should be easy to support for custom banners. Just kindly ask all banners to have a new `$-substitution` that gets replaced with the extensions.

davfsa:
Closing for now, maybe revisited at some point in the future

PR #1271, opened 2022-08-25
Cleanup and merge twemoji mapping test to noxfile

davfsa:
### Checklist


davfsa:
Nvm, what i thought that would not be a problem might end up being one (importing hikari in the noxfile), so it will have to be separate then

PR #1272, opened 2022-08-26
Add ability to save resources

parafoxia:
### Summary
This PR adds the ability to save attachments (and more generally, resources) to disk.

* Adds the `Resource.save()` method
* Adds tests for `Resource.save()` and `Resource.read()`

### Checklist

### Related issues
None


hikari/files.py
```diff
@@ -473,6 +473,38 @@ async def read(
...
+            `builtins.None`, the resource will be saved in the
+            current working directory with the same name as it has on
+            Discord.
+        executor : typing.Optional[concurrent.futures.Executor]
+            The executor to run in for blocking operations.
+            If `builtins.None`, then the default executor is used for the
+            current event loop.
+        """
+        loop = asyncio.get_running_loop()
+        path = path or self.filename
```
parafoxia, line 501: I'm tempted to add some things here to (1) ensure the same filetype, and (2) account for directories being passed.

hikari/files.py
```diff
@@ -473,6 +473,38 @@ async def read(
         async with self.stream(executor=executor) as reader:
             return await reader.read()
 
+    async def save(
+        self,
+        *,
+        path: typing.Optional[typing.Union[pathlib.Path, str]] = None,
```
parafoxia, line 479: This should use Pathish.

hikari/files.py
```diff
@@ -473,6 +473,38 @@ async def read(
         async with self.stream(executor=executor) as reader:
             return await reader.read()
 
+    async def save(
```
davfsa, line 476: Maybe worth checking if the file exists before writing to it and erroring out if a `force` kwarg only arg is not set to True

hikari/files.py
davfsa: You can later just cast the path to `pathlib.Path` and use that directly.

`path.exists`, `path.expanduser`, `path.open` and so on.

tests/hikari/test_files.py
```diff
@@ -202,3 +205,83 @@ async def test_test_context_manager_when_target_is_dir(self):
...
+                    try:
+                        await attachment.save(path=path)
+                    except AttributeError as exc:
+                        if str(exc) != "'NotImplementedType' object has no attribute '__aexit__'":
+                            raise
+
+                    with open(path, "rb") as f:
+                        assert f.read() == data
+            finally:
+                path.unlink()
```
davfsa, line 287: You can make use of mocking here instead of writing to disk and having actual objects. Have a look at #1273 for example of this.

Furthermore, instead of creating an Attachment object, have a fixture like so:
```py
@pytest.fixture
def resource():
    class ResourceImpl(files.Resource):
        stream = mock.Mock()
        url = "https://some.where/file.txt"
        filename = "file.txt"

     return ResourceImpl()
```

I was going to write an example test for this to help you get started, but haven't been able to finish it. Lmk if you need any help

hikari/files.py
parafoxia: If a directory is passed the filename will need to be appended, no?

davfsa: Oh, that's what you mean. Yeah, you could make that an option. Should be easy enough with pathlib

hikari/files.py
parafoxia: Yeah that's a good idea.

hikari/files.py
parafoxia: Yeah I was thinking of doing that. I tend to handle it that way in my own projects.

hikari/files.py
```diff
@@ -473,6 +473,38 @@ async def read(
...
+        the entire resource into memory.
+
+        Parameters
+        ----------
+        path : typing.Optional[typing.Union[pathlib.Path, str]]
+            The path to save this resource to. If this is a string, the
+            path will be relative to the current working directory. If
+            `builtins.None`, the resource will be saved in the
+            current working directory with the same name as it has on
+            Discord.
```
davfsa, line 494:
```suggestion
            current working directory with the stored filename.
```

parafoxia: I think the original wording is better here. The meaning of "stored filename" is not particularly clear.

davfsa: The original wording is not really correct, as (for example) `URL` doesn't come from Discord. "stored filename" was refering to the `filename` attribute. Instead could just say that:

```suggestion
            current working directory under name in `filename`.
```

parafoxia: That's fair.

hikari/files.py
```diff
@@ -473,6 +473,47 @@ async def read(
...
+        path = path or self.filename
+
+        if not isinstance(path, pathlib.Path):
+            path = pathlib.Path(path)
+
+        if path.is_dir():
+            path = path / self.filename
+
+        if path.exists() and not force:
+            raise FileExistsError("file already exists; use `force=True` to overwrite")
```
davfsa, line 510: These are I/O blocking.

Instead, you can extract this logic to and run it all in an executor:

```py
def _open_file(path: pathlib.Path, filename: str, force: bool):
    if path.is_dir():
        path = path / self.filename

    if path.exists() and not force:
        raise FileExistsError(f"file {path} already exists; use `force=True` to overwrite")

    return path.open("wb")

# .....

with await loop.run_in_executor(executor, _open_file, path, self.filename, force):
   ...
```

hikari/files.py
```diff
@@ -473,6 +473,47 @@ async def read(
...
+        This writes the resource file in chunks, and so does not load
+        the entire resource into memory.
+
+        Parameters
+        ----------
+        path : typing.Optional[Pathish]
+            The path to save this resource to. If this is a string, the
+            path will be relative to the current working directory. If
+            `builtins.None`, the resource will be saved as
+            `self.filename` in the current working directory.
```
davfsa, line 494:
```suggestion
            `filename` in the current working directory.
```

tests/hikari/test_files.py
parafoxia: I'm not really sure tests you want me to do instead of these. I've looked at #1273 but am not getting a clear idea of how you want me to adapt them for this PR.

davfsa: There are 2 things:

1. Instead of creating an attachment object, you can just create a resource impl to use (codeblock above).
2. Instead of writing to disk, you can mock and make sure the correct things are called: https://github.com/hikari-py/hikari/blob/612b610fcf06bb1b335c02a7a098b2a1c7c1a0b2/tests/hikari/test_files.py#L74-L95

parafoxia: I understand that, I just can't visualise in my head *how* those would be achieved. Like I don't know how you'd probably test the features using mocks instead of writing to the disk.

hikari/files.py
```diff
@@ -473,6 +473,47 @@ async def read(
         async with self.stream(executor=executor) as reader:
             return await reader.read()
 
+    async def save(
+        self,
+        *,
+        path: typing.Optional[Pathish] = None,
+        executor: typing.Optional[concurrent.futures.Executor] = None,
+        force: bool = False,
```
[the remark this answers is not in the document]
parafoxia: Line 478 already does this.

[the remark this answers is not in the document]
davfsa: Oh, missed that. Maybe should be best to move it under `path` then, to allow for `resource.save("test.txt")`, but still force to use `resource.save("test.txt", force=True)`.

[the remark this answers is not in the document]
parafoxia: Yeah that might be an idea. I believe I did it like that to be consistent with how things are handled elsewhere, but I might've been imagining it. I'll verify either way when I can.

I agree that might be best, but elsewhere would need to be updated to be consistent if it is indeed like that elsewhere.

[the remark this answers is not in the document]
parafoxia: Turns out I was wrong, so can just apply this no probs.

davfsa:
> Honestly I think I'm just gonna have to leave the tests up to somebody else to take care of cos I can't figure this out. I've emulated what you've done, and I'm getting all sorts of errors and have absolutely no idea what I'm doing wrong.

@parafoxia mind pushing what you have (can be with broken tests too, or just skip them entirely).

parafoxia:
@FasterSpeeding, just bumping this as it's ready for review.

hikari/files.py
```diff
@@ -470,6 +480,40 @@ async def read(
...
+            The executor to run in for blocking operations.
+            If `None`, then the default executor is used for
+            the current event loop.
+        force : bool
+            Whether to overwrite an existing file. Defaults to `False`.
+        """
+        loop = asyncio.get_running_loop()
+        path = ensure_path(path or self.filename)
+
+        with await loop.run_in_executor(executor, _open_write_path, path, self.filename, force) as f:
```
FasterSpeeding, line 512: Closing a file can also be a blocking operation so you might want to do something like this instead:
```py
file = await loop.run_in_executor(executor, _open_write_path, path, self.filename, force)

try:
    ...

finally:
    await loop.run_in_executor(file.close)
```

hikari/files.py
```diff
@@ -470,6 +480,40 @@ async def read(
...
+            async with self.stream(executor=executor) as reader:
```
FasterSpeeding, line 513: just an idea but a specialisation of this which only makes one thread call to copy the file could prob be implemented specifically on the `File` resource implementation like the following:

```py
import shutil

def _to_write_path(path: pathlib.Path, default_filename: str, force: bool) -> typing.BinaryIO:
    if path.is_dir():
        path = path.joinpath(default_filename)

    if not force and path.exists():
        raise FileExistsError(f"file {path!r} already exists; use `force=True` to overwrite")

    return path


class File(...):
    async def _save_file(self, path, force):
        path = ensure_path(path or self.filename)
        shutil.copy2(self.path, _to_write_path(path, self.filename, force))

    async def save(self, path, executor, force):
        await loop.run_in_executor(executor, self._save_file, path, force)
```


This should optimise by avoiding multiple thread calls but also just better handles copying over the file's permissions and metadata (also shutil tries to use fast copies which avoids reading the data into memory).

hikari/files.py
```diff
@@ -470,6 +480,40 @@ async def read(
...
+        This writes the resource file in chunks, and so does not load
+        the entire resource into memory.
+
+        Parameters
+        ----------
+        path : typing.Optional[Pathish]
+            The path to save this resource to. If this is a string, the
+            path will be relative to the current working directory. If
+            `None`, the resource will be saved as `filename` in
+            the current working directory.
```
FasterSpeeding, line 501: Defaulting to saving as filename in the current working directory seems like odd behaviour, are you sure that's a good default behaviour to have?

hikari/files.py
```diff
@@ -470,6 +480,40 @@ async def read(
...
+        force : bool
+            Whether to overwrite an existing file. Defaults to `False`.
+        """
+        loop = asyncio.get_running_loop()
+        path = ensure_path(path or self.filename)
+
+        with await loop.run_in_executor(executor, _open_write_path, path, self.filename, force) as f:
+            async with self.stream(executor=executor) as reader:
+                async for chunk in reader:
+                    await loop.run_in_executor(executor, f.write, chunk)
```
FasterSpeeding, line 515: This could prob also be specialised for the Bytes impl (as an optimisation) to just save it all to the file at once instead of making multiple thread calls to chunk through it like the following
```py
class Bytes(...):
    async def _save_file(self, path: pathlib.Path, force: bool):
        _to_write_path(path, self.filename, force).write_bytes(self.data)

    async def save(self, path, executor, force):
        path = ensure_path(path or self.filename) 
        await loop.run_in_executor(executor, self._save_file, path, force)
```

hikari/files.py
davfsa: Forgot about that small detail, thanks for pointing it out!

davfsa: Changed!

hikari/files.py
davfsa: Implemented!

Thanks for the great idea!

hikari/files.py
davfsa: Again, thanks for the ideas!!

FasterSpeeding:
Test cases for the specialisations?

davfsa:
> Test cases for the specialisations?

Working on those atm :)

davfsa:
@FasterSpeeding tests done. I also found some bugs in the code, so some parts are update too :)

hikari/files.py
```diff
@@ -1072,6 +1142,22 @@ def stream(
...
         return _NoOpAsyncReaderContextManagerImpl(IteratorReader(self.filename, self.mimetype, self.data))
 
+    async def save(
+        self,
+        path: Pathish,
+        *,
+        executor: typing.Optional[concurrent.futures.Executor] = None,
+        force: bool = False,
+    ) -> None:
+        if not isinstance(self.data, bytes):
```
FasterSpeeding, line 1152: Could extend this to special case for bytearray and memoryview

davfsa: Expanded :)

FasterSpeeding (approved):

hikari/files.py
```diff
@@ -986,6 +1054,11 @@ def _assert_bytes(data: typing.Any) -> bytes:
         return data
 
 
+def _write_bytes(path: Pathish, default_filename: str, force: bool, data: bytes) -> None:
```
FasterSpeeding, line 1057:
```suggestion
def _write_bytes(
    path: Pathish, default_filename: str, force: bool, data: typing.Union[bytearray, bytes, memoryview]
) -> None:
```
Forgot this line needs to be updated as well

davfsa: Nice catch!

FasterSpeeding (approved):

PR #1273, opened 2022-08-27
Remove support for ProcessPoolExecutor executor when reading files

davfsa:
- Also cleans up code for reading using ThreadPool executors
 - It is more efficient to use a threadpool executor and it makes no sense to use anything else
   - Due to the nature of process pool, were were also not able to perform proper chunking when reading off the file

### Checklist

### DIFF
--- a/hikari/files.py
+++ b/hikari/files.py
@@ -36,7 +36,6 @@ __all__: typing.Sequence[str] = (
     "AsyncReaderContextManager",
     "Resource",
     "File",
-    "FileReader",
     "WebResource",
     "URL",
     "WebReader",
@@ -48,13 +47,11 @@ import abc
 import asyncio
 import base64
 import concurrent.futures
-import errno
 import inspect
 import io
 import mimetypes
 import os
 import pathlib
-import stat
 import typing
 import urllib.parse
 import urllib.request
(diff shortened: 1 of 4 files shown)


hikari/files.py
```diff
@@ -948,21 +866,23 @@ def stream(
...
 
         Parameters
         ----------
         executor : typing.Optional[concurrent.futures.Executor]
-            The executor to run the blocking read operations in. If
+            The thead executor to run the blocking read operations in. If
             `builtins.None`, the default executor for the running event loop
             will be used instead.
+
+            Only `concurrent.futures.Executor` is supported.
```
davfsa, line 879:
```suggestion
            Only `concurrent.futures.TheadPoolExecutor` is supported.
```

hikari/files.py
```diff
@@ -757,142 +754,63 @@ def filename(self) -> str:
...
         while True:
-            chunk = await loop.run_in_executor(self.executor, self._pointer.read, _MAGIC)
+            chunk = await loop.run_in_executor(self._executor, self._pointer.read, _MAGIC)
             yield chunk
             if len(chunk) < _MAGIC:
                 break
 
 
+def _open_path(path: pathlib.Path) -> typing.BinaryIO:
+    return path.expanduser().open("rb")
```
davfsa, line 779: Note for reviewers, `expanduser` will never raise a RuntimeError as stated in the docs and `open` will take care of opening directories with a `IsADirectory` error and unknown files with `FileNotFoundError`

FasterSpeeding (approved):

PR #1281, opened 2022-09-05
Call a context menu command a context menu command not a slash command.

ben2224:
### Summary
the above

### Checklist


davfsa (approved):
Thanks!

PR #1289, opened 2022-09-21
Deprecate CacheView.iterator in favour of using itertools

davfsa:
### Checklist

### DIFF
--- a/changes/1289.deprecation.md
+++ b/changes/1289.deprecation.md
@@ -1,0 +1,1 @@
+Deprecate `CacheView.iterator` in favour of using the `itertools` module.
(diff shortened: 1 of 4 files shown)


FasterSpeeding:
I mean this is still helpful by itself: its more ergonomic if you want to do an async operation over the entries in a cache using an async iterator and collect/iterate over the results rather than directly building an impl of them but it isn't necessary and is arguably misleading so I don't mind getting rid of it.

FasterSpeeding:
I mean this is still helpful by itself: its more ergonomic if you want to do an async operation over the entries in a cache using an async iterator and collect/iterate over the results rather than directly building an impl of them but it isn't necessary and is arguably misleading so I don't mind getting rid of it.

FasterSpeeding (approved):

davfsa:
That is the idea behind it. In async cache views, the iterator can still be added if needed :)

FasterSpeeding:
> That is the idea behind it. In async cache views, the iterator can still be added if needed :)

Fine by me then

PR #1324, opened 2022-10-24
Enable testing in CPython 3.11

davfsa:


davfsa:
Update README too to not show -dev for  the version

PR #1337, opened 2022-10-28
Catch UnrecognisedEntityErrors in rest list methods

FasterSpeeding:
### Checklist


hikari/impl/rest.py
```diff
@@ -3712,6 +3712,19 @@ async def fetch_application_command(
...
+    ) -> typing.List[commands.PartialCommand]:
+        command_objs: typing.List[commands.PartialCommand] = []
+        for payload in command_payloads:
+            try:
+                command_objs.append(self._entity_factory.deserialize_command(payload, guild_id=guild_id))
+
+            except errors.UnrecognisedEntityError:
+                pass
+
+        return command_objs
```
davfsa, line 3726: Why not allow extra values to pass to the function in `data_binding.cast_enum_array` so that we can get rid of this?

tests/hikari/impl/test_rest.py
```diff
@@ -2453,7 +2446,6 @@ async def test_fetch_pins(self, rest_client):
 
         assert await rest_client.fetch_pins(StubModel(123)) == [message1, message2]
         rest_client._request.assert_awaited_once_with(expected_route)
-        assert rest_client._entity_factory.deserialize_message.call_count == 2
```
davfsa, line 2456: Why remove these? We want to ensure it was called x amount of times

hikari/internal/data_binding.py
```diff
@@ -392,3 +394,33 @@ def put_snowflake_array(
...
+
+def cast_enum_array(cast: typing.Callable[[T_co], T], raw_values: typing.Iterable[T_co], /) -> typing.List[T]:
+    """Cast an array of enum variants while ignoring unrecognised variant types.
+
+    Parameters
+    ----------
+    cast : typing.Callable[[T_co], T]
+        Callback to cast each variant with.
+
+        This will ignore any variants which raies
```
davfsa, line 407:
```suggestion
        This will ignore any variants which raise
```

hikari/internal/data_binding.py
```diff
@@ -392,3 +394,33 @@ def put_snowflake_array(
...
             self[key] = [str(int(value)) for value in values]
+
+
+def cast_enum_array(cast: typing.Callable[[T_co], T], raw_values: typing.Iterable[T_co], /) -> typing.List[T]:
+    """Cast an array of enum variants while ignoring unrecognised variant types.
+
+    Parameters
+    ----------
+    cast : typing.Callable[[T_co], T]
+        Callback to cast each variant with.
```
davfsa, line 405:
```suggestion
        Callback to cast each variant to.
```

hikari/internal/data_binding.py
```diff
@@ -392,3 +394,33 @@ def put_snowflake_array(
         """  # noqa: E501 - Line too long
         if values is not undefined.UNDEFINED:
             self[key] = [str(int(value)) for value in values]
+
+
+def cast_enum_array(cast: typing.Callable[[T_co], T], raw_values: typing.Iterable[T_co], /) -> typing.List[T]:
```
davfsa, line 399: Personally, i find `cast_enum_array` a confusing name. Would `cast_variants_array` be better?

hikari/impl/rest.py
FasterSpeeding: taking `*args, **kwargs` for everything which uses it seems a bit inefficient when it isn't meant to meet every use case and practically never could (this is just 1 very specific use case); it doesn't cover the (more common) case where arrays of raw variants are deserialized to `dict[id, Unique]` partially cause out of 4 cases only 3 of them would be covered by a helper function as the 4th uses a type assertion which'd be left out due to avoiding unnecessary complexity (3 uses being less than half of the amount of times `cast_variants_array` is used)

davfsa (approved):

PR #1341, opened 2022-11-02
Replace empty spaces with underscores in filename

davfsa:
### Checklist


norinorin:
Discord also seems to strip out non-alphanumeric characters but full stop. Is this behaviour documented?

FasterSpeeding:
This feels like an inadequate fix also are you sure this isn't a bug or behaviour which should be fixed/handled on discord's end in a similar fashion to how https://github.com/discord/discord-api-docs/pull/4292#issuecomment-1003234845 was

davfsa:
Yeah, might be better. Will re-scope the pr

PR #1348, opened 2022-11-07
chore: update Discord API version in README

thenishantsapkota:
### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 <a href="https://hikari-py.dev/hikari"><img height="20" alt="Documentation status" src="https://img.shields.io/badge/documentation-up-00FF00.svg"></a>
 </p>
 
-An opinionated, static typed Discord microframework for Python3 and asyncio that supports Discord's v9 REST and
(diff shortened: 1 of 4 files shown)


davfsa:
It completely flew over my head to check for this. Iirc, there should be another v9 lying around (think it's in `shard.py`). Mind fixing that one too?

Ty for the pr tho!

novanai:
There's 2 instances of `V8` in `hikari.impl.shard`, 2 in `hikari.impl.rest` and 1 in `hikari.api.shard` too btw :P

thenishantsapkota:
I couldn't find any instances of `v9` anywhere.
Are you sure it exists?
Please confirm.

davfsa:
Think those are all. I believe they were just never updated from v8 to v9. Thanks for taking the time to do this!

davfsa (approved):

thenishantsapkota:
> Think those are all. I believe they were just never updated from v8 to v9. Thanks for taking the time to do this!

:bow:

PR #1357, opened 2022-11-13
Add Python 3.12 support

FasterSpeeding:
one day:tm:

(This will fail, just opening this as a place to track the compatibility issues)


davfsa:
Readme :)

davfsa:
Re-trigger CI

hikari/internal/fast_protocol.py
```diff
@@ -32,7 +32,7 @@
     from typing_extensions import Self
 
 _Protocol: FastProtocolChecking = NotImplemented
-_IGNORED_ATTRS = typing.EXCLUDED_ATTRIBUTES + ["__qualname__", "__slots__"]
+_IGNORED_ATTRS = frozenset(tuple(typing.EXCLUDED_ATTRIBUTES) + ("__qualname__", "__slots__"))
```
yakMM, line 35:
```suggestion
_IGNORED_ATTRS = frozenset(typing.EXCLUDED_ATTRIBUTES) | {"__qualname__", "__slots__"}
```

davfsa: That could work too, the main reason for the tuple was a little simplificacion, as I thought that `|` a frozenset would yield a normal set, but it's apparently a frozen one too

```py
>>> a = frozenset({1,2,3}) | {4,5,6}
>>> a
frozenset({1, 2, 3, 4, 5, 6})
```

TIL!

requirements.txt
```diff
@@ -1,4 +1,5 @@
-aiohttp~=3.8
+# Temporary!
+aiohttp==3.9.0rc0
```
davfsa, line 2:
```suggestion
aiohttp~=3.9
```

davfsa (approved):
And we have liftoff!

PR #1368, opened 2022-11-20
Allow specifying a filename to URL

davfsa:
### Summary
The use-case for this is when fetching images from an API that uses a URL similar to `https://my-api.com/api/modify-image/`, which would lead to the erroneous, `modify-image` as the filename. We can prevent that by allowing users to specify their own filename

### Checklist


hikari/files.py
```diff
@@ -733,17 +737,21 @@ class URL(WebResource):
...
+        self._filename = filename
 
     @property
     def url(self) -> str:
         return self._url
 
     @property
     def filename(self) -> str:
+        if self._filename:
+            return self._filename
```
FasterSpeeding, line 753: test case to make sure this doesn't ever get accidentally reverted?

davfsa: I might have forgotten to push it. Will do so later!

davfsa: Added!

FasterSpeeding (approved):

PR #1383, opened 2022-12-03
Support yielding in interaction handler

davfsa:
### Summary
Allow for proper defering in interaction handling in RESTBot

Example code:
```py
async def handle_command(interaction):
    yield interaction.build_deferred_response()

    await interaction.edit_initial_response("Hi!")
```

### Checklist


FasterSpeeding:
Who is this feature for, shouldn't ppl just be using a command/component client instead of doing it raw?

> Allow for proper deferring in interaction handling in RESTBot

Also this is kind of misleading, you could always defer and then give follow up responses just using futures

hikari/impl/interaction_server.py
```diff
@@ -441,7 +459,15 @@ async def on_interaction(self, body: bytes, signature: bytes, timestamp: bytes)
         if listener := self._listeners.get(type(interaction)):
             _LOGGER.debug("Dispatching interaction %s", interaction.id)
             try:
-                result = await listener(interaction)
+                call = listener(interaction)
+
+                if inspect.isasyncgen(call):
+                    result = await call.__anext__()
+                    asyncio.create_task(_consume_generator_handler(call))
```
FasterSpeeding, line 466: as always you'd need to save these tasks somewhere to make sure they aren't cancelled by GC

davfsa: Due to the way asyncio is built, the task will never get gc'ed, but should probably be keeping it somehow anyways. Forgot about that, thanks for pointing it out!

davfsa:
> Also this is kind of misleading, you could always defer and then give follow up responses just using tasks

This is exactly what this process is doing. It provides an easier way to create a task for followups by just using the same handler

hikari/impl/interaction_server.py
FasterSpeeding: > Due to the way asyncio is built, the task will never get gc'ed

That isn't guaranteed behaviour but I do remember having issues with production runs randomly cancelling long running tasks before.

The documented behaviour is that 
> https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

"Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks."

davfsa: Resolved :)

thesadru:
Wouldn't it potentially be better to provide this as a helper wrapper function? Could still be implicitly `subscribe(aiterator())` but explicitly as `subscribe(utils.somewrap(aiterator))`. Should declutter the already pretty big method.

davfsa:
> Wouldn't it potentially be better to provide this as a helper wrapper function? Could still be implicitly `subscribe(aiterator())` but explicitly as `subscribe(utils.somewrap(aiterator))`. Should declutter the already pretty big method.

I am not sure what you mean here. a) How would a helper function aid here? b) What would this declutter?

thesadru:
If you could wrap the generator to a standard async function when being registered it should be able to abstract stuff away from the `on_interaction` method.

davfsa:
> If you could wrap the generator to a standard async function when being registered it should be able to abstract stuff away from the `on_interaction` method.
> 

Considering we would have to create a task anyways and then store it in a list (because the function may be called again before the generator completed), we could end up with a bunch of objects that act as async functions that each store a list of tasks of running generators, it would be much more expensive than the 5-6 lines this adds to an existing function.

I would much rather keep it the way it is atm, if we were to go this way.

hikari/impl/interaction_server.py
```diff
@@ -365,6 +384,18 @@ async def close(self) -> None:
         await self._server.cleanup()
         self._server = None
         self._application_fetch_lock = None
+
+        for task in self._running_generator_listeners:
+            if not task.done() and not task.cancelled():
+                task.cancel()
```
FasterSpeeding, line 390: This method is documented as a graceful shutdown but this doesn't feel graceful, would it be worth it adding a force argument or changing the docs around this?

davfsa: That is indeed true, didnt even think about it. Will revisit

Thanks for pointing it out!

davfsa: Fixed to be in-line with the way aiohttp handles shutdown

FasterSpeeding (approved):

PR #1386, opened 2022-12-04
Remove incorrect `is_nsfw` from threads

### LINKED ISSUE #1328: GuildThreadChannel.is_nsfw should be removed
This is just never supplied, a thread's nsfw state is implicitly inherited from the parent channel and never supplied on the thread object itself

Might also be worth refining this further, are we sure we can't just do `is_nsfw=data["nsfw"]` or `is_nsfw=data.get("nsfw", False)` for the channel models where its applicable and just leave it out for everything else

davfsa:
### Summary
`is_nsfw` now also defaults to `False` when excluded from the payload. It seems like Discord will always send the nsfw field when it is set to `True`, but randomly exclude it when its `False`.

### Checklist

### Related issues
Closes #1328

### DIFF
--- a/changes/1386.bugfix.md
+++ b/changes/1386.bugfix.md
@@ -1,2 +1,2 @@
 Remove incorrect `is_nsfw` field from threads.
-- The "NSFW" status is inherited from the parent object and sent in threads.
+- The "NSFW" status is inherited from the parent object and not sent for threads.
--- a/changes/1386.bugfix.md
+++ b/changes/1386.bugfix.md
@@ -1,0 +1,2 @@
+Remove incorrect `is_nsfw` field from threads.
+- The "NSFW" status is inherited from the parent object and sent in threads.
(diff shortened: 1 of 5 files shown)


changes/1386.bugfix.md
```diff
@@ -0,0 +1,2 @@
+Remove incorrect `is_nsfw` field from threads.
+- The "NSFW" status is inherited from the parent object and sent in threads.
```
davfsa, line 2:
```suggestion
- The "NSFW" status is inherited from the parent object and not sent for threads.
```

changes/1386.bugfix.md
```diff
@@ -0,0 +1,2 @@
+Remove incorrect `is_nsfw` field from threads.
```
FasterSpeeding, line 1: You should document that the parent attribute was moved from GuildChannel to PermissibleGuildChannel as well since that's the change which effects more ppl here really

davfsa: Changed :)

FasterSpeeding (approved):

PR #1431, opened 2022-12-29
Add a warning when not running in optimized mode

davfsa:
### Checklist

### DIFF
--- a/hikari/internal/ux.py
+++ b/hikari/internal/ux.py
@@ -23,7 +23,14 @@
 """User-experience extensions and utilities."""
 from __future__ import annotations
 
-__all__: typing.Sequence[str] = ("init_logging", "print_banner", "supports_color", "HikariVersion", "check_for_updates")
+__all__: typing.Sequence[str] = (
+    "init_logging",
+    "print_banner",
+    "warn_if_not_optimized",
+    "supports_color",
+    "HikariVersion",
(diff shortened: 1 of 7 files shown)


hikari/internal/ux.py
```diff
@@ -233,6 +240,16 @@ def print_banner(
         stdout.write(string.Template(raw_banner).safe_substitute(args))
 
 
+def warn_if_not_optimized() -> None:
+    """Log a warning if not running in optimization mode."""
+    if __debug__:
+        _LOGGER.warning(
+            "You are running on no-optimization level, which may speed-down your application. "
```
GoogleGenius, line 247:
```suggestion
            "You are not running on any optimization level, which may slow down your application. "
```

davfsa: Would this be better?

```py
        "You are running on optimization level 0 (no optimizations), which may slow down your application. "
```

as it's not that you are not running on an optimization level, you are just running on optimization level 0.

GoogleGenius: I think that works!

davfsa: Cool, I'll change it when I have a computer to ensure it doesn't go over the character limit.

Thanks for the suggestion!

FasterSpeeding (approved):

PR #1442, opened 2023-01-02
Run CI tests on `bash` to exit fast on error on Windows

davfsa:
Looks like windows is the odd-one-out again. https://github.com/actions/runner-images/issues/6668.

Been sitting on this for a while now so they would hopefully fix it, but it looks like its intended behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Test Installation
+        shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


FasterSpeeding:
> Looks like windows is the odd-one-out again.

Well yeah it's documented behaviour that github actions default to different shells between windows and linux/macos when you don't explicitly specify it

davfsa:
Yeah, but the docs state that windows should default to `pwsh`, which should exit quick on error:
image

But that isnt the case: https://github.com/hikari-py/hikari/actions/runs/3818983261

PR #1460, opened 2023-01-09
Add speedups with orjson

Lunarmagpie:
### Checklist

This is a draft of adding speedups with orjson. This PR changes `data_binding.dump_json` to return a bytes object because I thought that was a good idea because it fits with orjson better. Casting the result of `orjson.dumps` to a string takes extremely long.

### DIFF
--- a/hikari/internal/data_binding.py
+++ b/hikari/internal/data_binding.py
@@ -109,7 +109,8 @@ else:
 
         JSONDecodeError = json.JSONDecodeError
 
-        def dump_json(obj: typing.Union[JSONArray, JSONObject], **kwargs: typing.Any) -> bytes:
+        def dump_json(obj: typing.Union[JSONArray, JSONObject], **kwargs: typing.Any) -> bytes:  # noqa: D103
+            # First definition already has docstring.
             return json.dumps(obj, **kwargs).encode("UTF-8")
 
         load_json = json.loads
--- a/hikari/internal/data_binding.py
+++ b/hikari/internal/data_binding.py
@@ -85,8 +85,8 @@ if typing.TYPE_CHECKING:
     JSONDecodeError: typing.Type[Exception] = Exception
     """Exception raised when loading an invalid JSON string."""
 
-    def dump_json(_: typing.Union[JSONArray, JSONObject], /, *, indent: int = ...) -> str:
-        """Convert a Python type to a JSON string."""
+    def dump_json(_: typing.Union[JSONArray, JSONObject], /, *, indent: int = ...) -> bytes:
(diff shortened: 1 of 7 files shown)


parafoxia:
I thought ujson had been decided on? Both mine and Dav's tests showed ujson to be faster, and in my tests, orjson was actually slower on average.

hikari/internal/data_binding.py
```diff
@@ -85,25 +85,35 @@
...
 
-    JSONDecodeError = json.JSONDecodeError
-    """Exception raised when loading an invalid JSON string."""
+        JSONDecodeError = json.JSONDecodeError
+
+        def dump_json(obj: typing.Union[JSONArray, JSONObject], **kwargs: typing.Any) -> bytes:  # noqa: D103
+            # First definition already has docstring.
+            return json.dumps(obj, **kwargs).encode("UTF-8")
+
+        load_json = json.loads
```
parafoxia, line 116: This might cause slowdowns for those without orjson installed. I think it should be a set requirement if we use it, as there's no OS-specific weirdness (that I'm aware of) like there is with hikari[speedups].

Lunarmagpie:
> I thought ujson had been decided on? Both mine and Dav's tests showed ujson to be faster, and in my tests, orjson was actually slower on average.

I tested speed in a way where variance from discord and the internet doesn't matter and found it to be significantly faster.

davfsa:
Even if orjson is significantly faster in an isolated benchmark perspective, it doesn't necessarily translate to a hikari application.

Orjson outputs in bytes, so if we later need it as a string, that might just render the optimizations useless and make it even slower than using the stdlib or ujson.

Furtheremore, this PR includes a lot of decoding and encoding that is not really needed and I don't agree with. If it adds so many changes, it's worth to stick with ujson

Lunarmagpie:
> Orjson outputs in bytes, so if we later need it as a string, that might just render the optimizations useless and make it even slower than using the stdlib or ujson.

I benchmarked it and orjson is still faster than stdlib and ujson.


> Furtheremore, this PR includes a lot of decoding and encoding that is not really needed and I don't agree with. If it adds so many changes, it's worth to stick with ujson

Fair enough. Thats more than a good enough reason to use ujson instead.

Lunarmagpie:
Im closing this because the encoding/decoding required to make orjson effective isn't that great.

parafoxia:
Yeah I don't think we have sufficient data right now to make an assertion. We should test more in different circumstances, or identify why exactly Hikari slows down orjson so much (maybe to do with constant encoding/decoding). Could that be something that changes?

PR #1470, opened 2023-01-13
Remove description_localisations from context menu

FasterSpeeding:
### Checklist


changes/1470.change.md
```diff
@@ -0,0 +1 @@
+Move `description_localizations` from `PartialCommand` to `SlashCommand` (removing it from `ContextMenuCommand`).
```
davfsa, line 1: This should maybe be called `1470.bugfix.md`

hikari/commands.py
```diff
@@ -249,12 +249,9 @@ class PartialCommand(snowflakes.Unique):
     version: snowflakes.Snowflake = attr.field(eq=False, hash=False, repr=True)
     """Auto-incrementing version identifier updated during substantial record changes."""
 
-    name_localizations: typing.Mapping[typing.Union[locales.Locale, str], str] = attr.field(factory=dict)
+    name_localizations: typing.Mapping[typing.Union[locales.Locale, str], str] = attr.field()
```
davfsa, line 252: A bit out of scope of this pr, but maybe worth fixing it too, but these should probably also get `attr.field(eq=False, hash=False, repr=False)`.

hikari/commands.py
```diff
@@ -481,6 +478,9 @@ class SlashCommand(PartialCommand):
         This will be inclusively between 1-100 characters in length.
     """
 
+    description_localizations: typing.Mapping[typing.Union[locales.Locale, str], str] = attr.field()
```
davfsa, line 481: Dito above

davfsa (approved):

PR #1471, opened 2023-01-13
Add support for audit log entry events

FasterSpeeding:
### Checklist


hikari/intents.py
```diff
@@ -219,13 +219,17 @@ class Intents(enums.Flag):
...
+    GUILD_MODERATION = 1 << 2
     """Subscribes to the events listed below.
 
+    * `GUILD_AUDIT_LOG_ENTRY_CREATE`
     * `GUILD_BAN_ADD`
     * `GUILD_BAN_REMOVE`
     """
 
+    GUILD_BANS = GUILD_MODERATION
+    """Deprecated alias of `GUILD_BANS`."""
```
davfsa, line 231: Can we do proper deprecation for this through a property?

hikari/impl/entity_factory.py
```diff
@@ -767,61 +767,81 @@ def _deserialize_member_move_entry_info(
...
+        entries = {}
+        for entry_payload in payload["audit_log_entries"]:
+            try:
+                entry = self.deserialize_audit_log_entry(entry_payload, guild_id=guild_id)
+
+            except errors.UnrecognisedEntityError as exc:
+                _LOGGER.debug(exc.reason)
+
+            else:
+                entries[entry.id] = entry
```
davfsa, line 844:
```suggestion
                entries[entry.id] = entry
 
            except errors.UnrecognisedEntityError as exc:
                _LOGGER.debug(exc.reason)
```

changes/1471.changes.md
```diff
@@ -0,0 +1 @@
+Renamed `Intents.GUILD_BANS` to `Intents.GUILD_MODERATION`.
```
davfsa, line 1: `.changes.md` doesnt exist. This would fall under `.deprecation.md`

davfsa (changes requested):

davfsa:
Thanks for working on this @FasterSpeeding!

hikari/impl/entity_factory.py
FasterSpeeding: I prefer limiting the scope of the try block to the bit I actually expect to raise when possible, tends to help with readability

davfsa: What I dont like here is the lil jump that it does. In this case, the code isnt that big, and it should be quite clear that adding it to a dict shoudnt be raising a `UnrecognisedEntityError`

FasterSpeeding: that jump is just the natural flow of a try, except, else statement though

hikari/intents.py
FasterSpeeding: Would require making a whole descriptor class for it since can't mix classmethod and property on 3.8

davfsa: #1473 should aid with this :)

davfsa (approved):
LGTM. I added an extra commit to use `enum.deprecated` in the deprecated member and merged the master branch :)

davfsa (approved):

davfsa (approved):

PR #1473, opened 2023-01-14
Add utilities to aid with enum member deprecation

davfsa:
### Checklist

### DIFF
--- a/hikari/internal/deprecation.py
+++ b/hikari/internal/deprecation.py
@@ -24,7 +24,7 @@
 
 from __future__ import annotations
 
-__all__: typing.Sequence[str] = ("warn_deprecated",)
+__all__: typing.Sequence[str] = ("warn_deprecated", "check_if_past_removal")
 
 import typing
 import warnings
@@ -33,12 +33,36 @@ from hikari import _about as hikari_about
 from hikari.internal import ux
 
 
-def warn_deprecated(name: str, /, *, removal_version: str, additional_info: str, stack_level: int = 3) -> None:
-    """Issue a deprecation warning.
+def check_if_past_removal(what: str, /, *, removal_version: str) -> None:
+    """Check if a deprecation is passed its removal version.
 
     Parameters
     ----------
-    name : str
+    what : str
+        What is being deprecated.
+
+    Other Parameters
+    ----------------
+    removal_version : str
+        The version it will be removed in.
+
+    Raises
+    ------
+    DeprecationWarning
+        If the deprecated item is past its removal version.
+    """
+    if ux.HikariVersion(hikari_about.__version__) >= ux.HikariVersion(removal_version):
+        raise DeprecationWarning(f"{what!r} is passed its removal version ({removal_version})")
+
+
+def warn_deprecated(what: str, /, *, removal_version: str, additional_info: str, stack_level: int = 3) -> None:
+    """Issue a deprecation warning.
+
+    If the item is past its deprecation version, an error will be raised instead.
+
+    Parameters
+    ----------
+    what : str
         What is being deprecated.
 
     Other Parameters
@@ -49,12 +73,16 @@ def warn_deprecated(name: str, /, *, removal_version: str, additional_info: str,
         Additional information on the deprecation for the user.
     stack_level : int
         The stack level to issue the warning in.
+
+    Raises
+    ------
+    DeprecationWarning
+        If the deprecated item is past its removal version.
     """
-    if ux.HikariVersion(hikari_about.__version__) >= ux.HikariVersion(removal_version):
-        raise DeprecationWarning(f"{name!r} is passed its removal version ({removal_version})")
+    check_if_past_removal(what, removal_version=removal_version)
 
     warnings.warn(
-        f"{name!r} is deprecated and will be removed in `{removal_version}`. {additional_info}",
+        f"{what!r} is deprecated and will be removed in `{removal_version}`. {additional_info}",
         category=DeprecationWarning,
         stacklevel=stack_level,
     )
--- a/hikari/internal/enums.py
+++ b/hikari/internal/enums.py
@@ -23,7 +23,7 @@
 """Implementation of parts of Python's `enum` protocol to be more performant."""
 from __future__ import annotations
 
-__all__: typing.Sequence[str] = ("Enum", "Flag")
+__all__: typing.Sequence[str] = ("deprecated", "Enum", "Flag")
 
 import functools
 import operator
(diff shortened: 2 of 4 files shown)


hikari/internal/enums.py
```diff
@@ -35,14 +35,45 @@
...
+            additional_info=f"Use '{self._alias}' instead.",
+        )
+
+        return self._alias
+
+
+class deprecated:
+    """Used to denote that an enum member is a deprecated alias of another."""
+
+    __slots__ = ("alias", "removal_version")
```
FasterSpeeding, line 62: the slots are wrong here

hikari/internal/enums.pyi
```diff
@@ -33,11 +33,19 @@
...
 from typing import Iterator as __Iterator
 from typing import Sequence as __Sequence
 from typing import TypeVar as __TypeVar
 from typing import Union as __Union
 
+class deprecated:
+    value: __Any
+    removal_version: str
+
+    def __init__(self, value: __Any, /, *, removal_version: str): ...
```
FasterSpeeding, line 46: This doesn't seem to work as expected if you try to access the raw value using `ENUM_MEMBER.value`. At runtime this'll error since the enum member doesn't have a `value` attribute and at type checking the `value` attribute will be typed as `deprecated` (instead of the raw type).

Did you consider typing this like the following:

```py
def deprecated(value: _T, /, *, removal_version: str) -> _T:  ...
```

Where it'd be relying on the fact that type-checking logic would take the returned `_T` and wrap it in a specialisation of the enum's type for the class attribute while using _T for `EnumMember.value`.

hikari/internal/enums.py
```diff
@@ -35,14 +35,45 @@
...
+        # Import kept in-line because of circular import issues
+        from hikari.internal import deprecation
+
+        deprecation.warn_deprecated(
+            self._name,
+            removal_version=self._removal_version,
+            additional_info=f"Use '{self._alias}' instead.",
+        )
+
+        return self._alias
```
FasterSpeeding, line 56: this is returning the enum member's name rather than value which seems like not what you wanted

hikari/internal/enums.py
```diff
@@ -35,14 +35,45 @@
...
+
+
+class deprecated:
+    """Used to denote that an enum member is a deprecated alias of another."""
+
+    __slots__ = ("alias", "removal_version")
+
+    def __init__(self, value: typing.Any, /, *, removal_version: str) -> None:
```
FasterSpeeding, line 64: this naming vs how it actually works seems a bit off here: you called it "value" but then pass the enum member's name instead.

hikari/internal/enums.py
```diff
@@ -58,6 +89,11 @@ def __setitem__(self, name: str, value: typing.Any) -> None:
         if name == "" or name == "mro":
             raise TypeError(f"Invalid enum member name: {name!r}")
 
+        if isinstance(value, deprecated):
+            value = _DeprecatedAlias(name, self.values_to_names[value.value], value.removal_version)
+            super().__setitem__(name, value)
```
FasterSpeeding, line 94: This behaviour seems to need testing (both for the type checking and runtime behaviour tbh)

hikari/internal/enums.pyi
```diff
@@ -33,11 +33,19 @@
...
 
 import enum as __enum
+from typing import Any as __Any
 from typing import Iterator as __Iterator
 from typing import Sequence as __Sequence
 from typing import TypeVar as __TypeVar
 from typing import Union as __Union
 
+class deprecated:
+    value: __Any
```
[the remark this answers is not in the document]
davfsa: As much as I would love to type it as a function (i did try), mypy will treat it as the correct type, and this will cause issues in `enums.py` when we try to access values on it or try to `isinstance` it.

hikari/internal/enums.py
davfsa: How this should be used is the following:

```py
class SomeEnum(int, enums.Enum):
    VAL = 1
    DEPRECATED = enum.deprecated(VAL, removal_version="...")
```

Since enum members are not "built" until the end of the class being defined, it is the same as doing

```py
class SomeEnum(int, enums.Enum):
    VAL = 1
    DEPRECATED = enum.deprecated(1, removal_version="...")
``` 

So, in this case the name is correct. In the case this still causes problems, would another name be more helpful? If so, which one? (I have no suggestions)

hikari/internal/deprecation.py
```diff
@@ -33,6 +33,23 @@
 from hikari.internal import ux
 
 
+def check_if_past_removal(name: str, /, *, removal_version: str) -> None:
+    """Check if a name is past its removal version.
```
davfsa, line 37: Document errors being raised here and in `warn_deprecated`

hikari/internal/enums.pyi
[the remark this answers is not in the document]
FasterSpeeding: why does that matter when `deprecated` isn't meant to be seen nor handled outside of enum.py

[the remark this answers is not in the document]
davfsa: That is true. Forgot to comment, but I tested using `reveal_type` and the returned type seems to be correct even without these changes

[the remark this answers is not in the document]
FasterSpeeding: It isn't

```py
from hikari.internal import enums
import warnings

warnings.simplefilter("ignore", category=DeprecationWarning)

class FooEnum(int, enums.Enum):
    FOO = 1

    BAR = enums.deprecated(FOO, removal_version="3.1.0")


print("type:", FooEnum.FOO, type(FooEnum.FOO))
print("member:", FooEnum.FOO.value, type(FooEnum.FOO.value))
print("member:", FooEnum.BAR, type(FooEnum.BAR))
print("value:", FooEnum.BAR.value, type(FooEnum.BAR.value))

import typing
if typing.TYPE_CHECKING:
    reveal_type(FooEnum.FOO.value)
    reveal_type(FooEnum.BAR.value)

```
```py
hikari\internal\time.py:93:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
test.py:19:17: note: Revealed type is "Literal[1]?"
test.py:20:17: note: Revealed type is "hikari.internal.enums.deprecated"
Found 1 error in 1 file (checked 1 source file)
```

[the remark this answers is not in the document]
davfsa: Oh, thats interesting. What python version are you using? Because that same code works for me:

```
nox > Running session mypy
nox > Re-using existing virtual environment at .nox/mypy.
nox > python -m pip install -r requirements.txt -r speedup-requirements.txt -r server-requirements.txt -r dev-requirements/mypy.txt -r dev-requirements/formatting.txt
nox > stubgen hikari/__init__.py hikari/api/__init__.py hikari/events/__init__.py hikari/impl/__init__.py hikari/interactions/__init__.py -o . --include-private --no-import
Processed 5 modules
Generated files under hikari/
nox > mypy -p hikari --config pyproject.toml
hikari/intents.py:420:13: note: Revealed type is "Literal[hikari.intents.Intents.DEPRECATED]?"
Success: no issues found in 99 source files
nox > mypy -p examples --config pyproject.toml
hikari/__init__.pyi:76: note: In module imported here,
examples/hello_world.py:14: note: ... from here:
hikari/intents.py:420:13: note: Revealed type is "Literal[hikari.intents.Intents.DEPRECATED]?"
Success: no issues found in 6 source files
nox > Session mypy was successful.
```

*Had to just add a random deprecated enum member to intents because mypy refused to pick up the stubs for hikari.internal after installing it*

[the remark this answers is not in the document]
davfsa: Fixed :)

FasterSpeeding (approved):

PR #1480, opened 2023-01-20
Fix `ChannelFollowWebhook` not including source in all cases

davfsa:
### Checklist

### DIFF
--- a/hikari/webhooks.py
+++ b/hikari/webhooks.py
@@ -855,11 +855,19 @@ class ChannelFollowerWebhook(PartialWebhook):
         This will be `None` when received within an audit log.
     """
 
-    source_channel: channels_.PartialChannel = attr.field(eq=False, hash=False, repr=True)
-    """The partial object of the channel this webhook is following."""
+    source_channel: typing.Optional[channels_.PartialChannel] = attr.field(eq=False, hash=False, repr=True)
+    """The partial object of the channel this webhook is following.
 
-    source_guild: guilds_.PartialGuild = attr.field(eq=False, hash=False, repr=True)
-    """The partial object of the guild this webhook is following."""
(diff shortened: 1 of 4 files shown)


hikari/webhooks.py
```diff
@@ -855,11 +855,21 @@ class ChannelFollowerWebhook(PartialWebhook):
         This will be `None` when received within an audit log.
     """
 
-    source_channel: channels_.PartialChannel = attr.field(eq=False, hash=False, repr=True)
-    """The partial object of the channel this webhook is following."""
+    source_channel: undefined.UndefinedOr[channels_.PartialChannel] = attr.field(eq=False, hash=False, repr=True)
```
FasterSpeeding, line 858: why not just Optional

davfsa: Just marks a difference between the channel actually existing and Discord just not wanting to send it VS Discord not sending it because its deleted (if that could be a possibility in the future)

FasterSpeeding: this is specifically the *channel follow* webhook type though

davfsa: I might be missing something, but not really sure what you mean? I know that its dumb that Discord dont send it, but thats how things are:

image

https://discord.com/developers/docs/resources/webhook#webhook-object-webhook-structure

FasterSpeeding: as in this webhook type has to be linked to a guild channel so that future case you specified shouldn't happen but also in the case that it does discord wouldn't distinguish between it and that other case in the model so that reason for using Undefined over None there just doesn't really work

davfsa: Fixed

FasterSpeeding (approved):

PR #1486, opened 2023-01-25
Add `orjson` as an optional speedup and allow to pass custom `json.dumps` and `json.loads` functions to all components.

### LINKED ISSUE #1453: Pottential speedup: ujson or orjson
A potential speedup we could add to the mix is [ujson](https://pypi.org/project/ujson/), which after some quick testing, it seems to be a drop in replacement to the standard libraries implementation, but performs a lot better, which is already noticeable on startup, so could be even more significant for large bots, so might be worth looking into:

*stdlib:*
image

*ujson:*
image

*Typehints are provided through [typeshed](https://github.com/python/typeshed/blob/main/stubs/ujson/ujson.pyi), so shouldnt have an issue with typecheckers*

----

If you want to provide feedback on this issue and test it yourself, you can easily do it by adding the following lines to your bot script (after installing ujson):

```py
import ujson
from hikari.internal import data_binding

data_binding.dump_json = ujson.dumps
data_binding.load_json = ujson.loads
```

davfsa:
### Checklist

### Related issues
Closes #1453


hikari/impl/rest.py
```diff
@@ -749,6 +771,13 @@ async def _perform_request(
         if auth:
             headers[_AUTHORIZATION_HEADER] = auth
 
+        data: typing.Union[None, aiohttp.BytesPayload, aiohttp.FormData] = None
+        if json:
+            if form_builder:
+                raise ValueError("Can only provide one of 'json' or 'form_builder', not both")
```
FasterSpeeding, line 777: If you've got cases like this which aren't meant to happen and thus aren't covered by tests maybe consider putting a `# pragma: no cover` comment on it (assuming that works with hikari's tooling)

hikari/internal/data_binding.py
```diff
@@ -75,62 +76,131 @@
...
+    data = json_dumps(obj)
+
+    if encode:
+        return data if isinstance(data, bytes) else data.encode(_UTF_8)
+
+    return data if isinstance(data, str) else data.decode(_UTF_8)
+
+
+@typing.final
+class JSONPayload(aiohttp.BytesPayload):
```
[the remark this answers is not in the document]
davfsa: You hit exactly what i wanted to do, make rest a bit shorter. So for that sake, will leave it here :)

hikari/impl/rest.py
davfsa: I just forgot to add a test case for it 😅

Ty for pointing it out!

hikari/errors.py
```diff
@@ -317,7 +318,8 @@ def __str__(self) -> str:
             try:
                 value += _dump_errors(self.errors).strip("\n")
             except KeyError:
-                value += data_binding.dump_json(self.errors, indent=2)
+                # Use the stdlib json.dumps here to be able to indent
+                value += json.dumps(self.errors, indent=2)
```
Lunarmagpie, line 322: Could a function like `data_binding.json_pretty_print` work here?

davfsa: Since we are only doing this in this one specific place rn, its fine for this to be a one off

hikari/impl/rest.py
davfsa: Fixed!

hikari/internal/data_binding.py
```diff
@@ -75,62 +76,131 @@
 Stringish = typing.Union[str, int, bool, undefined.UndefinedType, None, snowflakes.Unique]
 """Type hint for any valid that can be put in a StringMapBuilder"""
 
+JSONEncoder = typing.Callable[[typing.Union[JSONArray, JSONObject]], typing.Union[str, bytes]]
+"""Type hint for hikari-valid JSON encoders.
+
+A hikari-valid JSON encoder is one which will take in a JSON-ish object and output either `str`
```
Lunarmagpie, line 82: I think the phrasing `hikari-comptible` would be less awkward here. This isn't an approval-blocking issue to me tho.

Lunarmagpie (commented):
This LGTM. I checked the performance of using `isinstance` to check if decoding/encoding is required and the performance loss is negligible compared to parsing a large data structure.

hikari/internal/data_binding.py
davfsa: Changed!

tests/hikari/impl/test_rest.py
```diff
@@ -1715,6 +1719,27 @@ async def test___aenter__and__aexit__(self, rest_client):
 
         rest_client.close.assert_awaited_once_with()
 
+    @hikari_test_helpers.timeout()
+    async def test_perform_request_errors_if_both_json_and_form_builder_passed(self, rest_client):
+        route = routes.Route("GET", "/something/{channel}/somewhere").compile(channel=123)
+
+        with pytest.raises(ValueError):
```
davfsa, line 1726:
```suggestion
        with pytest.raises(ValueError, match="Can only provide one of 'json' or 'form_builder', not both"):
```

tests/hikari/impl/test_rest.py
```diff
@@ -1715,6 +1719,27 @@ async def test___aenter__and__aexit__(self, rest_client):
 
         rest_client.close.assert_awaited_once_with()
 
+    @hikari_test_helpers.timeout()
+    async def test_perform_request_errors_if_both_json_and_form_builder_passed(self, rest_client):
+        route = routes.Route("GET", "/something/{channel}/somewhere").compile(channel=123)
+
+        with pytest.raises(ValueError, match="Can only provide one of 'json' or 'form_builder', not both"):
```
davfsa, line 1726:
```suggestion
        with pytest.raises(ValueError, match="Can only provide one of 'json' or 'form_builder', not both"):
```

hikari/internal/data_binding.py
```diff
@@ -75,62 +76,131 @@
...
+    json_dumps: JSONEncoder = default_json_dumps,
+) -> bytes:
+    ...
 
-    load_json = json.loads
-    """Convert a JSON string to a Python type."""
 
-    JSONDecodeError = json.JSONDecodeError
-    """Exception raised when loading an invalid JSON string."""
+def dump_json(
```
FasterSpeeding, line 144: this whole function just feels like a unnecessary extra steps ngl (esp considering the whole point of this is to plop down into pretty important hot paths). Are there even any cases where encode isn't True cause if not then this is just unnecessary complexity?

davfsa: I dont like it either, but in the hopes of providing a more agnostic interface, we need to ensure that the deserializers that we get passed behave as we expect them to (thank you orjson for being the special one!)

FasterSpeeding: It's not that hard for ppl to do `dumps=lambda d: dumps(d).encode()`

davfsa: Sorted!

hikari/internal/data_binding.py
```diff
@@ -75,62 +75,87 @@
...
+    default_json_dumps = orjson.dumps
+    default_json_loads = orjson.loads
+except ModuleNotFoundError:
     import json
 
-    dump_json = json.dumps
-    """Convert a Python type to a JSON string."""
+    def default_json_dumps(obj: typing.Union[JSONArray, JSONObject]) -> bytes:
+        """Encode a JSON object to a `str`."""
+        return json.dumps(obj, separators=(",", ":")).encode()
```
davfsa, line 116:
```suggestion
        return json.dumps(obj, separators=(",", ":")).encode(_UTF_8)
```

davfsa: Also, consider extracting the separators out of there to not keep building the tuple constantly

hikari/internal/data_binding.py
```diff
@@ -75,62 +75,87 @@
...
+    default_json_dumps = orjson.dumps
+    default_json_loads = orjson.loads
+except ModuleNotFoundError:
     import json
 
-    dump_json = json.dumps
-    """Convert a Python type to a JSON string."""
+    def default_json_dumps(obj: typing.Union[JSONArray, JSONObject]) -> bytes:
+        """Encode a JSON object to a `str`."""
+        return json.dumps(obj, separators=(",", ":")).encode(_UTF_8)
```
davfsa, line 116:
```suggestion
    _json_separators = (",", ":")

    def default_json_dumps(obj: typing.Union[JSONArray, JSONObject]) -> bytes:
        """Encode a JSON object to a `str`."""
        return json.dumps(obj, separators=_json_separators).encode(_UTF_8)
```

FasterSpeeding (approved):

PR #1491, opened 2023-02-05
Pass strings instead of bytes to add_field for payload_json

FasterSpeeding:
### Summary
Passing bytes here leads to add_field treating the field as files and setting the filename to payload_json Discord will then treat this as an attachment/file rather than as the actual json payload.

### Checklist

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -1505,7 +1505,7 @@ class RESTClientImpl(rest_api.RESTClient):
             body.put("message_reference", message_reference)
 
         if form_builder is not None:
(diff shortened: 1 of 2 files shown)


FasterSpeeding:
This could possibly also be fixed by bringing back that other approach which used instance checking instead to avoid re-decoding bytes which were already encoded from a string but imo this isn't a hot path so it shouldn't matter enough performance wise either way.

FasterSpeeding:
> This could possibly also be fixed by bringing back that other approach which used instance checking instead to avoid re-decoding bytes which were already encoded from a string but imo this isn't a hot path so it shouldn't matter enough performance wise either way.

PR #1492, opened 2023-02-05
Unify client and server multipart body building

FasterSpeeding:
### Summary
Just an alternative proposal to #1491

### Checklist


hikari/internal/data_binding.py
```diff
@@ -130,36 +130,58 @@ def __init__(
...
-            form.add_field(field[0], field[1], content_type=field[2])
+            body_payload = aiohttp.BytesPayload(field[1], content_type=field[2])
+            body_payload.set_content_disposition("form-data", name=field[0])
+            form.append_payload(body_payload)
 
         for name, resource in self._resources:
-            stream = await stack.enter_async_context(resource.stream(executor=executor))
-            mimetype = stream.mimetype or _APPLICATION_OCTET_STREAM
-            form.add_field(name, stream, filename=stream.filename, content_type=mimetype)
+            async with resource.stream(executor=executor, head_only=True) as stream:
```
FasterSpeeding, line 179: maybe this could be optimised a tiny bit by adding a get_mimetype method to Resource which is specialised for impls where we already known the mimetype before reading to not do a head_only stream or something

FasterSpeeding:
It's prob about time edit semantics are brought to interaction server's file response logic (if possible) and this should make it easier to port the change (I could even do it in this PR)

hikari/internal/messages.py
```diff
@@ -0,0 +1,161 @@
...
+    user_mentions: undefined.UndefinedOr[
+        typing.Union[snowflakes.SnowflakeishSequence[users.PartialUser], bool]
+    ] = undefined.UNDEFINED,
+    role_mentions: undefined.UndefinedOr[
+        typing.Union[snowflakes.SnowflakeishSequence[guilds.PartialRole], bool]
+    ] = undefined.UNDEFINED,
+    edit: bool = False,
+) -> typing.Tuple[data_binding.JSONObjectBuilder, typing.Mapping[int, files.Resource[typing.Any]]]:
+    """Create a message create/edit payload."""
+    to_upload: typing.Dict[int, files.Resource[typing.Any]] = {}
```
FasterSpeeding, line 89: Could've done `typing.List[typing.Tuple[int, files.Resource[typing.Any]]` but stuck with Dict/Mapping for now since it keeps the type cleaner and this does reach the public interface a bit

hikari/impl/rest.py
```diff
@@ -1356,104 +1399,29 @@ def _build_message_payload(  # noqa: C901- Function too complex
...
-                attachments_payload.append({"id": attachment_id, "filename": resource.filename})
-                form_builder.add_resource(f"files[{attachment_id}]", resource)
-                attachment_id += 1
-
-            body.put("attachments", attachments_payload)
-
-        elif attachment is None or attachments is None:
-            body.put("attachments", None)
+        attachments, content = _to_list(attachment, attachments, content, _ATTACHMENT_TYPES, "attachment")
+        components, content = _to_list(component, components, content, special_endpoints.ComponentBuilder, "attachment")
```
FasterSpeeding, line 1403: This now allows passing a component builder for `content`

hikari/impl/special_endpoints.py
```diff
@@ -1150,48 +1150,21 @@ def set_user_mentions(
...
-
-        if (
-            not undefined.all_undefined(self.mentions_everyone, self.user_mentions, self.role_mentions)
-            or self.type is base_interactions.ResponseType.MESSAGE_CREATE
-        ):
-            data["allowed_mentions"] = mentions.generate_allowed_mentions(
-                self.mentions_everyone, undefined.UNDEFINED, self.user_mentions, self.role_mentions
-            )
+    ) -> typing.Tuple[typing.MutableMapping[str, typing.Any], typing.Mapping[int, files.Resource[files.AsyncReader]]]:
+        data, to_upload = message_utils.build_message_payload(
```
FasterSpeeding, line 1154: Adding support for the edit semantics around passing `None` for fields and allowing using `messages.Attachment` to avoid re-uploading attachments should be as easy as deciding whether `edit` should be True or False based on `self._type` and allowing None to be passed for some of these fields (I just haven't done that yet since that wasn't the focus of this PR)

hikari/api/interaction_server.py
```diff
@@ -87,7 +87,7 @@ def charset(self) -> typing.Optional[str]:
         raise NotImplementedError
 
     @property
-    def files(self) -> typing.Sequence[files_.Resource[files_.AsyncReader]]:
+    def files(self) -> typing.Mapping[int, files_.Resource[files_.AsyncReader]]:
```
FasterSpeeding, line 90: This and `InteractionResponseBuilder.build`'s return types are the only breaking changes in this PR

hikari/internal/data_binding.py
```diff
@@ -129,36 +130,58 @@ def __init__(self, value: typing.Any, dumps: JSONEncoder = default_json_dumps) -
...
 
     def __init__(self) -> None:
-        self._fields: typing.List[typing.Tuple[str, typing.Union[str, bytes], typing.Optional[str]]] = []
+        self._fields: typing.List[typing.Tuple[str, bytes, typing.Optional[str]]] = []
         self._resources: typing.List[typing.Tuple[str, files.Resource[files.AsyncReader]]] = []
 
-    def add_field(
-        self, name: str, data: typing.Union[str, bytes], *, content_type: typing.Optional[str] = None
-    ) -> None:
+    def add_field(self, name: str, data: bytes, *, content_type: typing.Optional[str] = None) -> None:
```
FasterSpeeding, line 164: Do other fields (e.g. String only) Need to have a content type set?

hikari/internal/messages.py
```diff
@@ -0,0 +1,161 @@
...
+            # Duplicates will cause Discord to error.
+            ids = {str(int(r)) for r in role_mentions}
+            allowed_mentions["roles"] = list(ids)
+
+        body.put("allowed_mentions", allowed_mentions)
+
+    if attachment_payloads:
+        body.put("attachments", attachment_payloads)
+
+    elif attachments is not undefined.UNDEFINED:
```
FasterSpeeding, line 146: Check that this behaves the same

hikari/internal/messages.py
```diff
@@ -0,0 +1,161 @@
...
+        body.put("attachments", None)
+
+    return body, to_upload
+
+
+def build_form_builder(
```
FasterSpeeding, line 152: Change name

davfsa:
Dont think this is needed anymore, the approach we went for in #1493 is probably enough.

Can be rebased and reopened if you see necessary :)

PR #1499, opened 2023-02-07
Use typing.Self for Self types

FasterSpeeding:
### Checklist


hikari/api/event_manager.py
```diff
@@ -35,6 +35,8 @@
 if typing.TYPE_CHECKING:
     import types
 
+    from typing_extensions import Self
```
FasterSpeeding, line 38: could type the returns as `typing_extensions.Self` everywhere but that's a bit too verbose for what it is and looks bad

davfsa: I think it's more readable this way. It also shouldn't be confusing to read anyways.

hikari/impl/entity_factory.py
```diff
@@ -2403,7 +2403,7 @@ def _deserialize_autocomplete_interaction_option(
             name=payload["name"],
             type=option_type,
             value=value,
-            options=suboptions,
+            options=suboptions,  # type: ignore[arg-type]  # MyPy gets the type equality wrong here.
```
FasterSpeeding, line 2406: This is specifically a bug with mypy's attrs handling

hikari/impl/entity_factory.py
```diff
@@ -2381,14 +2381,14 @@ def _deserialize_interaction_command_option(
...
```
davfsa, line 2384: How come this happens now?

FasterSpeeding:
`# Any has to be used here as we can't access Self type in this context`

hikari/interactions/command_interactions.py
```diff
@@ -101,8 +103,7 @@ class CommandInteractionOption:
     subcommand or group.
     """
 
-    # TODO: use typing.Self here
-    options: typing.Optional[typing.Sequence[CommandInteractionOption]] = attr.field(repr=True)
+    options: typing.Optional[typing.Sequence[Self]] = attr.field(repr=True)
```
davfsa, line 106: This wouldnt be correct tho. We have options inside other options, not an option inside itself

FasterSpeeding: Nested options are always the same type as the parent option

davfsa: Yeah, but they are not the same one, typing.Self implies it will be the same item, which it wont be in this case

FasterSpeeding: It doesn't imply the exact same instance (python's type checking semantics doesn't really have that capability for being aware of thay outside of some special casing for singletons and same variable checks), it just implies the same type.
It's just syntax sugar for a type variable that's bound by self's type. No different from doing
```py
_Foot = TypeVar("_FooT", bound="Foo")

class Foo:
    @property
    def sub_foo(self: _FooT) -> Sequence[_FooT]:
        raise NotImplementedError
```

davfsa: For the sake of not adding more type ignores, i would rather just leave this one as it is currently, it wont change much and remove the ignores

FasterSpeeding: It's literally a type fix though, it lets you options generically without having to cast on each recuse and stops the need to retype this variable on subclasses of Option

Changes
```py
_OptionT = typing.TypeVar("_OptionT", bound=hikari.CommandInteractionOption)
SUB_COMMAND_OPTION_TYPES: typing.Final[frozenset[hikari.OptionType]] = frozenset(
    [hikari.OptionType.SUB_COMMAND, hikari.OptionType.SUB_COMMAND_GROUP]
)


def flatten_options(
    name: str, options: typing.Optional[collections.Sequence[_OptionT]], /
) -> tuple[str, collections.Sequence[_OptionT]]:
    while options and (first_option := options[0]).type in SUB_COMMAND_OPTION_TYPES:
        name = f"{name} {first_option.name}"
        options = typing.cast("collections.Sequence[_OptionT]", first_option.options)

    return name, options or ()
```
to
```py
_OptionT = typing.TypeVar("_OptionT", bound=hikari.CommandInteractionOption)
SUB_COMMAND_OPTION_TYPES: typing.Final[frozenset[hikari.OptionType]] = frozenset(
    [hikari.OptionType.SUB_COMMAND, hikari.OptionType.SUB_COMMAND_GROUP]
)


def flatten_options(
    name: str, options: typing.Optional[collections.Sequence[_OptionT]], /
) -> tuple[str, collections.Sequence[_OptionT]]:
    while options and (first_option := options[0]).type in SUB_COMMAND_OPTION_TYPES:
        name = f"{name} {first_option.name}"
        options = first_option.options

    return name, options or ()
```

FasterSpeeding: This is just something that mypy needs to fix cause like i said works fine if you're using @dataclasses.dataclass or just a manual init

davfsa: Hopefully this will be fixed by mypy soon and we can remove the type ignores

davfsa (approved):

PR #1506, opened 2023-02-10
Fix sticker image_url property

HyperGH:
### Summary
Speculative fix for `image_url` possibly not accounting for GIF stickers, however it could also be possible that Discord just converts these to `.apng`s, sadly I wasn't able to verify as the client keeps refusing my gif uploads as "Invalid payload".

### Checklist


hikari/stickers.py
```diff
@@ -156,9 +156,15 @@ def image_url(self) -> files.URL:
...
+        then the extension will be `.json`, if it's `StickerFormatType.GIF` it will be `.gif`. Otherwise, it will be `.png`.
         """
-        ext = "json" if self.format_type is StickerFormatType.LOTTIE else "png"
+        ext = (
+            "json"
+            if self.format_type is StickerFormatType.LOTTIE
+            else "gif"
+            if self.format_type is StickerFormatType.GIF
+            else "png"
+        )
```
davfsa, line 167: It will probably be a lot better to use a mapping and then do:

```
_EMOJI_EXTENSIONS.get(self.format_type, "png")
```

HyperGH: I was reluctant to create a mapping for just 3 formats, but I could make it happen!

HyperGH: Honestly I do not think this is that much better, but there you go :P

hikari/stickers.py
```diff
@@ -79,6 +79,14 @@ class StickerFormatType(int, enums.Enum):
     """A GIF sticker."""
 
 
+_STICKER_EXTENSIONS: typing.Dict[typing.Union[StickerFormatType, int], str] = {
+    StickerFormatType.LOTTIE: "json",
+    StickerFormatType.GIF: "gif",
+    StickerFormatType.PNG: "png",
+    StickerFormatType.APNG: "png",
```
davfsa, line 86: Since we add the default as being png, we don't need these here

davfsa (approved):

PR #1533, opened 2023-03-05
Flatten the component builders

FasterSpeeding:
### Checklist

### DIFF
--- a/hikari/api/special_endpoints.py
+++ b/hikari/api/special_endpoints.py
@@ -1611,9 +1611,9 @@ class TextSelectMenuBuilder(SelectMenuBuilder, abc.ABC, typing.Generic[_ParentT]
             The option's description.
 
             This can be up to 100 characters long.
-        emoji
+        emoji : typing.Union[hikari.snowflakes.Snowflakeish, hikari.emojis.Emoji, str, hikari.undefined.UndefinedType]
             The option's display emoji.
-        is_default
+        is_default : bool
             Whether this option should be selected by default.
 
         Returns
--- a/hikari/api/special_endpoints.py
+++ b/hikari/api/special_endpoints.py
@@ -1691,7 +1691,6 @@ class TextInputBuilder(ComponentBuilder, abc.ABC):
 
     @property
     @abc.abstractmethod
-    #  @deprecation.deprecated("Use .is_required")
     def required(self) -> bool:
         """Deprecated alias for `hikari.api.special_endpoints.TextInputBuilder`."""
 
@@ -2029,7 +2028,6 @@ class MessageActionRowBuilder(ComponentBuilder, abc.ABC):
         custom_id: str,
         /,
         *,
-        # TODO: accept options here as well?
         placeholder: undefined.UndefinedOr[str] = undefined.UNDEFINED,
         min_values: int = 0,
         max_values: int = 1,
--- a/hikari/api/special_endpoints.py
+++ b/hikari/api/special_endpoints.py
@@ -1878,7 +1878,7 @@ class MessageActionRowBuilder(ComponentBuilder, abc.ABC):
     ) -> Self:
         """Add an interactive button component to this action row builder.
 
-        Either `emoji` xor `label` must be provided to be the button's
+        Either `emoji` or `label` (exclusively) must be provided to be the button's
         displayed label.
 
         Parameters
@@ -1913,7 +1913,7 @@ class MessageActionRowBuilder(ComponentBuilder, abc.ABC):
     ) -> Self:
         """Add a link button component to this action row builder.
 
-        Either `emoji` xor `label` must be provided to be the button's
+        Either `emoji` or `label` (exclusively) must be provided to be the button's
         displayed label.
 
         Parameters
--- a/hikari/api/special_endpoints.py
+++ b/hikari/api/special_endpoints.py
@@ -2029,6 +2029,7 @@ class MessageActionRowBuilder(ComponentBuilder, abc.ABC):
         custom_id: str,
         /,
         *,
+        # TODO: accept options here as well?
         placeholder: undefined.UndefinedOr[str] = undefined.UNDEFINED,
         min_values: int = 0,
         max_values: int = 1,
--- a/hikari/api/special_endpoints.py
+++ b/hikari/api/special_endpoints.py
@@ -1976,7 +1976,6 @@ class MessageActionRowBuilder(ComponentBuilder, abc.ABC):
     @abc.abstractmethod
     def add_channel_menu(
         self,
-        type_: typing.Union[components_.ComponentType, int],
         custom_id: str,
         /,
         *,
@@ -1990,8 +1989,6 @@ class MessageActionRowBuilder(ComponentBuilder, abc.ABC):
 
         Parameters
         ----------
-        type_ : typing.Union[hikari.components.ComponentType, int]
-            The type for the select menu.
         custom_id : str
             A developer-defined custom identifier used to identify which menu
             triggered component interactions.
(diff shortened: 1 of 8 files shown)


hikari/impl/special_endpoints.py
```diff
@@ -1542,10 +1532,9 @@ def custom_id(self) -> str:
 
 @attr_extensions.with_copy
 @attr.define(kw_only=True, weakref_slot=False)
-class _SelectOptionBuilder(special_endpoints.SelectOptionBuilder["_TextSelectMenuBuilderT"]):
+class _SelectOptionBuilder(special_endpoints.SelectOptionBuilder):
```
FasterSpeeding, line 1535: Make this impl public so it can be more easily passed to inits

hikari/api/special_endpoints.py
```diff
@@ -1597,50 +1572,58 @@ def set_max_values(self, value: int, /) -> Self:
...
             The user-facing name of this option, max 100 characters.
         value : str
             The developer defined value of this option, max 100 characters.
+        description : hikari.undefined.UndefinedOr[str]
+            The option's description.
+
+            This can be up to 100 characters long.
+        emoji
+            The option's display emoji.
+        is_default
```
davfsa, line 1616: Missing typing

hikari/api/special_endpoints.py
```diff
@@ -1887,95 +1865,199 @@ def add_component(self, component: ComponentBuilder, /) -> Self:
...
+    ) -> Self:
+        """Add an interactive button component to this action row builder.
 
-    @typing.overload
-    @abc.abstractmethod
-    def add_button(
-        self, style: typing.Literal[components_.ButtonStyle.LINK, 5], url: str, /
-    ) -> LinkButtonBuilder[Self]:
-        ...
+        Either `emoji` xor `label` must be provided to be the button's
```
davfsa, line 1881:
```suggestion
        Either `emoji` or `label` must be provided to be the button's
```

hikari/impl/special_endpoints.py
```diff
@@ -1527,33 +1525,36 @@ def url(self) -> str:
...
     )
-    _emoji_id: undefined.UndefinedOr[str] = attr.field(alias="emoji_id", default=undefined.UNDEFINED)
-    _emoji_name: undefined.UndefinedOr[str] = attr.field(alias="emoji_name", default=undefined.UNDEFINED)
-    _is_default: bool = attr.field(alias="is_default", default=False)
+    _emoji_id: undefined.UndefinedOr[str] = attr.field(init=False, default=undefined.UNDEFINED)
+    _emoji_name: undefined.UndefinedOr[str] = attr.field(init=False, default=undefined.UNDEFINED)
+    _is_default: bool = attr.field(alias="is_default", default=False, kw_only=True)
+
+    def __attrs_post_init__(self) -> None:
+        self._emoji_id, self._emoji_name = _build_emoji(self._emoji)
```
davfsa, line 1557: Instead of storing this in this way, cant we just keep `self._emoji` and then calculate what it should be in the `build`?

hikari/impl/special_endpoints.py
```diff
@@ -1682,29 +1674,96 @@ def build(self) -> typing.MutableMapping[str, typing.Any]:
...
+            max_values=max_values,
+            is_disabled=is_disabled,
+        )
+        self._options = list(options)
+        self._parent = parent
+
+    @property
+    def parent(self) -> _ParentT:
+        if self._parent is None:
+            raise RuntimeError("This menu has no parent")
```
davfsa, line 1742: Should we allow text select menu builders without a parent?

hikari/impl/special_endpoints.py
```diff
@@ -1785,15 +1841,22 @@ def value(self) -> undefined.UndefinedOr[str]:
         return self._value
 
     @property
-    def required(self) -> undefined.UndefinedOr[bool]:
+    def required(self) -> bool:
```
davfsa, line 1844:
```suggestion
    @deprecation.deprecated
    def required(self) -> bool:
```

davfsa (changes requested):
+1 the changes

hikari/api/special_endpoints.py
FasterSpeeding: that's an exclusive or

hikari/impl/special_endpoints.py
FasterSpeeding: How else would you pass them to `TextSelectMenuBuilder.__init__`

hikari/impl/special_endpoints.py
FasterSpeeding: These models are designed to be reused, it just does the check once rather than doing it every time its built

hikari/impl/special_endpoints.py
```diff
@@ -1682,29 +1674,96 @@ def build(self) -> typing.MutableMapping[str, typing.Any]:
...
-    # Any has to be used here as we can't access Self type in this context
-    _options: typing.List[special_endpoints.SelectOptionBuilder[typing.Any]] = attr.field(alias="options", factory=list)
+    _options: typing.List[special_endpoints.SelectOptionBuilder] = attr.field()
+    _parent: typing.Optional[_ParentT] = attr.field()
+    _type: typing.Literal[component_models.ComponentType.TEXT_SELECT_MENU] = attr.field()
+
+    if not typing.TYPE_CHECKING:
+        # This will not work with the generated for attrs copy methods.
+        __copy__ = None
+        __deepcopy__ = None
```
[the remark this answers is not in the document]
FasterSpeeding: https://github.com/hikari-py/hikari/blob/master/hikari/internal/attr_extensions.py assumes an init which the metadata returned by attr.fields, which is not the case here

hikari/impl/special_endpoints.py
FasterSpeeding: deprecation.deprecated won't make it to the next release without including `typing_extensions` as a dependency since MyPy's next release is planned for the 4th of April https://github.com/python/mypy/issues/14916

hikari/api/special_endpoints.py
davfsa: Would an easier language be: `Exactly one of `emoji` or `label` must be provided to be the button's`?

hikari/impl/special_endpoints.py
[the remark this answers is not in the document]
davfsa: Yeah, but we dont seem to be using that here to generate copies (i might be missing something)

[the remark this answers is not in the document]
FasterSpeeding: `__copy__` and `__deepcopy__` are inherited like any other class attribute

hikari/impl/special_endpoints.py
davfsa: Ah, thats my bad, forgot we already had it in master

hikari/impl/special_endpoints.py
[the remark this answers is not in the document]
davfsa: Maybe a better question would be, why does that class need copy methods?

hikari/api/special_endpoints.py
FasterSpeeding: I just went with the common dev/logic term cause its concise. Otherwise I would prob phrase it like ``Either `emoji` or `label` must be provided to be the button's displayed label, but not both.`` but that's not as concise.

hikari/impl/special_endpoints.py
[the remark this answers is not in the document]
FasterSpeeding: rn `attr_extensions.with_copy` is just applied to public classes where attrs has been used to generate its init. same rule here

hikari/api/special_endpoints.py
davfsa: `xor` might be more confusing, so I would rather keep it more "english"

davfsa (approved):
Time to test it in the wild!

PR #1558, opened 2023-03-24
Add UserCredentialsStrategy

theonlydoublee:
### Summary
Add `UserCredentialsStrategy` to `hikari\impl\rest.py` similar to `ClientCredentialsStrategy` to allow simple OAuth2 token generation and refreshing.

I do not know how to make the tests for this except by copying from the ones for `ClientCredentialsStrategy` and I tried but I do not know how to use the mock stuff.  I am also not sure which tests I will need.  Don't be afraid to be harsh giving feedback on what to do or I did.

### Checklist

### Related issues
- None

### DIFF
--- a/hikari/impl/rest.py
+++ b/hikari/impl/rest.py
@@ -214,6 +214,132 @@ class ClientCredentialsStrategy(rest_api.TokenStrategy):
             self._token = None
 
 
+class UserCredentialsStrategy(rest_api.TokenStrategy):
+    """Strategy class for handling client credential OAuth2 authorization.
+
+    Parameters
+    ----------
+    client_id : typing.Optional[hikari.snowflakes.SnowflakeishOr[hikari.guilds.PartialApplication]]
+        Object or ID of the application this client credentials strategy should
+        authorize as.
+    client_secret : str
+        Client secret to use when authorizing.
+    auth_code : str
+        Auth code given from Discord when user authorizes
+    redirect_uri: str
+        The redirect uri that was included in the authorization request
+
+    Other Parameters
+    ----------------
+    scopes : typing.Sequence[str]
+        The scopes to authorize for.
+    """
+
+    __slots__: typing.Sequence[str] = (
+        "_client_id",
+        "_client_secret",
+        "_exception",
+        "_expire_at",
+        "_lock",
+        "_scopes",
+        "_token",
+        "_code",
+        "_redirect_uri",
+        "_refresh_token",
+    )
+
+    def __init__(
+        self,
+        client_id: snowflakes.SnowflakeishOr[guilds.PartialApplication],
+        client_secret: str,
+        auth_code: str,
+        redirect_uri: str,
+        *,
+        scopes: typing.Sequence[typing.Union[applications.OAuth2Scope, str]] = (applications.OAuth2Scope.IDENTIFY,),
+    ) -> None:
+        self._client_id = snowflakes.Snowflake(client_id)
+        self._client_secret = client_secret
+        self._exception: typing.Optional[errors.ClientHTTPResponseError] = None
+        self._expire_at = 0.0
+        self._lock = asyncio.Lock()
+        self._scopes = scopes
+        self._token: typing.Optional[str] = None
+        self._refresh_token = None
+        self._code = auth_code
+        self._redirect_uri = redirect_uri
+
+    @property
+    def client_id(self) -> snowflakes.Snowflake:
+        """ID of the application this token strategy authenticates with."""
+        return self._client_id
+
+    @property
+    def _is_expired(self) -> bool:
+        return time.monotonic() >= self._expire_at
+
+    @property
+    def scopes(self) -> typing.Sequence[typing.Union[applications.OAuth2Scope, str]]:
+        """Sequence of scopes this token strategy authenticates for."""
+        return self._scopes
+
+    @property
+    def token_type(self) -> applications.TokenType:
+        return applications.TokenType.BEARER
+
+    async def acquire(self, client: rest_api.RESTClient | RESTApp) -> str:
+        if not self._code:
+            raise "Token has been invalidated. Unable to get current or new token"
+
+        if self._token and not self._is_expired:
+            return self._token
+
+        async with self._lock:
+            if self._token and not self._is_expired:
+                return self._token
+
+            if self._exception:
+                # If we don't copy the exception then python keeps adding onto the stack each time it's raised.
+                raise copy.copy(self._exception) from None
+
+            try:
+                if isinstance(client, RESTApp):
+                    client = client.acquire()
+                    client.start()
+
+                if not self._token:
+                    response = await client.authorize_access_token(
+                        client=self._client_id,
+                        client_secret=self._client_secret,
+                        code=self._code,
+                        redirect_uri=self._redirect_uri,
+                    )
+                else:
+                    response = await client.refresh_access_token(
+                        client=self._client_id, client_secret=self._client_secret, refresh_token=self._refresh_token
+                    )
+
+            except errors.ClientHTTPResponseError as exc:
+                if not isinstance(exc, errors.RateLimitTooLongError):
+                    # If we don't copy the exception then python keeps adding onto the stack each time it's raised.
+                    self._exception = copy.copy(exc)
+                await client.close()
+                raise
+
+            # Expires in is lowered a bit in-order to lower the chance of a dead token being used.
+            self._expire_at = time.monotonic() + math.floor(response.expires_in.total_seconds() * 0.99)
+            self._token = f"{response.access_token}"
+            self._refresh_token = response.refresh_token
+            await client.close()
+            return self._token
+
+    def invalidate(self, token: typing.Optional[str] = None) -> None:
+        if not token or token == self._token:
+            self._expire_at = 0.0
+            self._token = None
+            self._refresh_token = None
+            self._code = None
+
+
 class _RESTProvider(traits.RESTAware):
     __slots__: typing.Sequence[str] = ("_entity_factory", "_executor", "_rest")
 
--- a/changes/unknown.feature.md
+++ b/changes/unknown.feature.md
@@ -1,0 +1,2 @@
+Add `UserCredentialsStrategy` to `hikari\impl\rest.py` similar to `ClientCredentialsStrategy` to allow simple OAuth2
+token generation and refreshing


hikari/impl/rest.py
```diff
@@ -214,6 +214,132 @@ def invalidate(self, token: typing.Optional[str]) -> None:
...
+                return self._token
+
+            if self._exception:
+                # If we don't copy the exception then python keeps adding onto the stack each time it's raised.
+                raise copy.copy(self._exception) from None
+
+            try:
+                if isinstance(client, RESTApp):
+                    client = client.acquire()
+                    client.start()
```
davfsa, line 307: This should not be managed by this class

hikari/impl/rest.py
```diff
@@ -214,6 +214,132 @@ def invalidate(self, token: typing.Optional[str]) -> None:
...
+                    # If we don't copy the exception then python keeps adding onto the stack each time it's raised.
+                    self._exception = copy.copy(exc)
+                await client.close()
+                raise
+
+            # Expires in is lowered a bit in-order to lower the chance of a dead token being used.
+            self._expire_at = time.monotonic() + math.floor(response.expires_in.total_seconds() * 0.99)
+            self._token = f"{response.access_token}"
+            self._refresh_token = response.refresh_token
+            await client.close()
```
davfsa, line 332: Same here

This will be always called by an active rest client

hikari/impl/rest.py
```diff
@@ -214,6 +214,132 @@ def invalidate(self, token: typing.Optional[str]) -> None:
             self._token = None
 
 
+class UserCredentialsStrategy(rest_api.TokenStrategy):
```
davfsa, line 217:
```suggestion
class OAuthCredentialsStrategy(rest_api.TokenStrategy):
```

hikari/impl/rest.py
```diff
@@ -214,6 +214,132 @@ def invalidate(self, token: typing.Optional[str]) -> None:
...
+        """Sequence of scopes this token strategy authenticates for."""
+        return self._scopes
+
+    @property
+    def token_type(self) -> applications.TokenType:
+        return applications.TokenType.BEARER
+
+    async def acquire(self, client: rest_api.RESTClient | RESTApp) -> str:
+        if not self._code:
+            raise "Token has been invalidated. Unable to get current or new token"
```
davfsa, line 291: This should raise a (I believe) RuntimeError and should be documented

hikari/impl/rest.py
```diff
@@ -214,6 +214,132 @@ def invalidate(self, token: typing.Optional[str]) -> None:
...
+            except errors.ClientHTTPResponseError as exc:
+                if not isinstance(exc, errors.RateLimitTooLongError):
+                    # If we don't copy the exception then python keeps adding onto the stack each time it's raised.
+                    self._exception = copy.copy(exc)
+                await client.close()
+                raise
+
+            # Expires in is lowered a bit in-order to lower the chance of a dead token being used.
+            self._expire_at = time.monotonic() + math.floor(response.expires_in.total_seconds() * 0.99)
+            self._token = f"{response.access_token}"
```
davfsa, line 330:
```suggestion
            self._token = str(response.access_token)
```

hikari/impl/rest.py
```diff
@@ -214,6 +214,132 @@ def invalidate(self, token: typing.Optional[str]) -> None:
...
+        self._redirect_uri = redirect_uri
+
+    @property
+    def client_id(self) -> snowflakes.Snowflake:
+        """ID of the application this token strategy authenticates with."""
+        return self._client_id
+
+    @property
+    def _is_expired(self) -> bool:
+        return time.monotonic() >= self._expire_at
```
davfsa, line 278: This should be a function, not a property

changes/1558.feature.md
```diff
@@ -0,0 +1,2 @@
+Add `UserCredentialsStrategy` to `hikari\impl\rest.py` similar to `ClientCredentialsStrategy` to allow simple OAuth2
+token generation and refreshing
```
davfsa, line 2:
```suggestion
Add `OAuthCredentialsStrategy` to `hikari.impl.rest` for OAuth2 flow tokens.
```

davfsa (changes requested):
Thank you for having a go at this!

Testing should be similar to `ClientAuthorizationStrategy`, but if you run into any issues, feel free to message me on Discord or posting it here and will help you out!

hikari/impl/rest.py
```diff
@@ -214,6 +214,132 @@ def invalidate(self, token: typing.Optional[str]) -> None:
             self._token = None
 
 
+class UserCredentialsStrategy(rest_api.TokenStrategy):
+    """Strategy class for handling client credential OAuth2 authorization.
```
davfsa, line 218: Forgot to add it to the review, but this information is also wrong

davfsa (changes requested):

hikari/impl/rest.py
theonlydoublee: Change client to user or is more than that incorrect

davfsa: Client is not really appropriate here. Should just be OAuth

theonlydoublee:
I ran into an issue with making the tests for `OAuthCredentialsStrategy` and the last tests I have done are getting the error `TypeError: object Mock can't be used in 'await' expression`.  The code is the only tests that are erroring and all for that same error.  Altered from `ClientCredentialsStrategy`

Code: [https://pastebin.com/pdFk7Fxf](https://pastebin.com/pdFk7Fxf)

Jonxslays:
Use a `mock.AsyncMock` instead of `mock.Mock`

theonlydoublee:
Getting an error of `pytest.PytestUnraisableExceptionWarning: Exception ignored in: <coroutine object AsyncMockMixin._execute_mock_call at 0x00000267A82C7A70>`

Full Nox Traceback: [https://pastebin.com/f6qkWLkU]( https://pastebin.com/f6qkWLkU)
Failing Test Code: [https://pastebin.com/MzV1D7BX](https://pastebin.com/MzV1D7BX)
Updated Class: [https://pastebin.com/TbHKr2Ad](https://pastebin.com/TbHKr2Ad)

tests/hikari/impl/test_rest.py
```diff
@@ -292,6 +292,199 @@ def test_invalidate_when_token_is_stored_token(self):
...
+
+        mock_rest.refresh_access_token.assert_awaited_once_with(
+            client=123456789, client_secret="123123123", refresh_token="refresh.mock.token"
+        )
+        assert new_token != token
+        assert new_token == "mockmock.tokentoken.mocktoken"
+
+    @pytest.mark.asyncio()
+    async def test_acquire_handles_token_being_set_before_lock_is_acquired(self, mock_token):
+        lock = asyncio.Lock()
```
davfsa, line 395: Why the use of a lock here?

tests/hikari/impl/test_rest.py
```diff
@@ -292,6 +292,199 @@ def test_invalidate_when_token_is_stored_token(self):
...
+        strategy._expire_at = 10.0
+        strategy._token = "token"
+
+        strategy.invalidate(None)
+
+        assert strategy._expire_at == 0.0
+        assert strategy._token is None
+
+    def test_invalidate_when_token_is_stored_token(self):
+        strategy = rest.OAuthCredentialsStrategy(client=123456789, client_secret="123123123", auth_code="auth#code", redirect_uri="https://web.site/auth/discord")
```
davfsa: This strategy could be extracted to a fixture and used across all tests, instead of having to instantiate it everywhere

tests/hikari/impl/test_rest.py
```diff
@@ -292,6 +292,199 @@ def test_invalidate_when_token_is_stored_token(self):
...
+#############################
+# OAuthCredentialsStrategy #
+#############################
+
+
+class TestOAuthCredentialsStrategy:
+    @pytest.fixture()
+    def mock_token(self):
+        return mock.Mock(
+            applications.PartialOAuth2Token,
```
davfsa, line 305: Spec setting is really inefficient, and unless we have a good reason to do so, shouldnt be doing it

theonlydoublee:
Thought I fixed those typing errors for mypy

hikari/impl/rest.py
```diff
@@ -243,37 +243,36 @@ class UserCredentialsStrategy(rest_api.TokenStrategy):
...
         self._redirect_uri = redirect_uri
 
     @property
     def client_id(self) -> snowflakes.Snowflake:
         """ID of the application this token strategy authenticates with."""
         return self._client_id
 
-    @property
     def _is_expired(self) -> bool:
         return time.monotonic() >= self._expire_at
```
davfsa: Should be moved under all the properties

tests/hikari/impl/test_rest.py
theonlydoublee: Followed tests from ClientCredentialsStrategy

tests/hikari/impl/test_rest.py
theonlydoublee: Could you explain what it means forspec setting?

tests/hikari/impl/test_rest.py
davfsa: Ah, ok. Got why thats there now. Its to prevent the gatherer from attempting to go too quickly. Dont think its necessary in this case, but its anyones guess. Ill do some tests and see if it changes anything

tests/hikari/impl/test_rest.py
theonlydoublee: Removed the spec and tests still seem to work

tests/hikari/impl/test_rest.py
theonlydoublee: Running the test without locking seems fine.  Ran multiple times and no errors

theonlydoublee:
Everything should be ready

hikari/api/rest.py
```diff
@@ -2960,6 +2960,12 @@ async def authorize_access_token(
     ) -> applications.OAuth2AuthorizationToken:
         """Authorize an OAuth2 token using the authorize code grant type.
 
+        .. warning::
+            There is no way to ensure what scopes are granted in the token,
```
FasterSpeeding, line 2964: There's seemingly no reason to put this note here since there isn't a `scopes` parameter? Also this change seems outside the scope of this PR

hikari/api/rest.py
```diff
@@ -2995,16 +3001,12 @@ async def refresh_access_token(
         client: snowflakes.SnowflakeishOr[guilds.PartialApplication],
         client_secret: str,
         refresh_token: str,
-        *,
-        scopes: undefined.UndefinedOr[
-            typing.Sequence[typing.Union[applications.OAuth2Scope, str]]
-        ] = undefined.UNDEFINED,
```
FasterSpeeding, line 3001: This shouldn't be removed. This is a part of the Oauth2 spec and Discord does use this field these days.

FasterSpeeding:
I don't know how I feel about this tbh. The reason this wasn't implemented in the first place was since it felt like having this implemented as a standard thing would just be encouraging bad behaviour (which is still the case).

When handling Oauth2 tokens you'd really want to be storing the current version of the token (likely in a database) to persist it between bot restarts and only refreshing it after it's expired but this doesn't really allow for that.

Also I'm not sure what the use case is for this.

theonlydoublee:
I was thinking of using it for a restapp for interacting with user's information.  If this isn't used, I'll use it for my personal project without it being in hikari itself.

davfsa:
> When handling Oauth2 tokens you'd really want to be storing the current version of the token (likely in a database) to persist it between bot restarts and only refreshing it after it's expired but this doesn't really allow for that.

Could that be fixed by just allowing passing the token and refresh token to the strategy?

PR #1569, opened 2023-04-02
Revert changing the default nox -s ordering

FasterSpeeding:
### Summary
Without this running `nox -s` will lead to flake8 running before reformatting, potentially giving unnecessary black would reformat this errors.

### Checklist

### DIFF
--- a/pipelines/format.nox.py
+++ b/pipelines/format.nox.py
@@ -32,7 +32,7 @@ from pipelines import nox
 GIT = shutil.which("git")
 
 
-@nox.session(default_session=True)
+@nox.session()
 def reformat_code(session: nox.Session) -> None:
     """Remove trailing whitespace in source, run isort, codespell and then run black code formatter."""
     session.install(*nox.dev_requirements("formatting"))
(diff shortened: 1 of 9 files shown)


pipelines/nox.py
```diff
@@ -30,7 +30,7 @@
 from pipelines import config as _pipelines_config
 
 # Default sessions should be defined here
-_options.sessions = []
+_options.sessions = ["reformat-code", "pytest-all-features", "flake8", "slotscheck", "mypy", "verify-types"]
```
FasterSpeeding, line 33: Seems more compelling to run `pytest-all-features` by default rather than `pytest` and if install speed is a concern `reuse_venv=True` could help with that

pipelines/format.nox.py
```diff
@@ -43,7 +43,7 @@ def reformat_code(session: nox.Session) -> None:
     session.run("black", *config.PYTHON_REFORMATTING_PATHS)
 
 
-@nox.session(default_session=True, venv_backend="none")
+@nox.session(venv_backend="none")
 def check_trailing_whitespaces(session: nox.Session) -> None:
```
FasterSpeeding, line 47: Doesn't really make sense to run this by default when `reformat-code` is already calling the same logic just in write mode

pipelines/nox.py
davfsa: `reuse_venv` is True by default. The reason it was not put there when we split the pipelines was because of not installing packages that might not be installable in the specific environment (like the need for a compiler). Nowadays, that is mostly useless, mostly because we even install `mypy` with all speedups.

davfsa (approved):

PR #1570, opened 2023-04-02
Add deprecation utility decorator

FasterSpeeding:
### Summary
Gonna wait until the next MyPy release for this

### Checklist


hikari/internal/deprecation.py
```diff
@@ -99,3 +99,16 @@ def warn_deprecated(
...
+    from typing_extensions import deprecated
+
+else:
+
+    def deprecated(*args):
+        """Mark a function, overload, or class as deprecated for type-checkers.
+
+        This has no runtime side-effects.
+        """
+        return lambda value: value
```
davfsa, line 114:
```suggestion
    def deprecated(func):
        """Mark a function, overload, or class as deprecated for type-checkers.

        This has no runtime side-effects.
        """
        return func
```
And make use of ParamSpec?

FasterSpeeding: This also decorates classes and paramspec won't work for that and isn't necessary (the return type is just `Callable[[_T], _T]` + this isn't being type checked since its essentially in a `if not typing.TYPE_CHECKING` block

davfsa:
@FasterSpeeding mypy was bumped to 1.2.0 in #1583, which should unblock this pr :)

PR #1577, opened 2023-04-02
Configure black to skip magic trailing commas

FasterSpeeding:
### Summary
These are usually accidental or coincidental and this just makes the behaviour more consistent.

The only place I can think of where you'd actively want this behaviour is for examples, but examples in documentation aren't reformatted by black right now anyways.

### Checklist


davfsa:
There are a lot of cases where I dont like how black behaves and the magic trailing commas help alleviate that and make it easier to read. Going quickly through the changes, I can see why this is an option, but im a bit torn

davfsa (approved):
Overall code is much cleaner, so lets go with it

PR #1621, opened 2023-05-25
Add new interaction channel attribute

davfsa:
### Checklist


davfsa:
I am not sure whether to get rid of `BaseInteraction.get_channel` and `BaseInteraction.fetch_channel`, as they are a bit replaced here. The only difference is the more concrete type

hikari/impl/entity_factory.py
```diff
@@ -2542,7 +2542,7 @@ def deserialize_command_interaction(
             guild_id=guild_id,
             guild_locale=locales.Locale(payload["guild_locale"]) if "guild_locale" in payload else None,
             locale=locales.Locale(payload["locale"]),
-            channel_id=snowflakes.Snowflake(payload["channel_id"]),
+            channel=self.deserialize_partial_channel(payload["channel"]),
```
FasterSpeeding, line 2545: the channel objects here include a lot more fields than just the partial channel's fields

e.g.
```json
{
    "type": 0,
    "topic": null,
    "theme_color": null,
    "rate_limit_per_user": 0,
    "position": 20,
    "permissions": "140737488355327",
    "parent_id": "561885448721399808",
    "nsfw": false,
    "name": "dwads",
    "last_message_id": "1114922806820479057",
    "id": "717586538925523004",
    "icon_emoji": {
        "name": "\ud83e\udd16",
        "id": null
    },
    "guild_id": "561884984214814744",
    "flags": 0,
    "default_auto_archive_duration": 60
}
```
```json
{
    "type": 1,
    "last_message_id": "1096465893028003910",
    "id": "463185043796262922",
    "flags": 0
}
```

FasterSpeeding: This was also mentioned in the PR which documented this field https://github.com/discord/discord-api-docs/pull/6051#discussion_r1156415144, just another case of Discord having shit docs tbh

FasterSpeeding: This is worth looking into again

davfsa: Changed, this seem to have been a `InteractionChannel` object

hikari/impl/entity_factory.py
```diff
@@ -2532,28 +2499,35 @@ def _deserialize_interaction_member(
...
+    def _deserialize_interaction_channel(
+        self, payload: data_binding.JSONObject
+    ) -> base_interactions.InteractionChannel:
+        channel_id = snowflakes.Snowflake(payload["id"])
+
+        thread_metadata = (
+            self._deserialize_thread_metadata(payload["thread_metadata"]) if "thread_metadata" in payload else None
+        )
+
+        return base_interactions.InteractionChannel(
```
FasterSpeeding, line 2511: what about the other fields? while the documentation for resolved channels is likely just out of date in its list of included fields, the documentation for `interaction.channels` doesn't specify fields which would be included and there's just fields (including channel type specific fields) being missed here for that

FasterSpeeding: Bump

davfsa: For now, I will keep it at just resolving the fields that are documented. If Discord document more in the future, this can be extended. Otherwise, we will have a mess of objects because they might be different in different circumstances that are not documented.

If you spot any fields that are consistency sent, we can try and get it documented in the docs and add it to hikari

FasterSpeeding: interaction.channel does not have a specific set of fields documented for it, that only applies to resolved options

FasterSpeeding: would be nice for interaction.channel to not miss fields

PR #1653, opened 2023-07-10
Support text in stage channels

novanai:
### Summary

Support text in stage channels

### Checklist


tests/hikari/impl/test_entity_factory.py
```diff
@@ -2034,6 +2036,7 @@ def test_deserialize_guild_stage_channel_with_null_fields(self, entity_factory_i
                 "user_limit": 3,
                 "rtc_region": None,
                 "type": 6,
+                "last_message_id": None,
```
FasterSpeeding, line 2039: Should prob add `assert channel.last_message_id is None` to this test case and `test_deserialize_guild_stage_channel_with_unset_fields`

novanai: done :)

hikari/impl/entity_factory.py
```diff
@@ -1173,6 +1173,9 @@ def deserialize_guild_stage_channel(
             snowflakes.Snowflake(overwrite["id"]): self.deserialize_permission_overwrite(overwrite)
             for overwrite in payload["permission_overwrites"]
         }
+        last_message_id = (
+            snowflakes.Snowflake(msg_id) if (msg_id := payload.get("last_message_id")) is not None else None
+        )
```
davfsa, line 1178:
```suggestion
        last_message_id = snowflakes.Snowflake(payload["last_message_id"]) if "last_message_id" in payload else None
```

Should make this a nicer one-liner

novanai: It was at first, I dunno if black was too happy with it though

davfsa: Line length is bellow 120, so shouldn't be too many issues. Regardless, reads nicer IMO

novanai: Okay :)

davfsa (approved):

PR #1662, opened 2023-07-18
Allow passing initial `seq`, `session_id` and `resume_url` attributes to the shard, as well as retrieving them.

davfsa:
- Additionally export all impl classes to `hikari` namespace


hikari/api/shard.py
```diff
@@ -104,6 +104,21 @@ def is_connected(self) -> bool:
...
         """Return the total number of shards expected in the entire application."""
 
+    @property
+    @abc.abstractmethod
+    def session_id(self) -> typing.Optional[str]:
+        """The session id for the shard."""
+
+    @property
+    @abc.abstractmethod
+    def seq(self) -> typing.Optional[int]:
```
FasterSpeeding, line 114: guaranteeing these (esp seq) for all shard impls would conflict with shard impls which aren't running the shard logic in the current process

davfsa: We are not going to make a "extremely generic" API, because then we might as well have an empty class, because each implementation could do whatever they want and decide how it works.

The idea for our API is to expose what external interface we expect our components to have.

If someone wants to make their own interface and implementation of what we do, they can always define and implement their own

PR #1679, opened 2023-08-04
feat: add CacheSettings.only_my_member

### LINKED ISSUE #1678: Option to only cache the bot member
## Summary

This requests aims to introduce a new option for enabling the members cache ONLY for the bot member.
Similar to `CacheComponents.ME` but for the member.

## Why is this needed?

The bot member is very useful for permissions calculations and enabling the members caches every member, increasing memory consumption and CPU consumption on startup.

## Ideal implementation

see pr :)

Le0Developer:
### Summary

Add CacheSettings.only_my_member to restrict the member cache to only cache the bot.
This is useful for situations when you only need the bot member (eg permission calculations).

### Checklist

### Related issues

Closes #1678


davfsa (dismissed):
LGTM

hikari/api/config.py
```diff
@@ -199,3 +199,13 @@ class CacheSettings(abc.ABC):
     @abc.abstractmethod
     def components(self) -> CacheComponents:
         """Cache components to use."""
+
+    @property
+    @abc.abstractmethod
+    def only_cache_my_member(self) -> bool:
```
davfsa, line 205:
```suggestion
    def only_my_member(self) -> bool:
```
Maybe better naming? Since everything relates to caching

hikari/api/config.py
```diff
@@ -199,3 +199,13 @@ class CacheSettings(abc.ABC):
...
+        """Reduce the members cache to only the bot itself.
```
davfsa, line 206: A little context on why this could be useful could be added here

changes/idk.feat.md
```diff
@@ -0,0 +1 @@
+Add `CacheSettings.only_cache_my_member` to only cache the bot member.
```
davfsa, line 1: File should be renamed to `1679.feature.md`

hikari/impl/event_manager.py
```diff
@@ -370,6 +375,7 @@ async def on_guild_create(  # noqa: C901, CFQ001 - Function too complex and too
             self._auto_chunk_members
             and self._intents & intents_.Intents.GUILD_MEMBERS
             and (payload.get("large") or not presences_declared)
+            and (self._cache is None or not self._cache.settings.only_my_member)
```
Le0Developer, line 378: This disables auto_chunk_members when using the `only_my_member` option.
Maybe `auto_chunk_members` should be changed to None by default, which means "auto", so that it can still be explicitly enabled by passing `True` (defaults to `True` currently)

hikari/impl/event_manager.py
```diff
@@ -367,7 +372,13 @@ async def on_guild_create(  # noqa: C901, CFQ001 - Function too complex and too
         # payload if presence intents are also declared, so if this isn't the case then we also want
         # to chunk small guilds.
         if (
-            self._auto_chunk_members
+            (
+                self._auto_chunk_members is True
```
FasterSpeeding, line 376: This `auto_chunk_members` change seems to just over-complicate it imo.
Why can't the `self._cache_enabled_for(config.CacheComponents.MEMBERS)` bit just have a check added on to make it 

```py
(
    self._cache and self._cache_enabled_for(config.CacheComponents.MEMBERS)
    and not self._cache.settings.only_my_member
)
```

and avoid changing the type of `auto_chunk_members` all together

hikari/impl/event_manager.py
```diff
@@ -371,7 +376,10 @@ async def on_guild_create(  # noqa: C901, CFQ001 - Function too complex and too
             and self._intents & intents_.Intents.GUILD_MEMBERS
             and (payload.get("large") or not presences_declared)
             and (
-                self._cache_enabled_for(config.CacheComponents.MEMBERS)
+                (
+                    self._cache_enabled_for(config.CacheComponents.MEMBERS)
+                    and (self._cache is None or not self._cache.settings.only_my_member)
+                )
```
davfsa, line 382: This definitely needs some commenting.

Not only that, maybe it's worth extracting it to a variable above to make it more legible

Le0Developer:
```suggestion
                (
                    # disable auto chunking if we are only interested in the bot member (only_my_member=True)
                    self._cache_enabled_for(config.CacheComponents.MEMBERS)
                    and (self._cache is None or not self._cache.settings.only_my_member)
                )
```

davfsa (approved):
Everything looks good (unless I missed something)

PR #1683, opened 2023-08-09
Add custom status support

davfsa:


davfsa:
`name` is still required to be passed for custom statuses for bots, even if its not used, which is not the best DX.

Might be worth having some syntactical sugar to default `name` to the value of `state` if its passed.

Thoughts?

davfsa:
> `name` is still required to be passed for custom statuses for bots, even if its not used, which is not the best DX.
> 
> Might be worth having some syntactical sugar to default `name` to the value of `state` if its passed.
> 
> Thoughts?

Implemented

PR #1702, opened 2023-08-30
Fix attachments not being removed in message edits

novanai:
### Checklist


FasterSpeeding:
if any message edit endpoint isn't properly handling `None` for `attachmnts` that should be reported to discord

davfsa (approved):

PR #1804, opened 2024-01-05
Add defaults for auto-populated selects

### LINKED ISSUE #1776: Implement default select values
https://github.com/discord/discord-api-docs/pull/6442

hypergonial:
### Summary
Add `default_values` attribute for auto-populated selects. This did require some fairly **significant restructuring** to make type-safe, and honestly I'm still not quite happy with the result, so feedback is welcome.

The select types have been completely seperated, so the new hierarchy of components looks something like this:

```
SelectMenu
TextSelect(SelectMenu)
AutoPopulatingSelect(SelectMenu)
ChannelSelect(AutoPopulatingSelect) 
RoleSelect(AutoPopulatingSelect)
UserSelect(AutoPopulatingSelect)
MentionableSelect(AutoPopulatingSelect)
```

This PR also deprecates `ActionRowBuilder.add_select_menu()` as type-specific methods have been added for each select type. I feel like seperating the select types is better for type safety and potentially more future-proof if Discord decides add specific behaviours to a given select type down the line.

### Example

```py
row = rest.build_message_action_row().add_user_menu(
    "Test",
    placeholder="Amongus",
    min_values=1,
    max_values=2,
    default_values=[
        hikari.impl.SelectDefaultBuilder(
            163979124820541440, type=hikari.SelectDefaultType.USER
        ),
        hikari.impl.SelectDefaultBuilder(
            1184650183401803906, type=hikari.SelectDefaultType.USER
        ),
    ],
)
```

I tried an alternative interface where the snowflake's type could be inferred from the select's type, however this is not possible due to it being possible to mix default types in mentionable selects, so at the moment this is the best I could come up with. 

The implementation tries to be type-safe to the extent that adding an invalid default type fails type-checking:

```py
row = rest.build_message_action_row().add_user_menu(
    "Test",
    placeholder="Amongus",
    min_values=1,
    max_values=2,
    default_values=[
        hikari.impl.SelectDefaultBuilder( # Type error
            163979124820541440, type=hikari.SelectDefaultType.ROLE
        )
    ],
)
``` 

### Checklist

### Related issues

Closes #1776

### DIFF
--- a/hikari/api/special_endpoints.py
+++ b/hikari/api/special_endpoints.py
@@ -46,6 +46,11 @@ __all__: typing.Sequence[str] = (
     "InteractionModalBuilder",
     "MessageActionRowBuilder",
     "ModalActionRowBuilder",
+    "SelectDefaultBuilder",
+    "AutoPopulatedSelectMenuBuilder",
+    "UserSelectMenuBuilder",
+    "RoleSelectMenuBuilder",
+    "MentionableSelectMenuBuilder",
 )
 
 import abc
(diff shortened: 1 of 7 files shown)


davfsa:
I need to have a think about the restructure, but will most probably forget about it, so please ping me in the near future if I dont respond :)

hikari/api/special_endpoints.py
```diff
@@ -2086,6 +2207,10 @@ def add_select_menu(
         `MessageActionRowBuilder.add_channel_menu` and
         `MessageActionRowBuilder.add_text_menu`.
 
+        .. deprecated:: 2.0.0.dev123
```
FasterSpeeding, line 2210: Just leaving this here as a note that these versions have to be bumped before this is merged

PR #1810, opened 2024-01-20
Migrate documentation to mkdocs

### LINKED ISSUE #1806: Consider using mkdocs material
### Summary

We should consider using [mkdocs material](https://squidfunk.github.io/mkdocs-material/) for the documentation.

I have the following reasons as to why:
- Much better search that is very responsive
- Nicer looking interface
- The entire site doesn't have to reload for every navigation (Instant Navigation)
- Highly extensible and customizable
- Properly builds `objects.inv` so it is actually usable by other repos
- Supports advanced markdown features such as [tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/) & admonishments ([full list](https://squidfunk.github.io/mkdocs-material/reference/))
- Is also [supported by readthedocs](https://docs.readthedocs.io/en/stable/intro/getting-started-with-mkdocs.html) so documentation versioning & deployment should be simple

### Why is this needed?

The current documentation is very slow to load, the search is not very responsive (if it works at all), and `objects.inv` is almost completely unusable by other documentation generators.

### Ideal implementation

Something like [this](https://arc.hypergonial.com/)

If given the go-ahead, **I'd be willing to work on a PR** to switch the documentation to mkdocs-material with a similar configuration as the one I use, linked above.

### Checklist

hypergonial:
### Summary
Reimplement the documentation in mkdocs material.

Deployments:

- ~~[Github Pages](https://hikari.hypergonial.com/)~~ outdated
- [ReadtheDocs](https://hikari-rtfd.hypergonial.com/en/latest/)

I think it is worth considering an alternative to readthedocs (like pages) given the performance of readthedocs is so abysmal comparatively, however, in that case, versioning needs to be considered.

### Stuff missing/not done

- [x] ~~Docstrings were not adjusted for the new admonishment styles~~
- [x] ~~PR branch needs to be removed from github action triggers (or pages action removed entirely)~~ Pages action was removed

These will be done if #1806 is accepted and a documentation host is decided on, so this PR can proceed.

### Related issues
Closes #1806 

Feedback would be appreciated.

P.S.: I sneaked `arc` into the readme.


davfsa:
Started working on this a while while back (https://github.com/davfsa/hikari/tree/task/mkdocs) but the main showstopper was https://github.com/mkdocstrings/griffe/issues/96 and eventually waiting for the module to come out of sponsor only territory, which seems like it is the case now.

Ill append the work I did onto this.

I also would like to not use github pages since repos have a size limit and we will quickly hit it with how heavy our docs are. Readthedocs will have to suffice for now, as well as maybe finding a nicer way to display the version switcher, which i have some ideas for.

Thanks for working on this!

hypergonial:
Just tested how `objects.inv` behaves, and it seems like other docs still only get partial (albeit much more than none) information. The top-level module `hikari` is not picked up for some reason, and some objects like `UNDEFINED` or `Resourcish` are also not linked properly, unsure why.

image

davfsa:
After some conflict merging, this is what I came up with.

image

If that looks alright, I can push it and we can see if it requires any tweaks

---

Note:
I already have the admonition changes stashed, the only thing that would need changing is the links

hypergonial:
> After some conflict merging, this is what I came up with.
> 
> image
> 
> If that looks alright, I can push it and we can see if it requires any tweaks
> 
> Note: I already have the admonition changes stashed, the only thing that would need changing is the links

Not sure on that purple tbh, but otherwise looks nice!

davfsa:
Pushed my changes, please have a look at lmk what you think/would like changed (anybody reading this is also welcome to see and try it out here https://hikari-py--1810.org.readthedocs.build/en/1810/)

Furthermore, when it comes to versioning, might be "ok" to leave it with the little widget at the bottom, since the only other thing I can think about is highjacking mkdocs mike integration and providing the version.json ourselves in the gh-pages branch (and then somehow figuring out how to disable the readthedocs widget)

hypergonial:
Autoscrolling when searching is broken, it doesn't scroll to the object that was searched for. This wasn't an issue before the merge.

docs/gen_ref_pages.py
```diff
@@ -0,0 +1,56 @@
...
+
+with mkdocs_gen_files.open("reference/summary.md", "w") as nav_file:
+    for item in nav.items():
+        path = pathlib.Path(item.filename).with_suffix("")
+
+        if path.name == "index":
+            path = path.parent
+
+        full_name = ".".join(path.parts)
+        nav_file.write("    " * item.level + f"* [{full_name}]({item.filename})\n")
```
hypergonial, line 56: Why was this changed? mkdocs_gen_files can just write the literate nav as is

davfsa:
> Autoscrolling when searching is broken, it doesn't scroll to the object that was searched for. This wasn't an issue before the merge.

Works for me :thinking:

---

Also, now that I get to use it a bit more, god is it painful to use when compared to the ones we currently have, at least in readthedocs.

Also, fun fact, one build of the docs alone is 36MB, which is quite high, so getting somewhere to host this could be feasible, but expensive after a while

docs/gen_ref_pages.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+import pathlib
+
+import mkdocs_gen_files
+
+nav = mkdocs_gen_files.Nav()
+
+for path in sorted(pathlib.Path("hikari").rglob("*.py")):
+    module_path = path.relative_to(".").with_suffix("")
```
hypergonial, line 13: Probably shouldn't rely on CWD for this?

docs/javascript/mathjax.js
```diff
@@ -0,0 +1,16 @@
+window.MathJax = {
+    tex: {
+        inlineMath: [["\\(", "\\)"]],
```
hypergonial, line 3: We use LaTeX?

mkdocs.yml
```diff
@@ -27,69 +34,87 @@ theme:
...
-  - pymdownx.inlinehilite
-  - pymdownx.snippets
+  - pymdownx.arithmatex:
+      generic: true
+  - pymdownx.betterem:
+      smart_enable: all
   - pymdownx.tabbed:
       alternate_style: true
-      combine_header_slug: true
-      slugify: !!python/object/apply:pymdownx.slugs.slugify
```
hypergonial: Was this removed intentionally?

docs/javascript/mathjax.js
davfsa: Yep, in the explanation for ratelimits and in some random places

docs/gen_ref_pages.py
davfsa: mkdocs will ensure they get build in a specific way, which i would rather than traversing the paths

mkdocs.yml
davfsa: Nope, nice catch

hypergonial:
> > Autoscrolling when searching is broken, it doesn't scroll to the object that was searched for. This wasn't an issue before the merge.
> 
> Works for me 🤔
> 
> Also, now that I get to use it a bit more, god is it painful to use when compared to the ones we currently have, at least in readthedocs.
> 
> Also, fun fact, one build of the docs alone is 36MB, which is quite high, so getting somewhere to host this could be feasible, but expensive after a while

That's why I suggested seeking alternative hosting. Readthedocs is really awful to use, but other hosts do not seem to have that issue afaict.

davfsa:
> That's why I suggested seeking alternative hosting. Readthedocs is really awful to use, but other hosts do not seem to have that issue afaict.

Any suggestions are welcome, I dont know any

hypergonial:
> > Autoscrolling when searching is broken, it doesn't scroll to the object that was searched for. This wasn't an issue before the merge.
> 
> Works for me 🤔
> 
> Also, now that I get to use it a bit more, god is it painful to use when compared to the ones we currently have, at least in readthedocs.
> 
> Also, fun fact, one build of the docs alone is 36MB, which is quite high, so getting somewhere to host this could be feasible, but expensive after a while

Oh this works on Firefox but broke under Chromium, I just tested. That's fun.

docs/gen_ref_pages.py
davfsa: Left-over from my past testing ig

davfsa: Just realised why I had it. The way mkdocs resolves page titles is weird and it will resolve first the nav title instead of the page title, leading to:
image

instead of the `hikari.sessions` that I would like

hypergonial:
> > That's why I suggested seeking alternative hosting. Readthedocs is really awful to use, but other hosts do not seem to have that issue afaict.
> 
> Any suggestions are welcome, I dont know any

One silly idea I had is that we could split the main offenders (`hikari.channels` and `hikari.guilds`) out into multiple files so that mkdocs doesn't freeze the whole browser trying to load those 2 pages.

docs/gen_ref_pages.py
hypergonial: Hmm

hypergonial:
> I did some more testing, on Chromium, even clicking items in the toc doesn't seem to work, the scrolling just stops briefly after starting. Maybe it is to do with the custom CSS you added?

Yeah this is definitely the case, smooth scrolling breaks jump actions on Chromium.

hypergonial:
Favicon is also failing to load on Chromium:

image

dev-requirements/mkdocs.txt
```diff
@@ -0,0 +1,23 @@
+mkdocs==1.5.3
```
hypergonial, line 1: I don't think this is necessary, since mkdocs-material already depends on it, and it could cause versioning conflicts.

davfsa:
> Yeah this is definitely the case, smooth scrolling breaks jump actions on Chromium.

Reverted it. Seems to only happen on large pages, its weird

davfsa:
> Favicon is also failing to load on Chromium: (both mirrors)
> 
> image

Its not even being requested through http, so its really confusing

hypergonial:
> > Favicon is also failing to load on Chromium: (both mirrors)
> > image
> 
> Its not even being requested through http, so its really confusing

Yeah I don't understand it either, weird.

dev-requirements/mkdocs.txt
davfsa: Rather incompatibilities produce a error that we will catch when bumping dependencies than broken docs which are hard to fix

mkdocs.yml
```diff
@@ -0,0 +1,165 @@
...
+  - pymdownx.superfences
+  - pymdownx.tilde
+  - pymdownx.emoji:
+      emoji_index: !!python/name:material.extensions.emoji.twemoji
+      emoji_generator: !!python/name:material.extensions.emoji.to_svg
+  - pymdownx.highlight:
+      anchor_linenums: true
+      line_spans: __span
+      pygments_lang_class: true
+  - pymdownx.arithmatex:
```
hypergonial, line 101: This should also be removed if MathJax is not needed.

hypergonial:
Disabling instant navigation definitely helps, I'd keep it disabled.

hypergonial:
Is there a way we can remove this level of unnecessary nesting inside `hikari`?
image

The index page could probably be moved up into `Reference` (and it made clickable)

davfsa:
> Disabling instant navigation definitely helps, I'd keep it disabled.

Good to hear, also improves mobile usage.

The last thing to figure out is the docs switch for readthedocs, which I will look into in a bit :)

hypergonial:
You broke the API reference landing page. :p Also the top-level `hikari.__init__` should be documented somewhere otherwise it will be missing from objects.inv. (although said objects.inv has other issues too so idk)

hypergonial:
> > Disabling instant navigation definitely helps, I'd keep it disabled.
> 
> Good to hear, also improves mobile usage.
> 
> The last thing to figure out is the docs switch for readthedocs, which I will look into in a bit :)

Don't forget the admonish fixes in docstrings & possibly doing smth about `objects.inv`

davfsa:
> You broke the API reference landing page

`/reference/` never worked, if that is what you mean.

> Also the top-level hikari.__init__ should be documented somewhere otherwise it will be missing from objects.inv.
Well then it needs to be added back, no getting around that one. 

I rather have a "useless" drop down than potential issues somewhere else

> objects.inv has other issues too so idk
> possibly doing smth about objects.inv

Mind dumping exactly what you mean? I havent had a look at it, but it should be alright

hypergonial:
> > You broke the API reference landing page
> 
> `/reference/` never worked, if that is what you mean.
> 
> > Also the top-level hikari.**init** should be documented somewhere otherwise it will be missing from objects.inv.
> > Well then it needs to be added back, no getting around that one.
> 
> I rather have a "useless" drop down than potential issues somewhere else
> 
> > objects.inv has other issues too so idk
> > possibly doing smth about objects.inv
> 
> Mind dumping exactly what you mean? I havent had a look at it, but it should be alright

https://hikari-rtfd.hypergonial.com/en/latest/reference/hikari/

image

davfsa:
> This is where you get directed after clicking "Reference" in top nav.

You are coming from an old page, otherwise it would redirect you to `reference/hikari/api`. And that "broken" page you see is actually an interesting 404, which i will have to have a look into. I pressume we need to ignore mkdocs one and use rtfm one

hypergonial:
We could also use https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-sections-with-sections to flatten the navigation a bit, perharps. But this can only be enabled site-wide, or not at all.

davfsa:
> We could also use https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-sections-with-sections to flatten the navigation a bit, perharps. But this can only be enabled site-wide, or not at all.

That will not help, as it will just merge what we see as the left TOC with the right one if it can, which is just bloat

hypergonial:
> > This is where you get directed after clicking "Reference" in top nav.
> 
> You are coming from an old page, otherwise it would redirect you to `reference/hikari/api`. And that "broken" page you see is actually an interesting 404, which i will have to have a look into. I pressume we need to ignore mkdocs one and use rtfm one

image

It seems like readthedocs breaks mkdocs's 404 page entirely, unsure why, it should look smth like this, and this works on github pages.

davfsa:
The 404 issue will be fixed when the docs are actually deployed:

> MkDocs automatically generates a 404.html which Read the Docs will use. However, assets will not be loaded correctly unless you define the [site_url](https://www.mkdocs.org/user-guide/configuration/#site_url) configuration value as your site’s [canonical base URL](https://docs.readthedocs.io/en/stable/guides/canonical-urls.html#mkdocs).

https://docs.readthedocs.io/en/stable/reference/404-not-found.html

hypergonial:
> The 404 issue will be fixed when the docs are actually deployed:
> 
> > MkDocs automatically generates a 404.html which Read the Docs will use. However, assets will not be loaded correctly unless you define the [site_url](https://www.mkdocs.org/user-guide/configuration/#site_url) configuration value as your site’s [canonical base URL](https://docs.readthedocs.io/en/stable/guides/canonical-urls.html#mkdocs).
> 
> https://docs.readthedocs.io/en/stable/reference/404-not-found.html

Oh okay, good.

hypergonial:
The `objects.inv` situation seems to have resolved, now all objects are properly linked! (Except for `hikari.UNDEFINED` for some reason, but `hikari.UndefinedType` is fine? I give up at this point lol)

hypergonial:
Do we want to consider setting the object ordering to `source` to avoid stuff like [this](https://hikari-py--1810.org.readthedocs.build/en/1810/reference/hikari/permissions) from happening? The objects should still be grouped by category though, imo.

See the config [here](https://mkdocstrings.github.io/python/usage/configuration/members/#members_order)

davfsa:
I honestly dont mind those bitflags not being ordered :shrug:

Might make it nicer to search what you are looking for by them being sorted alphabetically than by value

mkdocs.yml
```diff
@@ -0,0 +1,161 @@
...
+            merge_init_into_class: true
+            separate_signature: true
+            show_signature_annotations: true
+            show_source: false
+            show_submodules: false
+            signature_crossrefs: true
+            show_symbol_type_heading: true
+            show_symbol_type_toc: true
+            show_root_toc_entry: false
+            line_length: 70
```
hypergonial, line 159: This can be raised to `80` and it should fill the codeblocks to it's max (without overscroll)

davfsa:
I have pushed the markdown fixes. All that would need to be fixed are the reference links using https://github.com/mkdocstrings/autorefs (included as part of the python handler of mkdocstrings)

hypergonial:
> I have pushed the markdown fixes. All that would need to be fixed are the reference links using https://github.com/mkdocstrings/autorefs (included as part of the python handler of mkdocstrings)

Are you working on it, or should I go over them?

davfsa:
> Are you working on it, or should I go over them?

They are insanely tricky, since its not as easy as a search and replace, and there are probably hundreds of them. I will work on it tomorrow, most probably end up parsing all the docstrings, trying to link stuff and figuring out which ones link and replacing those

hypergonial:
> > Are you working on it, or should I go over them?
> 
> They are insanely tricky, since its not as easy as a search and replace, and there are probably hundreds (edit: there are 1130) of them. I will work on it tomorrow, most probably end up parsing all the docstrings, trying to link stuff and figuring out which ones link and replacing those

Any updates on this? Should I try to take a look myself?

davfsa:
image

593 more to go (down from +3k)

hypergonial:
> image
> 
> 593 more to go (down from +3k)
> 
> I quite urge you not to touch the branch, as rebasing this will be insanely painful

Dear god..

Understood!

FasterSpeeding (approved):

PR #1814, opened 2024-01-21
Allow subscribing to generic events

davfsa:


hikari/impl/event_manager_base.py
```diff
@@ -379,6 +379,10 @@ def _enabled_for_event(self, event_type: typing.Type[base_events.Event], /) -> b
         return False
 
     def _check_event(self, event_type: typing.Type[typing.Any], nested: int) -> None:
+        # Extract the underlying type from generics
+        if hasattr(event_type, "__origin__"):
```
hypergonial, line 383: Can we not use `typing.get_origin()` here?

PR #1867, opened 2024-04-02
Properly handle initial opcode as being RECONNECT

beagold:
### Checklist

### Related issues
None

### DIFF
--- a/hikari/impl/shard.py
+++ b/hikari/impl/shard.py
@@ -822,7 +822,7 @@ class GatewayShardImpl(shard.GatewayShard):
         initial_op = hello_payload[_OP]
 
         if initial_op == _RECONNECT:
-            # It is impossible that we receive RECONNECT as the initial opcode when the node we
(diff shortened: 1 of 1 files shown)


hikari/impl/shard.py
```diff
@@ -818,16 +818,29 @@ async def _connect(self) -> typing.Tuple[asyncio.Task[None], ...]:
             url=url,
         )
 
-        # Expect initial HELLO
         hello_payload = await self._ws.receive_json()
-        if hello_payload[_OP] != _HELLO:
+        initial_op = hello_payload[_OP]
+
+        if initial_op == _RECONNECT:
+            # It is impossible that we receive RECONNECT as the initial opcode when the node we
```
hypergonial, line 825:
```suggestion
            # It is possible that we receive RECONNECT as the initial opcode when the node we
```

Unless I'm misinterpreting it?

FasterSpeeding (approved):

PR #1877, opened 2024-04-08
Add make_link() to MessageReference

Le0Developer:
### Summary
Adds the `make_link()` function of Message to MessageReference.

### Checklist

- [ ] I have run `nox` and all the pipelines have passed. (my setup is borked)


hikari/messages.py
```diff
@@ -309,6 +309,18 @@ class MessageReference:
...
+    def make_link(self) -> str:
+        """Generate a jump link to this message or channel in case for follow add messages.
+
+        Returns
+        -------
+        str
+            The jump link to the message.
+        """
+        guild_id_str = "@me" if self.guild_id is None else str(self.guild_id)
+        message_id_str = "" if self.id is None else str(self.id)
```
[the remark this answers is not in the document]
Le0Developer: This was copied from the original `make_link`.
I guess this is because mypy/pyright might complain about guild_id_str being a str | hikari.Snowflakeish union?

changes/1877.feature.md
```diff
@@ -0,0 +1,2 @@
+Add `make_link()` function to `MessageReference` class.
+
```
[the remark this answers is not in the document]
Le0Developer: I could swear something complained about this in the past when there wasn't an extra newline

hikari/messages.py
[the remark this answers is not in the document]
Le0Developer: Even then, the `str()` cast in the original is still useless.

hikari/messages.py
```diff
@@ -309,6 +309,18 @@ class MessageReference:
...
+        """Generate a jump link to the referenced message or channel (in case for follow add messages).
+
+        Returns
+        -------
+        str
+            The jump link to the message reference target.
+        """
+        guild_id_str = "@me" if self.guild_id is None else self.guild_id
+        message_id_str = "" if self.id is None else self.id
+        return f"{urls.BASE_URL}/channels/{guild_id_str}/{self.channel_id}/{message_id_str}"
```
FasterSpeeding, line 322: The behaviour of falling back to just making a channel link here r33lw a bit weird to say the least.

For one it gives weirdly formatted channel links which in-client will show as having a random trailing `/`.
And for the other point I don't really see the logic in doing that for a "message reference link", feels like it'd make more sense to just return None in that case to indicate the message can't be linked to.

Le0Developer: The only case where id is None is for **Channel** Follow Add messages, so I think it does make sense to fallback to only the channel link then.
Also the extra / is an oversight, I assumed discord would handle that correctly but alas

FasterSpeeding: I would personally still consider having separate `make_message_link` and `make_channel_link` methods just to keep it more explicit (it's only really called `make_link` on message since it's linking to the message itself)

But also these can just be "link" properties in a similar sense to the _url properties since they don't take any arguments.

Le0Developer: Done

FasterSpeeding (approved):

PR #1915, opened 2024-05-07
Run nox with uv as the backend when available

davfsa:
### Summary
Run nox with uv as the backend when available

### Checklist

~~- [ ] I have made unittests according to the code I have added/modified/deleted.~~ N/A

### Related issues
None


davfsa:
Considering whether we should push for nox to be installed with `uv` through `nox[uv]` and just dropping venv all together

PR #1933, opened 2024-06-03
Add Embed.clear_fields() method

tandemdude:
### Summary
Add `clear_fields` method to `Embed` to remove all existing fields from the embed.

### Checklist

### DIFF
--- a/hikari/embeds.py
+++ b/hikari/embeds.py
@@ -897,8 +897,7 @@ class Embed:
         Embed
             This embed. Allows for call chaining.
         """
-        if self._fields:
-            self._fields = None
+        self._fields = None
         return self
 
     def __repr__(self) -> str:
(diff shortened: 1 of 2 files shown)


hikari/embeds.py
```diff
@@ -889,6 +889,18 @@ def remove_field(self, index: int, /) -> Embed:
...
+    def clear_fields(self) -> Embed:
+        """Remove all existing fields from this embed.
+
+        Returns
+        -------
+        Embed
+            This embed. Allows for call chaining.
+        """
+        if self._fields:
+            self._fields = None
```
davfsa, line 901:
```suggestion
        self._fields = None
```

tandemdude: 🆗

PR #1958, opened 2024-07-07
Add documentation to `ShardCloseCode` and `AuditLogEventType`

mplatypus:
### Summary
Added documentation to `ShardCloseCode` and `AuditLogEventType`

### Checklist

### Related issues
Nothing related.


hikari/audit_logs.py
```diff
@@ -293,52 +293,99 @@ class AuditLogEventType(int, enums.Enum):
...
     MESSAGE_BULK_DELETE = 73
+    "Multiple messages were deleted."
     MESSAGE_PIN = 74
+    "Message was pinned to a channel."
     MESSAGE_UNPIN = 75
+    "Message was unpinned from a channel."
     INTEGRATION_CREATE = 80
+    "App was added to server."
     INTEGRATION_UPDATE = 81
+    "App was updated (as an example, its scopes were updated)."
```
davfsa, line 362:
```suggestion
     INTEGRATION_UPDATE = 81
    """An integration was updated.

    For example, its scopes were updated.
    """
```

davfsa (changes requested):
Thanks for the pull request!

Couple of notes:
- Docstrings should begin and end with `"""`, not `"`
- I am not too sure about adding documentation to the shard close codes, but I wouldn't mind it. It seems a bit arbitrary

davfsa:
Also, changelog fragment is missing :)

mplatypus:
@davfsa
> * I am not too sure about adding documentation to the shard close codes, but I wouldn't mind it. It seems a bit arbitrary

What about the 4000-4999? as found here [mdn](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code) it does state that those close codes are private, and therefore unique. I was thinking adding it to all of them just looked nicer imo.

davfsa:
Yeah, that's part of the spec, 4xxx range is for individual use after agreement. It's fine leaving them documented ig.

Would you mind fixing the docstring issue (wrapping with triple quotes instead of single) and I think this is good to merge

mplatypus:
was literally just doing that right now haha.

PR #1969, opened 2024-07-18
Compile development dependencies with uv compile

davfsa:
Waiting on https://github.com/dependabot/dependabot-core/pull/10040 for dependabot support and then this should be good to merge


davfsa:
Closing this for now, will revisit in the future

PR #1991, opened 2024-08-02
Add missing AuditLog event types

syncblaze:
### Summary
I've noticed there are missing event types in the AuditLogEventType class, so i added them

### Checklist


hikari/audit_logs.py
```diff
@@ -327,18 +327,34 @@ class AuditLogEventType(int, enums.Enum):
     INTEGRATION_CREATE = 80
     INTEGRATION_UPDATE = 81
     INTEGRATION_DELETE = 82
+    STAGE_INSTANCE_CREATE = 83
+    STAGE_INSTANCE_UPDATE = 84
+    STAGE_INSTANCE_DELETE = 85
```
tandemdude, line 332: These are addressed in the stage instances PR - which I have just given Dav a patch to fix up

hikari/audit_logs.py
```diff
@@ -327,18 +327,34 @@ class AuditLogEventType(int, enums.Enum):
...
     THREAD_CREATE = 110
     THREAD_UPDATE = 111
     THREAD_DELETE = 112
+    APPLICATION_COMMAND_PERMISSION_UPDATE = 121
+    AUTO_MODERATION_RULE_CREATE = 140
+    AUTO_MODERATION_RULE_UPDATE = 141
+    AUTO_MODERATION_RULE_DELETE = 142
+    AUTO_MODERATION_BLOCK_MESSAGE = 143
+    AUTO_MODERATION_FLAG_TO_CHANNEL = 144
+    AUTO_MODERATION_USER_COMMUNICATION_DISABLED = 145
```
tandemdude, line 348: These are addressed in the automod PR (which I might be looking into redoing soon)

tandemdude (commented):
Not sure about the onboarding and home settings ones and whether they should be addressed in a PR adding actual support for their features instead or not.

syncblaze:
I understand your thinking on adressing them in their own PR but for example im working privatly on a bot which keeps track on specific audit log events, and for that I dont need to have the onboarding or home settings features implemented. I just need the AuditLogEventType. I've already handled these cases diffrently in my project, I just want to give an example on why it might still be a good idea to add them now.

davfsa (approved):

PR #1998, opened 2024-08-05
Document versioning scheme as EffVer

davfsa:
### Summary
After some discussion with multiple members in the Discord channel, EffVer was brought to my attention and it seems to be a way better fit.

Next release will be a stable release `2.0.0` and `EffVer` will be applied from then on

Closes #1997


Tmpod:
For reference: https://jacobtomlinson.dev/effver

PR #2021, opened 2024-08-17
Add None value handling for JSONObject builder

syncblaze:
### Summary
If whe build a json object, the value should not be added to the json if its undefined type, but if its None it should just be added as None. These None checks were performed in the rest functions etc which used the JSON Object builder until now, but it makes far more sense to put this None check into the json builder.

### Checklist


syncblaze:
> Merging this as is could cause some slowdowns because of the double check for `None` that are lying around, would be nice to incorporate that into this pr too

i can do that no problem

PR #2066, opened 2024-09-29
Allow specifying a reason for message deletions

maddymeows:
### Summary

Allow specifying a reason for message deletions.

Has some caveats, namely:

- Audit log entries merging seems to erase the reason at times. 
- Bulk delete reason doesn't seem to work fully from Discord's end, however their docs do say it's supported.

### Checklist

### Related issues

None

### DIFF
--- a/changes/2066.feature.md
+++ b/changes/2066.feature.md
@@ -1,0 +1,1 @@
+Allow specifying a reason for message deletions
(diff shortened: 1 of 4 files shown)


davfsa (approved):
Great PR, thanks for fixing this oversight!

If Discord doesnt properly support the reason in the audit log, then thats something that is probably worth bringing up with them

PR #2079, opened 2024-10-16
Removed ratelimit_per_user field from create_news_channel

syncblaze:
### Summary
Removed ratelimit_per_user field from create_news_channel since news channels dont have a ratelimit field

### Checklist


Le0Developer:
> since news channels dont have a ratelimit field

they do though...

https://github.com/user-attachments/assets/c604432f-e71e-4b39-ada6-7b591bab3f7c

Besides the "janky" way of setting them using the normal client, they work as normal.

syncblaze:
> > since news channels dont have a ratelimit field
> 
> they do though...
> 
> Besides the "janky" way of setting them using the normal client, they work as normal.

Well are u sure the ratelimit affects users if the channel is set as news channel? 

If so, then it would be needed to add the ratelimit field to the GuildNewsChannel.

Le0Developer:
> Well are u sure the ratelimit affects users if the channel is set as news channel?

yep
image

syncblaze:
> > Well are u sure the ratelimit affects users if the channel is set as news channel?
> 
> yep 

then my bad, do you know if the ratelimit field is sent in the news channel payload? 

Ig it should be sent, that means i will add the ratelimit field to the NewsChannel class in the next hours

PR #2196, opened 2025-03-06
Components V2!

mplatypus:
### Summary

Adds support for components V2!

### Checklist


hikari/messages.py
```diff
@@ -172,6 +172,9 @@ class MessageFlag(enums.Flag):
     IS_VOICE_MESSAGE = 1 << 13
     """This message is a voice message."""
 
+    IS_COMPONENTS_V2 = 1 << 15
```
davfsa, line 175: Could we automatically detect and set it when making the request instead of having users set it?

This flag is an unfortunate implementation decision that cannot be removed (as stated by devs), so we could at least remove it for the user

hikari/components.py
```diff
@@ -310,6 +400,197 @@ class TextInputComponent(PartialComponent):
...
+    """Represents a media gallery item."""
+
+    media: MediaResource = attrs.field()
+    """The media for the gallery item."""
+
+    description: typing.Optional[str] = attrs.field()
+    """The description of the gallery item."""
+
+    spoiler: bool = attrs.field()
+    """If the gallery item has a spoiler."""
```
davfsa, line 525:
```suggestion
    """Whether the gallery item is marked as a spoiler."""
```

hikari/components.py
```diff
@@ -310,6 +400,197 @@ class TextInputComponent(PartialComponent):
...
+    """Represents a container component."""
+
+    accent_color: undefined.UndefinedNoneOr[colors.Color] = attrs.field()
+    """The accent colour for the container.
+
+    If undefined, no accent color is set.
+    """
+
+    spoiler: bool = attrs.field()
+    """If the container has a spoiler."""
```
davfsa, line 561:
```suggestion
    """Whether the container is marked as a spoiler."""
```

hikari/components.py
```diff
@@ -310,6 +400,197 @@ class TextInputComponent(PartialComponent):
...
+
+@attrs.define(kw_only=True, weakref_slot=False)
+class ContainerComponent(PartialComponent):
+    """Represents a container component."""
+
+    accent_color: undefined.UndefinedNoneOr[colors.Color] = attrs.field()
+    """The accent colour for the container.
+
+    If undefined, no accent color is set.
+    """
```
davfsa, line 558:
```suggestion
    accent_color: typing.Optional[colors.Color] = attrs.field()
    """The accent colour for the container."""
```

davfsa (changes requested):

davfsa:
`spoiler` fields should also be renamed to `is_spoiler`

hikari/components.py
mplatypus: The accent colour has three states, hidden = omitted (makes the message look like its got a box around it and thats it), default = None (the default colour value and then a colour) so wouldn't the original make more sense?

hikari/components.py
```diff
@@ -310,6 +400,197 @@ class TextInputComponent(PartialComponent):
...
+    """The width of media item."""
+
+    height: undefined.UndefinedNoneOr[int] = attrs.field(repr=True)
+    """The height of the media item."""
+
+    content_type: undefined.UndefinedNoneOr[str] = attrs.field(repr=True)
+    """The content type of the media item."""
+
+    loading_state: undefined.UndefinedNoneOr[MediaLoadingType] = attrs.field(repr=True)
+    """The loading state of the media item."""
```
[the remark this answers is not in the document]
mplatypus: The `proxy_resource` seems fine imo, unless it should be `undefined.UndefinedOr[...]` otherwise, should the rest just be `typing.Optional[...]`?

hikari/messages.py
mplatypus: We can, but there is a few components that are both v1 and v2 supportive, like the action row, and I believe they act different if you use the flag vs not use it. We could make it so any components that **require** the V2 flag, automatically adds it, but I would still leave the option for the end user to add it as well.

davfsa: Yes, that was my idea, force it when a components v2 is used

davfsa: Done this here [`fbb1775` (#2196)](https://github.com/hikari-py/hikari/pull/2196/commits/fbb1775f314cda529fa950a7cccac4818f5f177e)

hikari/impl/rest.py
```diff
@@ -1449,15 +1461,29 @@ def _build_message_payload(  # noqa: C901- Function too complex
...
             if component is not None:
-                serialized_components = [component.build()]
+                component_payload, component_attachments = component.build()
+                serialized_components = [component_payload]
+                final_attachments.extend(component_attachments)
+
+                if component.type in _V2_COMPONENT_TYPES:
+                    if flags is undefined.UNDEFINED:
+                        flags = 0
+                    flags |= messages_.MessageFlag.IS_COMPONENTS_V2
```
davfsa, line 1471: This needs testing

PR #2197, opened 2025-03-06
Fix application command deserialization when context is None

hypergonial:
### Checklist

### DIFF
--- a/hikari/impl/entity_factory.py
+++ b/hikari/impl/entity_factory.py
@@ -2324,7 +2324,7 @@ class EntityFactoryImpl(entity_factory.EntityFactory):
         ]
 
         context_types = [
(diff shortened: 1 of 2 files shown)


hikari/impl/entity_factory.py
```diff
@@ -2324,7 +2324,7 @@ def deserialize_slash_command(
         ]
 
         context_types = [
-            application_models.ApplicationContextType(int(context)) for context in payload.get("contexts", ())
+            application_models.ApplicationContextType(int(context)) for context in (payload.get("contexts", ()) or ())
```
davfsa, line 2327:
```suggestion
            application_models.ApplicationContextType(int(context)) for context in payload.get("contexts") or ()
```

davfsa (changes requested):

PR #2213, opened 2025-03-16
Type tests

mplatypus:
### Summary
Add typing to all tests, along with a pipeline that forces tests to stay typed.

This will make writing new tests easier, and cleaner.


pipelines/pyright.nox.py
```diff
@@ -52,3 +52,10 @@ def verify_types(session: nox.Session) -> None:
     """Verify the "type completeness" of types exported by the library using Pyright."""
     session.install(".", *nox.dev_requirements("pyright"))
     session.run("pyright", "--verifytypes", config.MAIN_PACKAGE, "--ignoreexternal")
+
+
+@nox.session()
+def verify_test_types(session: nox.Session) -> None:
+    """Verify the "type completeness" of the test types using Pyright."""
+    session.install(".", *nox.dev_requirements("pyright"))
+    session.run("pyright", config.TEST_PACKAGE)
```
davfsa, line 61: This isn't technically correct, as we are not verifying the types, but just running pyright on it.

`nox -s pyright_tests` or `nox -s pyright -- --tests-only` are probably more appropriate

tests/hikari/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+
+@pytest.fixture
+def hikari_guild_text_channel() -> channels.GuildTextChannel:
+    return channels.GuildTextChannel(
+        app=mock.Mock(),
+        id=snowflakes.Snowflake(4560),
+        name="guild_text_channel_name",
+        type=channels.ChannelType.GUILD_TEXT,
+        guild_id=mock.Mock(),  # FIXME: Can this be pulled from the actual fixture?
+        parent_id=mock.Mock(),  # FIXME: Can this be pulled from the actual fixture?
```
davfsa, line 32: Doing `.id` would work

tests/hikari/conftest.py
```diff
@@ -0,0 +1,126 @@
+from __future__ import annotations
```
davfsa, line 1: This file seems a bit arbitrary, what's the idea behind it?

pipelines/pyright.nox.py
mplatypus: Can fix that.

tests/hikari/conftest.py
mplatypus: With this, I did have it get the `hikari_partial_guild` object, and then use the id from it, but I kinda changed how I internally used it. This is more a me problem.

tests/hikari/conftest.py
mplatypus: conftest.py is a pytest feature, that allows any fixtures placed within, to be accessable to any tests within that directory (recursively). The goal for this, is to put all the fixtures that are identical, and used in more than one file, here, to minimize the amount of duplicated fixtures. I simply have not gotten around to moving them, its on my list of todo's

mplatypus:
> This pr is going to be insanely hard to review, so I'm mostly going to base it on tests passing + the pipeline passing.
> 
> Any new questionable additions i will comment on as I slowly review this

Yeah, its all good. I know this is going to be a massive PR. I will mark things that I think you need to see with a `FIXME:` with some reason as to why I think its important to view and/or if I have questions about it.

davfsa:
> they only occur sometimes.

Test order is randomized every time nox is ran, it's to catch flaxy tests or tests that behave differently based on what order they are run.

At the top of the pytest output you should see a seed flag with a random number next to it's it you pass it to nox like so: `nox -s -- <flag_here>` then you will be able to reproduce it consistently.

Here I would advice you to look at what order the tests are excuted and the most probable reason is that whatever is running before the first test is modifying the state of whatever is running afterwards.

If you provide me the flag, I can have a look too :)

I can also give a shot at reproducing it later

mplatypus:
I didn't know it did this. That's pretty useful.

Below are some seeds that succeed/fail when ran.

Successful seeds: `4147739088`, `3310411651`, `3105782950`
Failure seeds: `4274378829`, `1168526332`, `620612017`

davfsa:
> I didn't know it did this. That's pretty useful.
> 
> Below are some seeds that succeed/fail when ran.
> 
> Successful seeds: `4147739088`, `3310411651`, `3105782950` Failure seeds: `4274378829`, `1168526332`, `620612017`

Fixed :)

tests/hikari/conftest.py
davfsa: Thats nice to know, didnt know about that feature. Would be nice to document it in the files docstring (as well as add the missing license :P)

davfsa:
Todo: Change all `typing.MutableMapping` and `typing.Mapping` to use `dict` where applicable

pipelines/pyright.nox.py
```diff
@@ -38,6 +38,13 @@ def pyright(session: nox.Session) -> None:
     session.run("pyright")
 
 
+@nox.session()
+def pyright_tests(session: nox.Session) -> None:
+    """Perform type analysis on the tests using Pyright."""
+    nox.sync(session, self=True, extras=["speedups", "server"], groups=["pyright"])
```
davfsa, line 44:
```suggestion
    nox.sync(session, self=True, extras=["speedups", "server"], groups=["pyright", "pytest"])
```

PR #2238, opened 2025-03-20
Add `parent_id` and `lock_permissions` to `reposition_channels`

### LINKED ISSUE #523: Add `lock_permissions` to `reposition_channels`
https://github.com/discord/discord-api-docs/blob/master/docs/resources/Guild.md#modify-guild-channel-positions--patch-guildsguildiddocs_resources_guildguild-objectchannels

This might mean we need to also turn this into a special endpoint

rayakame:
### Summary
- Move `reposition_channels` to `special_endpoints` without breaking/deprecating anything.
- Add `parent_id` and `lock_permissions`for every channel.

### Checklist

### Related issues
closes #523


rayakame:
This should be the first Iteration of this feature.

**Tests and docs are not done yet!**
I want to clear up some things before I start working on these things.

1. **Naming**
I'm not really sure about the names yet. `ChannelRepositioner` doesn't sound right and same for `RepositionChannelHelper` .
But i'm not creative at all and wanted to ask about your guys opinions.
2. **Implementation**
With the current implementation we wouldn't need to deprecate anything and there are no breaking changes. Only problem
being that it could be pretty confusing since you can pass channels into the `rest.reposition_channels` and into the
`ChannelRepositioner`. Normally I wouldn't say that this is important enough for a breaking change or a deprecation but we
already have breaking changes planned for the next release so might as well do that.

Let me know about your opinions and I would be happy to implement them.

hikari/impl/special_endpoints.py
```diff
@@ -204,6 +205,73 @@ async def _keep_typing(self) -> None:
...
+        positions: typing.Mapping[int, snowflakes.SnowflakeishOr[channels.GuildChannel]],
+    ): ...
+
+    def __init__(
+        self,
+        guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
+        request_call: _RequestCallSig,
+        *,
+        positions: typing.Mapping[int, snowflakes.SnowflakeishOr[channels.GuildChannel]] = {},
+    ):
```
davfsa, line 229: Does this need to be a typing overload? Couldnt it just be a single definition?

hikari/api/special_endpoints.py
```diff
@@ -108,6 +124,46 @@ async def __aexit__(
     ) -> None: ...
 
 
+class ChannelRepositioner(abc.ABC):
+    __slots__: typing.Sequence[str] = ()
+
+    @typing.overload
+    @abc.abstractmethod
+    def __init__(self, guild: snowflakes.SnowflakeishOr[guilds.PartialGuild], request_call: _RequestCallSig): ...
```
davfsa, line 132: We should probably copy how GuildBuilder works when it comes to taking in a request call

rayakame:
I will take a look in a hour or so

hikari/impl/special_endpoints.py
rayakame: yeah your prob right the user shouldn't come into contact with that in the end

hikari/api/special_endpoints.py
rayakame: by making it an attrs class?

I mean I could do that but I'm doing some conversion in the constructor and I feel like a constructor is better here idk

rayakame: 👀

davfsa: I mean that it's left as impl detail. We do not define how init is meant to behave in the api.

But tbh, I am not sure what the best way to do it is now :P

rayakame: I changed a lot about the builders, including fixing your concerns here so this conversation can be closed for now.

hikari/api/special_endpoints.py
```diff
@@ -114,6 +114,151 @@ async def __aexit__(
...
+            If you are unauthorized to make the request (invalid/missing token).
+        hikari.errors.ForbiddenError
+            If you cannot access the channel.
+        hikari.errors.NotFoundError
+            If the channel is not found.
+        hikari.errors.RateLimitTooLongError
+            Raised in the event that a rate limit occurs that is
+            longer than `max_rate_limit` when making a request.
+        hikari.errors.InternalServerError
+            If an internal error occurs on Discord while handling the request.
```
rayakame, line 194: @davfsa I would need your opinion here, this would be (in my opinion) the perfect way to document the behaviour of this class, but sadly `__await__` docstrings are not displayed in the docs. What would you do here?

rayakame: How about we do this:

```py
    async def reposition(self) -> None:
        await self
```

this would allow developers to call `.reposition` instead of doing `await obj` while still not breaking the `RESTClient.reposition_channels` function.

You could either do this:
```py
await rest.reposition_channels(guild=... , positions=...)
```
or this:
```py
helper = rest.reposition_channels(guild=...)
helper.add_reposition_channel(position=... , channel=...)
await helper.reposition() # instead of doing `await helper`
```

and we can put all the documentation into the `.reposition()` function for better DX

davfsa: The fact that you can await it should be documented as part of the class description, if you want to make it really explicit, that would probably be the best DX. This is mostly for consistency with all the other special endpoints, and it is not something I want to change.

In the future, I can have a look at mkdocstrings and have it document `__await__` too

davfsa (changes requested):
Only small things. Thanks for working on this!

hikari/api/rest.py
```diff
@@ -5582,23 +5582,19 @@ async def reposition_channels(
         positions
             A mapping of of the object or the ID of an existing channel to
             the new position, relative to their parent category, if any.
+
+            !!! note
+                Instead of using the `positions` parameter, you should make
+                use of the returned [`hikari.api.special_endpoints.ChannelRepositioner`][].
```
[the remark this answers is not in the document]
rayakame: I didn't intend on putting that comment there because we intend on deprecating the old way of doing this, but rather to inform people that you can/should use the `ChannelRepositioner` because otherwise you cannot use the `parent_id` and `lock_permissions`.

I think if we do not talk about it somewhere most users wouldn't even notice that you can use the `ChannelRepositioner`.

davfsa (dismissed):

davfsa (approved):

PR #2241, opened 2025-03-21
Specialize Interaction Events

### LINKED ISSUE #1777: Specialize InteractionCreateEvent
### Summary

Specialize `InteractionCreateEvent` into the different interaction types, similar how `MessageCreateEvent` is specialized for guilds and dms

### Why is this needed?

Nicer DX

### Ideal implementation

Provide more concrete subclasses with correct typing for `interaction`. 

### Checklist

rayakame:
### Summary
As described here #1777 

### Checklist

### Related issues
Closes #1777


rayakame:
```py
import hikari

TOKEN = ""

bot = hikari.GatewayBot(TOKEN, intents=hikari.Intents.ALL)

@bot.listen(hikari.events.MessageCreateEvent)
async def on_message(event: hikari.events.MessageCreateEvent):
    if event.message.content.startswith("!test"):
        component = event.app.rest.build_message_action_row().add_interactive_button(
            hikari.ButtonStyle.PRIMARY,
            "custom_button_id",
            label="test"
        )
        await event.message.respond(
            "Test Button",
            component=component
        )

@bot.listen(hikari.events.InteractionCreateEvent)
async def test_normal_interaction_event(event: hikari.events.InteractionCreateEvent):
    print(event)

@bot.listen(hikari.events.ComponentInteractionCreateEvent)
async def test_component_interaction_event(event: hikari.events.ComponentInteractionCreateEvent):
    print(event)

if __name__ == "__main__":
    bot.run()
``` 

little test snippet

hikari/impl/event_factory.py
```diff
@@ -493,9 +496,16 @@ def deserialize_audit_log_entry_create_event(
...
+        interaction = self._app.entity_factory.deserialize_interaction(payload)
+        if isinstance(interaction, command_interactions.CommandInteraction):
+            return interaction_events.CommandInteractionCreateEvent(shard=shard, interaction=interaction)
+        if isinstance(interaction, command_interactions.AutocompleteInteraction):
+            return interaction_events.AutocompleteInteractionCreateEvent(shard=shard, interaction=interaction)
+        if isinstance(interaction, component_interactions.ComponentInteraction):
+            return interaction_events.ComponentInteractionCreateEvent(shard=shard, interaction=interaction)
+        if isinstance(interaction, modal_interactions.ModalInteraction):
+            return interaction_events.ModalInteractionCreateEvent(shard=shard, interaction=interaction)
+        return interaction_events.InteractionCreateEvent(shard=shard, interaction=interaction)
```
davfsa, line 508: A map would be wayy better here

hikari/impl/event_manager.py
```diff
@@ -827,7 +827,14 @@ async def on_webhooks_update(self, shard: gateway_shard.GatewayShard, payload: d
...
 
-    @event_manager_base.filtered(interaction_events.InteractionCreateEvent)
+    @event_manager_base.filtered(
+        (
+            interaction_events.ComponentInteractionCreateEvent,
+            interaction_events.ModalInteractionCreateEvent,
+            interaction_events.AutocompleteInteractionCreateEvent,
+            interaction_events.CommandInteractionCreateEvent,
+        )
+    )
```
davfsa, line 837: This could be kept as `@event_manager_base.filtered(interaction_events.InteractionCreateEvent)`. Since they are subclasses of it, this consumer will be enabled as long as one of them is

changes/2241.feature.md
```diff
@@ -0,0 +1 @@
+Added more specialized interaction events to improve developer experience
```
davfsa, line 1: Would be nice to list them here too :)

davfsa (changes requested):

hikari/impl/event_manager.py
rayakame: oh okay

hikari/interactions/base_interactions.py
```diff
@@ -212,7 +212,7 @@ class PartialInteraction(snowflakes.Unique, webhooks.ExecutableWebhook):
     application_id: snowflakes.Snowflake = attrs.field(eq=False, repr=False)
     """ID of the application this interaction belongs to."""
 
-    type: typing.Union[InteractionType, int] = attrs.field(eq=False, repr=True)
+    type: InteractionType = attrs.field(eq=False, repr=True)
```
rayakame, line 215: afaik these Unions are not needed and just add unnecessary overhead because for example here:
`self._interaction_events_map.get(interaction.type)` you would need to convert it to an InteractionType because it could be an int even though when deserializing it does this: https://github.com/hikari-py/hikari/blob/b3c356a1930568357ab4e7d6699959ca670cca09/hikari/impl/entity_factory.py#L2741

davfsa: I have wanted to do that for a while (do not deserialize unknown types).

I will get to it in a sec!

rayakame: no need to stress mypy complained that it could be an int and I didn't feel like converting it again so I just changed it in the class

hikari/impl/event_factory.py
```diff
@@ -490,12 +491,23 @@ def deserialize_audit_log_entry_create_event(
...
         self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject
     ) -> interaction_events.InteractionCreateEvent:
-        return interaction_events.InteractionCreateEvent(
-            shard=shard, interaction=self._app.entity_factory.deserialize_interaction(payload)
-        )
+        interaction = self._app.entity_factory.deserialize_interaction(payload)
+
+        if event := self._interaction_events_map.get(interaction.type):
+            return event(shard=shard, interaction=interaction)
+        return interaction_events.InteractionCreateEvent(shard=shard, interaction=interaction)
```
davfsa, line 510: We should never have to return a plain `InteractionCreateEvent`, we should error if we do not know what specific one to return

hikari/impl/event_factory.py
```diff
@@ -490,12 +491,23 @@ def deserialize_audit_log_entry_create_event(
...
     ######################
 
+    _interaction_events_map: dict[
+        base_interactions.InteractionType, typing.Type[interaction_events.InteractionCreateEvent]
+    ] = {
+        base_interactions.InteractionType.APPLICATION_COMMAND: interaction_events.CommandInteractionCreateEvent,
+        base_interactions.InteractionType.AUTOCOMPLETE: interaction_events.AutocompleteInteractionCreateEvent,
+        base_interactions.InteractionType.MESSAGE_COMPONENT: interaction_events.ComponentInteractionCreateEvent,
+        base_interactions.InteractionType.MODAL_SUBMIT: interaction_events.ModalInteractionCreateEvent,
+    }
```
davfsa, line 501: This should probably not be bound to the class, but rather outside of it (IMO)

davfsa (changes requested):

hikari/impl/event_factory.py
rayakame: Which error do you want to raise?

hikari/impl/event_factory.py
rayakame: in my opinion it would be better having this inside of the class. 

These mappings are also inside of the class in other parts of Hikari (for example entity factory) and its an implementation detail so its not really needed outside of the implementation class.

davfsa: I am saying not bound it to the class, because then each instance will also have the dict, which could easily be global.

Its just a nitpick tho

hikari/impl/event_factory.py
```diff
@@ -490,12 +501,23 @@ def deserialize_audit_log_entry_create_event(
...
         self, shard: gateway_shard.GatewayShard, payload: data_binding.JSONObject
     ) -> interaction_events.InteractionCreateEvent:
-        return interaction_events.InteractionCreateEvent(
-            shard=shard, interaction=self._app.entity_factory.deserialize_interaction(payload)
-        )
+        interaction = self._app.entity_factory.deserialize_interaction(payload)
+
+        if event := self._interaction_events_map.get(interaction.type):
+            return event(shard=shard, interaction=interaction)
+        raise errors.UnrecognisedEntityError(f"Unrecognised interaction type {interaction.type}")
```
davfsa, line 520:
```suggestion
        event = _interaction_events_map[interaction.type]
        return self._interaction_events_map[interaction.type](shard=shard, interaction=interaction)
```

If we ever get here and we do not deserialize it properly, its because we didnt update `_interaction_events_map` (and, as such, a keyerror), not because the entity was not recognised

davfsa (changes requested):

davfsa (approved):

davfsa:
Im going to go ahead and merge this. Thanks for your work!

PR #2281, opened 2025-04-01
Upgrade typing to use `X | Y` notation

davfsa:
### Summary
Forgot to do this as part of the ruff update (was already considering it then), but I do not think we need runtime typechecking and we are already suited to run without it, so this should be fine


davfsa:
Will open a follow up PR to address the audit issue

PR #2318, opened 2025-04-19
Add missing tests for components V2

mplatypus:
### Summary
Added the missing tests for the code that automatically adds the `IS_COMPONENTS_V2` flag to a message payload if the user adds a V2 component.

### Checklist

### DIFF
--- a/tests/hikari/impl/test_rest.py
+++ b/tests/hikari/impl/test_rest.py
@@ -1642,6 +1642,18 @@ class TestRESTClientImpl:
 
         assert payload.get("flags") is message_models.MessageFlag.IS_COMPONENTS_V2
 
+    def test__build_message_payload_with_components_v2_and_flags(self, rest_client):
+        component_1 = mock.Mock(type=components.ComponentType.ACTION_ROW, build=mock.Mock(return_value=({}, ())))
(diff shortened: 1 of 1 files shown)


tests/hikari/impl/test_rest.py
```diff
@@ -1557,6 +1558,90 @@ def test__build_message_payload_when_both_single_and_plural_args_passed(
...
+        assert payload.get("flags") is message_models.MessageFlag.IS_COMPONENTS_V2
+
+    def test__build_message_payload_with_mixed_components(self, rest_client):
+        component_1 = mock.Mock(type=components.ComponentType.ACTION_ROW, build=mock.Mock(return_value=({}, ())))
+        component_2 = mock.Mock(type=components.ComponentType.CONTAINER, build=mock.Mock(return_value=({}, ())))
+
+        payload, _ = rest_client._build_message_payload(components=[component_1, component_2])
+
+        assert payload.get("flags") is message_models.MessageFlag.IS_COMPONENTS_V2
+
```
davfsa, line 1644: Would be nice to also have an extra test to ensure that if the user passes flags, we do not override them, and instead just build on top of them

davfsa (approved):

PR #2338, opened 2025-04-29
Add AWEBP support, increase sticker asset availability and deprecate `x_url` properties

EtorixDev:
### Summary
- Switched stickers to always use the `MEDIA_PROXY_URL` (except for `LOTTIE`) to access more features (sizing, static `PNG` for `APNG`).
- Added `AWEBP` and `APNG` formats to `routes`.
  - `APNG` added for distinction from static `PNG` stickers.
- Marked shortcut `url` properties for deprecation with messages pointing users to the `make_url` functions.
- Updated all of the `make_url` functions.
  - Uses `image_format` instead of `ext` for accuracy with new formats not technically being extensions.
  - Uses `typing.Literal` to spell out exactly which formats are allowed per asset type.
  - Added `lossless` argument to query for lossless or compressed `WEBP`/`AWEBP`. Ignored for all other formats, as mentioned in the docstrings.
  - `ext` marked as deprecated and throws warning.
- Updated `CDNRoute.compile`.
  - Accepts a `setting` dictionary for new type-specific query parameters.
    - Accepts a `passthrough` key to return a static `PNG` for an `APNG` sticker. Ignored for all other formats.
    - Accepts an `animated` key to return an `AWEBP` for any `GIF`-accepting endpoints. Ignored for all other formats.
    - Accepts a `lossless` key to return a lossless or compressed `WEBP`/`AWEBP`. Ignored for all other formats.
  - No longer throws errors for non-sizeable endpoints.
    - The only endpoints previously that would error for this were Default Avatars and Stickers. Stickers actually can be resized when accessing the `MEDIA_PROXY_URL`. The only exception to this is `APNG` and `LOTTIE`. Accessing `APNG` or `LOTTIE` stickers, or Default Avatars, in the browser with a size query parameter still works, it's simply silently ignored as stated in the [docs](<https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints>). So instead of raising when providing a size when unsupported, the Default Avatar functions simply don't accept a size parameter and for the stickers, it's silently set to `None`. This is documented in the docstrings for the sticker url function.
- Updated a lot of docstrings, tests, and removed and added tests where applicable.

### Checklist


hikari/applications.py
```diff
@@ -467,6 +477,28 @@ def __hash__(self) -> int:
     def __eq__(self, other: object) -> bool:
         return self.user == other
 
+    @typing_extensions.override
+    def make_avatar_url(
+        self,
+        *,
+        image_format: typing.Literal["PNG", "JPEG", "JPG", "WEBP", "AWEBP", "GIF"] | None = None,
+        size: int | None = 4096,
+        lossless: bool | None = True,
```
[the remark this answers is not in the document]
EtorixDev: The client defaults to `True` for lossless on `WEBP` images so this is consistent with that. `WEBP` is also available for every single endpoint so it needs to be included for every method. It's ignored during calling `compile_to_file` if it's not applicable to the chosen format.

hikari/guilds.py
```diff
@@ -695,75 +702,134 @@ def global_name(self) -> str | None:
...
+            If `size` is specified but is not a power of two or not between 16 and 4096.
         """
         if self.guild_avatar_hash is None:
             return None
 
-        if ext is None:
-            ext = "gif" if self.guild_avatar_hash.startswith("a_") else "png"
+        if ext:
+            deprecation.warn_deprecated("ext", removal_version="2.4.0", additional_info="Use 'image_format' instead.")
+            image_format = ext.upper()  # type: ignore  # noqa: PGH003
```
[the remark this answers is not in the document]
EtorixDev: mypy/pylance seems to think `image_format` which is a literal can't be set to `ext` which is a string. It produces many `"str" is not assignable to...` errors. Since `ext` is marked for deprecation I figured this approach was fine.

hikari/guilds.py
```diff
@@ -772,22 +838,33 @@
...
             urls.CDN_URL,
             guild_id=self.guild_id,
             user_id=self.id,
             hash=self.guild_banner_hash,
             size=size,
-            file_format=ext,
+            file_format=image_format,
+            settings={
+                "animated": True if image_format == "AWEBP" else None,
+                "lossless": lossless if image_format in ("WEBP", "AWEBP") else None,
```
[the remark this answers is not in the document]
EtorixDev: It is set in 4 other files. Anywhere that `AWEBP` is possible it is set. Also that isn't line 1866.

hikari/users.py
```diff
@@ -663,84 +703,146 @@ def username(self) -> str:
...
 
         return routes.CDN_USER_AVATAR.compile_to_file(
-            urls.CDN_URL, user_id=self.id, hash=self.avatar_hash, size=size, file_format=ext
+            urls.CDN_URL,
+            user_id=self.id,
+            hash=self.avatar_hash,
+            size=size,
+            file_format=image_format,
+            settings={
+                "animated": True if image_format == "AWEBP" else None,
```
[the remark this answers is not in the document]
EtorixDev: It's actually not handled differently at all.

tests/hikari/internal/test_routes.py
```diff
@@ -124,109 +124,147 @@ def test_eq_operator__considers_path_template_only(self):
...
+        route = routes.CDNRoute("/foo/bar", {"WEBP", "AWEBP", "GIF"})
+        compiled_url = route.compile("http://example.com", file_format="AWEBP", hash="a_boooob")
+        assert "animated=" not in compiled_url, f"compiled_url={compiled_url}"
+
+    @pytest.mark.parametrize(("file_format", "size"), [("APNG", 128), ("LOTTIE", 128)])
+    def test_size_excluded_for_apng_and_lottie(self, file_format, size):
+        route = routes.CDNRoute("/foo/bar", {"APNG", "LOTTIE", "PNG", "JPG"})
+        compiled_url = route.compile("http://example.com", file_format=file_format, size=size)
+        assert f"size=" not in compiled_url, f"compiled_url={compiled_url}"
+
```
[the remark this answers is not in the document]
EtorixDev: Modified the `GIF` tests to also check `AWEBP` and `APNG`.

changes/2338.feature.md
```diff
@@ -0,0 +1 @@
+Add AWEBP support for asset URLs, increase sticker asset availability, & prepare various properties for deprecation.
```
davfsa, line 1: The deprecations should be listed in a different changelog fragment called `2338.deprecations.md`

hikari/internal/routes.py
```diff
@@ -214,10 +214,15 @@ def _(self, _: attrs.Attribute[typing.AbstractSet[str]], values: typing.Abstract
...
-    """Whether a `size` param can be specified."""
-
-    def compile(self, base_url: str, *, file_format: str, size: int | None = None, **kwargs: object) -> str:
+    def compile(
+        self,
+        base_url: str,
+        *,
+        file_format: str,
+        size: int | None = None,
+        settings: dict[str, str | bool | None] | None = None,
```
davfsa, line 223:
```suggestion
        size: int = None,
        lossless: undefined.UndefinedOr[bool] = undefined.UNDEFINED,
```

+ all the other ones

hikari/internal/routes.py
```diff
@@ -241,53 +248,68 @@ def compile(self, base_url: str, *, file_format: str, size: int | None = None, *
...
 
         if file_format not in self.valid_formats:
             raise TypeError(
                 f"{file_format} is not a valid format for this asset. Valid formats are: "
                 + ", ".join(self.valid_formats)
             )
 
-        if "hash" in kwargs and not str(kwargs["hash"]).startswith("a_") and file_format == GIF:
-            msg = "This asset is not animated, so cannot be retrieved as a GIF"
+        if "hash" in kwargs and not str(kwargs["hash"]).startswith("a_") and file_format in [AWEBP, GIF, APNG]:
```
davfsa, line 264:
```suggestion
        if "hash" in kwargs and not str(kwargs["hash"]).startswith("a_") and file_format in {AWEBP, GIF, APNG}:
```

In general, we want to use sets for `in` checks

hikari/internal/routes.py
```diff
@@ -241,53 +248,68 @@ def compile(self, base_url: str, *, file_format: str, size: int | None = None, *
...
+        query_params: dict[str, str] = {}
 
-        if size is not None:
-            if not self.is_sizable:
-                msg = "This asset cannot be resized."
-                raise TypeError(msg)
+        if file_format in ("APNG", "LOTTIE") and size:
+            size = None
 
+        if size is not None:
```
davfsa, line 274: This code is quite confusing.

I prefer the old way of checking if it should be resizable and erroring if it isnt.

Need to have a better look at this, but this is definitely not nice to read

hikari/internal/routes.py
```diff
@@ -241,53 +248,68 @@ def compile(self, base_url: str, *, file_format: str, size: int | None = None, *
...
+
+        if settings is not None:
+            if (passthrough := settings.get("passthrough")) is not None:
+                query_params["passthrough"] = str(passthrough).lower()
+            if (animated := settings.get("animated")) is not None:
+                query_params["animated"] = str(animated).lower()
+            if (lossless := settings.get("lossless")) is not None:
+                query_params["lossless"] = str(lossless).lower()
+
+        kwargs = {k: urllib.parse.quote(str(v)) for k, v in kwargs.items()}
```
davfsa, line 294: Comment got lost here

hikari/internal/routes.py
```diff
@@ -241,53 +248,68 @@ def compile(self, base_url: str, *, file_format: str, size: int | None = None, *
...
-            else:
+            if not (size_power.is_integer() and 4 <= size_power <= 12):
                 msg = "size must be an integer power of 2 between 16 and 4096 inclusive"
                 raise ValueError(msg)
 
+            query_params["size"] = str(size)
+
+        if settings is not None:
+            if (passthrough := settings.get("passthrough")) is not None:
+                query_params["passthrough"] = str(passthrough).lower()
```
davfsa, line 288: Use `data_binding.QueryBuilder` for this

hikari/internal/routes.py
```diff
@@ -597,51 +619,69 @@ def compile_to_file(
...
+CDN_FORMAT_EXT: typing.Final[dict[str, str]] = {
+    PNG: "png",
+    JPEG: "jpeg",
+    JPG: "jpg",
+    WEBP: "webp",
+    APNG: "png",
+    AWEBP: "webp",
+    GIF: "gif",
+    LOTTIE: "json",
+}
```
davfsa, line 640: I am quite meh about this.

I know this is kinda going a bit back to the beginning, but would it be better to keep it mostly how it was (with the `ext` parameter) and for AWEBP support, just allow passing `animated=True` and then the webp returned will be animated.

Mostly thinking this because the complexity added by this pull request, specially when handling and keeping things in mind is quite confusing, and I believe could be more straight forward.

davfsa (changes requested):

hikari/internal/routes.py
```diff
@@ -214,10 +214,15 @@ def _(self, _: attrs.Attribute[typing.AbstractSet[str]], values: typing.Abstract
...
-
-    def compile(self, base_url: str, *, file_format: str, size: int | None = None, **kwargs: object) -> str:
+    def compile(
+        self,
+        base_url: str,
+        *,
+        file_format: str,
+        size: int | None = None,
+        settings: dict[str, str | bool | None] | None = None,
+        **kwargs: object,
```
davfsa, line 224: Note to self: Revisit the usage of kwargs here. It will probably be better to just specify whatever we need as an argument

davfsa (approved):

PR #2348, opened 2025-05-03
rewrite hikari.api.cache

novanai:
### Summary
An attempt at standardising the format of all docstrings

### Checklist

- [ ] ~~I have made unittests according to the code I have added/modified/deleted.~~

### Related issues
N/A


hikari/api/cache.py
```diff
@@ -497,72 +483,71 @@ def get_member(
...
         -------
         CacheView[hikari.snowflakes.Snowflake, CacheView[hikari.snowflakes.Snowflake, hikari.guilds.Member]]
-            A view of guild IDs to views of user IDs to objects of the members
-            that were found from the cache.
+            A view of guild IDs to views of user IDs to member objects.
         """
 
     @abc.abstractmethod
     def get_members_view_for_guild(
         self, guild_id: snowflakes.Snowflakeish, /
```
novanai, line 513: Why does this use `guild_id` instead of `guild`?

novanai (commented):
Why are some cache methods called `set_` instead of `add_`?

hikari/api/cache.py
rayakame: because it only takes `Snowflakeish`  not `SnowflakeishOr` 

Should be changed to `guild: snowflakes.SnowflakeishOr[guilds.PartialGuild]` imo

novanai: That's what I was thinking

PR #2392, opened 2025-05-30
Properly close http response

davfsa:
This didnt become apparent until you were running massive workloads.

The response would eventually get cleaned up and the connection closed, but this should do it more eagerly


rayakame:
I hope you dont mind, since this had "allow edits from maintainers" on I pushed a fix for the failing tests directly to this branch

PR #2409, opened 2025-06-08
Add `GUILD_MEDIA` channel type

### LINKED ISSUE #2407: Support for channel type 16 (media channels)
### Summary

Add channel type 16 to supported channels so hikari can access them and fetch messages.
I guess this is pretty low priority, but with how similar media channels are to forum channels i would like to know why it’s not been implemented in the three years media channels already exist

### Why is this needed?

i have a server with media channels for gallery/clips and would like to be able to check for reactions/messages in these channels without having to rely on 100% uptime to catch events

### Ideal implementation

if i’m understanding this right media channels are just fancy forum channels, so the implementation should be pretty straightforward

### Checklist

- [x] I have searched the issue tracker and have made sure it's not a duplicate. If it is a follow up of another issue, I have specified it.

Forbidden-A:
### Checklist

### Related issues

Close #2407


hikari/impl/rest.py
```diff
@@ -3048,6 +3048,51 @@ async def create_guild_forum_channel(
...
+        rate_limit_per_user: undefined.UndefinedOr[time.Intervalish] = undefined.UNDEFINED,
+        default_auto_archive_duration: undefined.UndefinedOr[time.Intervalish] = undefined.UNDEFINED,
+        default_thread_rate_limit_per_user: undefined.UndefinedOr[time.Intervalish] = undefined.UNDEFINED,
+        default_forum_layout: undefined.UndefinedOr[channels_.ForumLayoutType | int] = undefined.UNDEFINED,
+        default_sort_order: undefined.UndefinedOr[channels_.ForumSortOrderType | int] = undefined.UNDEFINED,
+        available_tags: undefined.UndefinedOr[typing.Sequence[channels_.ForumTag]] = undefined.UNDEFINED,
+        default_reaction_emoji: str
+        | emojis.Emoji
+        | undefined.UndefinedType
+        | snowflakes.Snowflake = undefined.UNDEFINED,
```
davfsa, line 3073:
```suggestion
        default_reaction_emoji: typing.UndefinedOr[str | emojis.Emoji | snowflakes.Snowflake] = undefined.UNDEFINED,
```

PR #2427, opened 2025-06-19
Switch ratelimiting logic to a sliding window

davfsa:
This started rollout on 2022-11-02 @ 14:30 PT, but was never announced anywhere. It allows us to greatly improve throughput over time


davfsa:
This has been deployed at a large scale and it seems to function great!

It greatly improves throughput from 57 req/s to 67 req/s on a huge load. I call this a massive win!

PR #2501, opened 2025-08-22
New pin endpoints

mplatypus:
### Summary
Add support for new pin endpoints so you can actually fetch the 250 pins.

### Checklist


hikari/impl/special_endpoints.py
```diff
@@ -395,6 +395,51 @@ async def _next_chunk(self) -> typing.Generator[messages.Message, typing.Any, No
...
+        if response["has_more"] is False:
+            self._has_more = False
+
+        chunk = response["items"]
+
+        if chunk == []:
+            return None
+
+        self._first_id = chunk[-1]["pinned_at"]
+        return (self._entity_factory.deserialize_pinned_message(m) for m in chunk)
```
davfsa, line 440:
```suggestion
        if not self._has_more:
            return None

        query = data_binding.StringMapBuilder()
        query.put("before", self._first_id)
        query.put("limit", 50)

        response = await self._request_call(compiled_route=self._route, query=query)
        assert isinstance(response, dict)

        chunk = response["items"]
		if not chunk:
			self._has_more = False
			return None

        self._has_more = response["has_more"]

        self._first_id = chunk[-1]["pinned_at"]
        return (self._entity_factory.deserialize_pinned_message(m) for m in chunk)
```

PR #2575, opened 2025-10-17
Add soundboard support.

mplatypus:
### Summary
Added the soundboard sound object and all events/endpoints.

### Checklist


hikari/impl/entity_factory.py
```diff
@@ -4711,3 +4712,39 @@ def deserialize_auto_mod_rule(self, payload: data_binding.JSONObject) -> auto_mo
...
+                is_animated=False,  # FIXME: I am unsure if this is correct, but I have no clue on how I could validate this. # noqa: TD001, E501
+            )
+        if emoji_id and not emoji_name:
+            emoji = emoji_models.CustomEmoji(
+                id=emoji_id,
+                name="",  # FIXME: I am unsure if this is correct, but I have no clue on how I could validate this. # noqa: TD001, E501
+                is_animated=False,  # FIXME: I am unsure if this is correct, but I have no clue on how I could validate this. # noqa: TD001, E501
+            )
+        if not emoji_id and emoji_name:
+            emoji = emoji_models.UnicodeEmoji(emoji_name)
```
davfsa: Given we are not given `animated`, then lets always set it as False (Resolves the FIXMEs on it)

There will never be the case of `emoji_id and not emoji_name`, so that branch can be removed

Issue #2746, opened 2026-08-02
Missing Features & Fixes


rayakame:
I will try my best to collect a complete list of things that are missing or need fixing. I found a lot of stuff so i figured it would be best to keep it in a single issue. I will implement all of this, and link every pr that fixes stuff here, but everybody can help here.
## Missing:
- [ ] Group DM (PR: #2773):
  - https://discord.com/developers/docs/resources/user#create-group-dm
  - https://discord.com/developers/docs/resources/channel#group-dm-add-recipient
  - https://discord.com/developers/docs/resources/channel#group-dm-remove-recipient
- [ ] Subscriptions (PR: #2767):
  - https://docs.discord.com/developers/resources/subscription#list-sku-subscriptions
  - https://docs.discord.com/developers/resources/subscription#get-sku-subscription
  - https://discord.com/developers/docs/events/gateway-events#subscription-create
  - https://discord.com/developers/docs/events/gateway-events#subscription-update
  - https://discord.com/developers/docs/events/gateway-events#subscription-delete
- [ ] Voice Channel Status (PR: #2774):
  - https://docs.discord.com/developers/resources/channel#set-voice-channel-status
  - https://docs.discord.com/developers/events/gateway-events#voice-channel-status-update
  - https://docs.discord.com/developers/topics/permissions#permissions-bitwise-permission-flags:~:text=SET%5FVOICE%5FCHANNEL%5FSTATUS
  - https://docs.discord.com/developers/resources/audit-log#audit-log-events:~:text=VOICE%5FCHANNEL%5FSTATUS%5FCREATE
- [ ] Invite Target Users (PR: #2791):
  - https://discord.com/developers/docs/resources/invite#get-target-users
  - https://discord.com/developers/docs/resources/invite#update-target-users
  - https://discord.com/developers/docs/resources/invite#get-target-users-job-status
- [ ] https://discord.com/developers/docs/resources/message#search-guild-messages (PR: #2778)
- [ ] Soundboard (PR: #2575):
  - https://discord.com/developers/docs/resources/soundboard 
- [ ] Guild Widget (Partial PR: #2571):
  - https://discord.com/developers/docs/resources/guild#get-guild-widget
  - https://discord.com/developers/docs/resources/guild#get-guild-widget-image
- [x] `fetch_activity_instance` (PR: #2766)
- [x] Entitlements (PR: #2750)
- [x] Channel Info (PR: #2752)
[...]

PR #2768, opened 2026-08-05
True-up the invites module

### LINKED ISSUE #2226: True-up the `invites` module
Noticed a few discrepancies while working on #2212 / #2224 -

- `invites.Invite` is missing an enum field for [all the possible invite types](https://discord.com/developers/docs/resources/invite#invite-object-invite-types).
- While not officially documented other than the 'Friend' invite type, invites that point directly to a user are supported.
  - This raises the issue of invites not having either a guild(_id) or a channel(_id) associated with them, as the Invite class hints that `guild`, `guild_id`, and `channel` can be None but not `channel_id`. The `channel` field in an [Invite response object is described as being nullable](https://discord.com/developers/docs/resources/invite#invite-object-invite-structure)
  - These friend-specific invites look like this (note the `null` channel and `type: 2`)
    ```
    {
      "code": "<invitecode>",
      "inviter": { < inviter user object > },
      "channel": null,
      "type": 2,
      "uses": 0,
      "max_uses": 5,
      "max_age": 604800,
      "created_at": "2025-03-19T11:40:31.885200+00:00",
      "expires_at": "2025-03-26T11:40:31+00:00"
    }
    ```
    Image
  - The `max_uses` is always set to 5, and `max_age` is always 7 days/`expires_at` is always `created_at + 7 days`
- `guild_scheduled_event(_id)` is also listed as a possible (nullable) field on an Invite object, but it's missing from `invites.Invite` and should be added

rayakame:
### Summary
Trues up the invites module against the current invite docs (part of #2746, closes #2226).

- Added the `type`, `flags`, `roles`, `role_ids` and `guild_scheduled_event` fields to `Invite`, plus the `InviteType`/`InviteFlags` enums and the `InviteRole` model
- `channel`/`channel_id` are now nullable. Friend invites send `"channel": null` without a `channel_id`, so deserializing them used to blow up with a `KeyError`
- Added the missing `scheduled_event` parameter to `RESTClient.fetch_invite`

### Checklist

### Related issues
#2226


rayakame:
The `KeyError` for friend invites only occurs for `RESTApp`'s authenticated with a bearer token or no token.

Changed priority to `P1` because of this

changes/2768.feature.md
```diff
@@ -0,0 +1 @@
+True-up the invites module: add the `Invite.type` field with the new `InviteType` enum, the `Invite.flags` field with the new `InviteFlags` enum, the `Invite.guild_scheduled_event` field, the `Invite.roles` field with the new `InviteRole` model, the `Invite.role_ids` field, and the missing `scheduled_event` parameter to `RESTClient.fetch_invite`
```
davfsa: Would appreciate a more list like entry in the changelog

davfsa (approved):

Issue #1992, opened 2024-08-03
field "type" of ChannelOverwriteEntryInfo does not get deserialized properly


syncblaze:
### Steps to reproduce

1. Get an AuditLogEntry where the action_type is one of these: `CHANNEL_OVERWRITE_CREATE | CHANNEL_OVERWRITE_UPDATE | CHANNEL_OVERWRITE_DELETE`
2. `audit_log_entry.options.type` is of type `str `instead of `PermissionOverwriteType`
3. This is because discord returns this for some reason as a string instead of beeing consistend. [Look here](https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object-optional-audit-entry-info:~:text=string-,Type%20of%20overwritten%20entity%20%2D%20role%20(,),-CHANNEL_OVERWRITE_CREATE%20%26%20CHANNEL_OVERWRITE_UPDATE%20%26%20CHANNEL_OVERWRITE_DELETE)
4. Normal overwrite object uses int [as shown here](https://discord.com/developers/docs/resources/channel#overwrite-object)

### Expected result

If the bug wasn't there there should be a `PermissionOverwriteType` object instead of a `str` in the `type` field.

### Actual result

There is a `str`.

### System info

```shell
hikari (2.0.0.dev126) [f65737f4]
located at C:\Users\Server\PycharmProjects\Pyro\venv\Lib\site-packages\hikari
CPython 3.12.4 MSC v.1940 64 bit (AMD64)
Windows DESKTOP-UPP6BLS 11 10.0.22631 AMD64 AMD64 Family 23 Model 96 Stepping 1, AuthenticAMD
```

### Checklist

- [X] I have made sure to remove ANY sensitive information (bot token, passwords, credentials, personal details, etc.).
- [X] I have searched the issue tracker and have made sure it's not a duplicate. If it is a follow up of another issue, I have specified it.

Issue #1779, opened 2023-12-07
Consider using msgpec


davfsa:
### Summary

An idea I have had for a while, msgspec performs [insanely better](https://jcristharif.com/msgspec/benchmarks.html) than anything else out there, and it would allow us to simplify models code by an insane magnitude, as well as being able to finally freeze models without a performance penalty, leading to the removal of a lot of code in cache to ensure immutability.

Unfortunately, this would not come easily, as it could potentially be a pretty big breaking change, as we would only be able to provide fields that Discord themselves provide (we will ofc need converter functions, similar to what `EntityFactory` and `EventFactory` do, but they would be a lot simpler and for specific cases, since the rest will be handled by msgspec). This might mean the removal of `.app` and `.shard`, possibly passing them to the end developer some other way.

Additionally, some nice thing we might be able to implement is automatic checking of our implementation vs the discord spec to detect missing fields and new fields that we need to implement.

Some work on the ports that I have been doing: https://github.com/davfsa/hikari/tree/feature/msgspec

Links to:
https://github.com/hikari-py/hikari/discussions/1191
https://github.com/hikari-py/hikari/discussions/1110

### Why is this needed?

.

### Checklist

- [X] I have searched the issue tracker and have made sure it's not a duplicate. If it is a follow up of another issue, I have specified it.

mikeshardmind:
First off, I'm not a user of hikari I came across this as part of a discussion with someone else about more frameworks using type safe decoding, I have no stakes in this one way or another. Just providing a bit of extra info that might be relevant to any decision making process here.

msgspec is pretty tightly tied to CPython's internals, see: https://github.com/jcrist/msgspec/issues/22

This isn't necessarily bad, and as noted in the issue, several other libraries to accelerate CPython code are also non-portable across implementations. Going with msgspec, unless you keep around a non-msgspec fallback path will break existing pypy or graalpy compatibility (All of your current requirements and hikari itself appear to be compatible with both). While that could be a reasonable choice, I'm unsure of your user-base's expectations here for compatibility with other Python implementations.

davfsa:
Thanks a lot for your input @mikeshardmind. It is insanely welcome!!

[...]

PR #2770, opened 2026-08-05
Make enums and flags strict when deserializing

### LINKED ISSUE #1994: Make enum/flags strict when deserializing
### Summary

Remove the following lines to error on failure instead of returning what was cast in

Involves changing

https://github.com/hikari-py/hikari/blob/e13e29a796eafbfdd2cd6b4747f9fb98000d25c6/hikari/internal/enums.py#L148-L150

and `Flag` to ignore unknown bits, as well as making sure that the new errors are properly handled and/or ignored

### Why is this needed?

Would provide a nicer interface for developers and make it easier to catch deserialization errors (like #1992), as currently we just return what was cast in when we are unable to deserialize it properly

### Ideal implementation

`typing.Union` will need to be removed from types that allow an enum/flag and switch to only allowing the enum type.

### Checklist

rayakame:
### Summary
This implements #1994: casting an unknown value to one of our enums now creates an "unknown member" of the enum type instead of returning the raw value back, matching what `Flag` already did for unknown bits. Unknown members keep hold of the raw value, so they compare, hash and serialize like it, and get a generated `UNKNOWN <value>` name. I also added an `is_unknown` property to both `Enum` and `Flag` so these can be detected without poking at internals.

Since enums are now strict on both ends, I removed the `X | int`/`X | str` unions everywhere: from the read-path model fields as well as from the write-path REST parameters and builders, including the `typing.Literal` aliases that mixed enum members with raw ints. Passing a value of the wrong type (e.g. a `str` to an int based enum) raises `TypeError` now, but unknown values of the correct type never error. Deserialization paths where Discord is known to send loosely typed values (audit log change values, overwrite types and some nullable fields) got explicit guards so nothing breaks at runtime.

### Checklist

### Related issues
Close #1994


hikari/impl/entity_factory.py
```diff
@@ -483,10 +483,14 @@ def __init__(self, app: traits.RESTAware) -> None:
...
-            audit_log_models.AuditLogChangeKey.EXPLICIT_CONTENT_FILTER: guild_models.GuildExplicitContentFilterLevel,
-            audit_log_models.AuditLogChangeKey.DEFAULT_MESSAGE_NOTIFICATIONS: guild_models.GuildMessageNotificationsLevel,  # noqa: E501
+            audit_log_models.AuditLogChangeKey.MFA_LEVEL: _with_int_cast(guild_models.GuildMFALevel),
+            audit_log_models.AuditLogChangeKey.VERIFICATION_LEVEL: _with_int_cast(guild_models.GuildVerificationLevel),
+            audit_log_models.AuditLogChangeKey.EXPLICIT_CONTENT_FILTER: _with_int_cast(
+                guild_models.GuildExplicitContentFilterLevel
+            ),
+            audit_log_models.AuditLogChangeKey.DEFAULT_MESSAGE_NOTIFICATIONS: _with_int_cast(
+                guild_models.GuildMessageNotificationsLevel
+            ),
```
rayakame, line 493: These `_with_int_cast` additions came up in a claude code review task, i couldn't verify it against the discord docs and didn't had the time yet to verify this with a test bot, but claude recommended to just cast it for now to be safe.

I will check what discord actually sends here and maybe change it back to normal

rayakame: Update, this seems to be extremely cursed, discord sometimes sends integers as strs or ints in the same payload. I think we should stay on the safe side and leave the `_with_int_cast`

rayakame: Mixed integer types observed:
  - `CHANNEL_OVERWRITE_UPDATE` (action type 14) (over the gateway `GUILD_AUDIT_LOG_ENTRY_CREATE`):
  deny:  new_value=str('0')     old_value=int(1024)
  allow: new_value=str('1024')  old_value=int(0)
  - The same entry fetched over REST had all four values as strings???
  - `options.type` on that entry is the string "0" (#1992), while `type` inside permission_overwrites snapshot lists on CHANNEL_CREATE/CHANNEL_DELETE came as int.
  
The keys that claude flagged and I added the `_with_int_cast` all come as ints as far as i could test, but since they are cheap, i think it would be better to leave them for now

rayakame:
I think this PR is required for #1779:

```py
class TTT(msgspec.Struct):
    x: ButtonStyle | int
    
decoder_ttt = msgspec.json.Decoder(TTT)
decoded_ttt = decoder_ttt.decode(b'{"x": 1}')
```
```
  File "/home/christopher/dev/python/test-msgspec/main.py", line 51, in <module>
    decoder_ttt = msgspec.json.Decoder(TTT)
TypeError: Type unions containing a custom type may not contain any additional types other than `None` - type `hikari.components.ButtonStyle | int` is not supported
```

PR #2775, opened 2026-08-07
Add `start_at` parameter to `fetch_members`

### LINKED ISSUE #902: Can't supply a starting point to REST.fetch_members
### Summary

Can't supply a starting point to REST.fetch_members

### Why is this needed?

So you can supply a starting point for fetch member iteraction

### Ideal implementation

To be decided.

### Checklist

rayakame:
### Checklist

### Related issues

Close #902


hikari/api/rest.py
```diff
@@ -5957,7 +5957,11 @@ async def fetch_member(
 
     @abc.abstractmethod
     def fetch_members(
-        self, guild: snowflakes.SnowflakeishOr[guilds.PartialGuild]
+        self,
+        guild: snowflakes.SnowflakeishOr[guilds.PartialGuild],
+        /,
```
rayakame, line 6002: This is a breaking change. I've added the `/` to have a more standardized api since `fetch_bans` has the same "layout". Idk maybe its worth it? I don't think a lot of people use kwargs for guild here. We can also remove the `/` from the `fetch_bans` endpoint or just ignore the api differences

hikari/impl/special_endpoints.py
```diff
@@ -629,15 +629,17 @@ def __init__(
...
         super().__init__()
         self._guild_id = snowflakes.Snowflake(str(int(guild)))
         self._route = routes.GET_GUILD_MEMBERS.compile(guild=guild)
         self._request_call = request_call
         self._entity_factory = entity_factory
-        # This starts at the default provided by Discord instead of the max snowflake
+        # This defaults to the value provided by Discord instead of the max snowflake
         # because that caused Discord to take about 2 seconds more to return the first response.
-        self._first_id = undefined.UNDEFINED
+        self._first_id = first_id
```
davfsa: We can drop this comment.

I still find this issue hilarious, lol

hikari/api/rest.py
davfsa: Ngl, happy to break this.

I doubt it affects anyone

hikari/impl/special_endpoints.py
rayakame: i will drop it, I found it quite interesting while reading through here

rayakame: I'll drop it, found it quite interesting while reading through here

davfsa (approved):

PR #2778, opened 2026-08-10
Add support for searching guild messages

rayakame:
### Summary
Add support for searching guild messages: `RESTClient.search_messages` with all of the documented filters, the `MessageSearchResult` model, its accompanying enums and the `SearchNotIndexedError` error raised while the guild's search index is not yet available

### Checklist


rayakame:
I just wanted to leave a note here, that I did not use a lazy iterator like used in `fetch_messages`, because the endpoint does not use cursor pagination, but instead uses offset pagination, has a hard ceiling at approx 10k messages and has some weird behavior when the guild isn't indexed yet.

PR #2796, opened 2026-08-21
Make the `TextSelectMenuBuilder` parent type parameter covariant

rayakame:
### Summary
I found this while trying [ty](https://docs.astral.sh/ty/) just for fun on hikari. This did not get caught before because we don't use pyright yet for the internals, and mypy doesnt catch it.

The type variable is only used in output position, so declaring it invariant needlessly made differently-parameterized builders mutually incompatible and caused the `add_text_menu` overrides to violate LSP on paper. 

### Checklist


rayakame:
~~I might use this pr/branch to resolve more pyright errors~~

davfsa (approved):

PR #2797, opened 2026-08-21
Add an experimental ty type checking pipeline

rayakame:
Adds a `nox -s ty` session running [ty](https://github.com/astral-sh/ty) with a configuration mirroring the pyright setup (same include/exclude list, python 3.10 target). **Not intended to be merged before ty reaches 1.0**, opening as a draft to track the experiment.

Two useful findings came out of the exercise: the `TextSelectMenuBuilder` variance fix (#2796) and a heads-up that `asyncio.get_event_loop_policy` (used in `internal/aio.py`) is deprecated as of 3.14.

### DIFF
--- a/.github/workflows/ty.yml
+++ b/.github/workflows/ty.yml
@@ -1,0 +1,43 @@
+name: Ty
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  ty:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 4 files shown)


.github/workflows/ty.yml
```diff
@@ -0,0 +1,40 @@
+name: Ty
```
davfsa, line 1: Ty is fast enough we can run it right after mypy.

Then we can do `|| exit 0` to avoid it cancelling everything

rayakame: I'll do this properly when ty ever reaches 1.0, until then this is just to visualize how much is failing

PR #2799, opened 2026-08-21
Make hikari fully compatible with strict pyright

rayakame:
### Checklist


rayakame:
https://discord.com/channels/574921006817476608/1540486296361898114

hikari/webhooks.py
```diff
@@ -494,7 +494,7 @@ class PartialWebhook(snowflakes.Unique):
 
     @typing_extensions.override
     def __str__(self) -> str:
-        return self.name if self.name is not None else f"Unnamed webhook ID {self.id}"
+        return self.name if self.name != "" else f"Unnamed webhook ID {self.id}"
```
rayakame, line 497: According to the discord docs, the `name` property of a webhook is nullable.

But in the api spec and everywhere else, `name` is typed as string only. Ive changed all webhook entity factories to `name=payload["name"] or ""`, and this method just to be 100% safe. 

I also added unit tests to make sure this doesn't break

PR #2820, opened 2026-09-16
Fix gateway rate limit handling across reconnects

rayakame:
### Checklist


rayakame:
Fable found some things when reviewing that I did not address (yet):
```
Before, a payload stuck behind the gateway rate limit when the connection dropped failed 
within about a minute. Now it's held until the shard reconnects, which during a long Discord 
or network outage could be a very long time. One place where this shows: 
GatewayBot.update_presence waits on the slowest shard with no timeout at all. 
``` 

```
Smaller leftover: voice.py could now catch SessionInvalidatedError and retry the voice 
state update as a follow-up.
```

Lmk if i should fix any of them

